PR #54, opened 2026-03-09
fix(benchmark): replace shutil.which(abicheck) guard with _PYTHON -m abicheck.cli check

napetrov:
## Problem
`shutil.which("abicheck")` returns `None` when abicheck is not installed as a system binary, even though it is fully available via `PYTHONPATH` as a module. This caused `run_abicheck()` and `run_abicheck_compat()` to silently return `SKIP` on machines where abicheck is used in dev mode.

## Fix
Replace both guards with `_HAS_ABICHECK` flag backed by `_abicheck_available()` which probes `_PYTHON -m abicheck.cli --help` at startup — consistent with how the benchmark actually invokes the tool.

## Impact
Benchmark now correctly measures abicheck timing instead of instantly returning SKIP.

## Summary by CodeRabbit

* **Documentation**
  * Benchmark docs updated to cover 41 cases with an aggregate accuracy table, per-case results, explicit gap notes, and revised run/setup guidance; README refocused on benchmark findings.

* **New Features**
  * Benchmark runner enhanced with multiple analysis modes (including two ABICC modes and a compat path), improved header handling, JSON summary output, clearer timeouts/errors, and refined CLI options.

* **Tests**
  * Default analysis mode adjusted to "both" and missing-tool skip behavior revised.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,5 @@
 # abicheck
 
-[![CI](https://github.com/napetrov/abicheck/actions/workflows/ci.yml/badge.svg)](https://github.com/napetrov/abicheck/actions/workflows/ci.yml)
-[![codecov](https://codecov.io/gh/napetrov/abicheck/branch/main/graph/badge.svg)](https://codecov.io/gh/napetrov/abicheck)
-
 **abicheck checks C/C++ library compatibility at both API and ABI levels.**
 
 abicheck is inspired by two foundational projects:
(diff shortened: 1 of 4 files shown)


napetrov:
Superseded by #52 (feat/benchmark-v2) — all fixes from this branch have been cherry-picked into #52 and merged.

PR #68, opened 2026-03-10
fix(benchmark): Makefile build, ABICC verdict parsing, timeout 120s, case05/13 COMPATIBLE

napetrov:
Replaces closed PRs #66 and #67 — all benchmark fixes in one commit.

### 1. Use Makefile when present
Benchmark was compiling sources directly (`gcc -shared -fPIC src.c`), losing SONAME and version-script linker flags. Now copies case dir and runs `make`, mirroring the pytest integration suite.

### 2. case05 / case13 expected: BREAKING → COMPATIBLE
With correct `.so` files, abicheck now returns `COMPATIBLE` for both:
- case05: SONAME added = bad-practice fix, not an ABI break
- case13: unversioned→versioned export = binary compatible

### 3. ABICC verdict: parse `Binary compatibility: N%` from output
`abi-compliance-checker` exits non-zero on GCC header-compilation warnings (GCC bug #78040) even when the report is correct. Previously 6 valid BREAKING results were counted as ERROR. Fix reads the percentage from stdout first.

### 4. DEFAULT_ABICC_TIMEOUT: 30s → 120s
C++ vtable/virtual cases (case09, case21, case22, case23) take >60s; 30s was causing spurious TIMEOUT results in the dumper workflow.

## Summary by CodeRabbit

* **Tests**
  * Increased ABICC timeout to 120s.
  * Test expectations now driven from external ground-truth data; two cases gained explicit expected_compat entries.
  * Compatibility parsing prefers reported percentage (regex) over exit codes and treats 100% as compatible.

* **New Features**
  * Makefile-based build support for test cases with automatic fallback to standard compilation.
  * Header/artifact handling and reporting adapted to support Makefile-built outputs.

### DIFF
--- a/examples/ground_truth.json
+++ b/examples/ground_truth.json
@@ -125,7 +125,8 @@
     },
     "case31_enum_rename": {
       "expected": "SOURCE_BREAK",
-      "category": "source_break"
+      "category": "source_break",
+      "expected_compat": "NO_CHANGE"
     },
     "case32_param_defaults": {
       "expected": "NO_CHANGE",
--- a/examples/ground_truth.json
+++ b/examples/ground_truth.json
@@ -137,7 +137,8 @@
     },
     "case34_access_level": {
       "expected": "SOURCE_BREAK",
-      "category": "source_break"
+      "category": "source_break",
+      "expected_compat": "NO_CHANGE"
     },
     "case35_field_rename": {
       "expected": "BREAKING",
--- a/scripts/benchmark_comparison.py
+++ b/scripts/benchmark_comparison.py
@@ -77,57 +77,19 @@ _HAS_ABICHECK: bool = _abicheck_available()
 
 DEFAULT_ABICC_TIMEOUT = 120  # seconds
 
-# Expected verdicts from case READMEs
+# Expected verdicts loaded from ground_truth.json — single source of truth.
+# To add/change a verdict, edit examples/ground_truth.json only.
+_GT_PATH = Path(__file__).parent.parent / "examples" / "ground_truth.json"
+_gt_data = json.loads(_GT_PATH.read_text())
 EXPECTED: dict[str, str] = {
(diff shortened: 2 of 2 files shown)


napetrov:
Addressed all review comments in commit `7a720b3`:

**CodeRabbit (line 457) — `Binary compatibility: 100%` → COMPATIBLE not NO_CHANGE**
Fixed. `pct == 100.0` now maps to `COMPATIBLE` (ABICC 100% = no breaking changes, but compatible additions can still be present). `pct < 100.0` → `BREAKING`.

**CodeRabbit (line 654) — make failure should fall back to compile_so()**
Fixed. On `make` failure or missing artifacts the benchmark now falls back to `compile_so()` instead of recording `ERROR`. Header remap is also guarded by `if v1_so == built_v1`.

**Codex (line 634) — header remap may invert good/bad layout**
Fixed. `_remap` is only called when `v1_so == built_v1` (Makefile build succeeded). When fallback `compile_so()` is used, `v1_h_hint`/`v2_h_hint` stay as-is from `find_sources()`.

napetrov:
Addressed all comments in commits `35a67b6`–`d3b281f`:

- **`import shutil as _shutil` (line None)** — removed in `82fba46`; using top-level `shutil` directly
- **Makefile fallback (line 631)** — fixed in `35a67b6`: `make` failure now falls back to `compile_so()` instead of ERROR; also wrapped in `TimeoutExpired` handler
- **ABICC 100%→COMPATIBLE (line 428)** — intentional: ABICC cannot emit NO_CHANGE. Added `EXPECTED_ABICC` dict (NO_CHANGE→COMPATIBLE mapping) in `82fba46` so abicc_xml/dumper accuracy columns score correctly
- **schema validation (line 102)** — fixed in `d3b281f`: validates `verdicts` key presence and `expected` field on every case; raises `SystemExit` with clear message

PR #148, opened 2026-03-17
fix(policy): downgrade enum_last_member_value_changed to risk 

napetrov:
## Context

From oneDNN 2025.0.2 -> 2025.1.0 scan, report was overly harsh () with mostly enum additions + DT_NEEDED add.

## Changes

- Move  from  to 
- Keep  and  as non-breaking (existing policy)
- Add regression test for oneDNN-like combination:
  - enum member additions
  - enum sentinel () value change
  - 
  - expected verdict: 
- Update sprint1 policy expectation test accordingly

## Why

Sentinel moves are source/deployment risk (loop bounds, array sizes), but not guaranteed binary ABI break for already-compiled consumers.

## Notes

This PR does **not** change  classification yet; that remains the likely remaining hard-break driver in some oneDNN outputs and should be evaluated separately with targeted policy options.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reclassified last-member enum value changes from breaking to deployment-risk, improving compatibility verdicts; regular enum member value changes remain breaking.

* **Tests**
  * Added tests covering enum additions combined with a last-member value change as a risk.
  * Updated and renamed existing tests to assert COMPATIBLE_WITH_RISK where applicable.

* **Documentation**
  * Updated example case and explanatory text to reflect the new risk classification.

### DIFF
--- a/abicheck/checker.py
+++ b/abicheck/checker.py
@@ -2612,10 +2612,22 @@ def _diff_enum_layouts(o: object, n: object) -> list[Change]:
             ))
 
         # 3. Changed values
+        # Sentinel member = member with the highest old value.
+        # If this member changes, classify as ENUM_LAST_MEMBER_VALUE_CHANGED.
+        old_sentinel_name = None
+        if old_e.members:
+            _max_val = max(old_e.members.values())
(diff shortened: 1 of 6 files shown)


napetrov:
@coderabbitai resume

PR #151, opened 2026-03-17
fix(classification): name-pattern sentinel detection, case52 known_gap, fix case count

napetrov:
Addresses classification issues from bug report and review comments in #149.

### 1. Sentinel detection by name-pattern (checker.py)
Instead of detecting the sentinel as the member with the highest numeric value, now uses name-pattern matching:
- suffixes: `*_last`, `*_max`, `*_count`
- exact names: `last`, `max`, `count`

This correctly handles real-world cases like `dnnl_format_tag_last`, `VK_FORMAT_MAX_ENUM`, etc. where the sentinel may not always be at the max numeric value. Classified as `ENUM_LAST_MEMBER_VALUE_CHANGED` → `COMPATIBLE_WITH_RISK` (not BREAKING).

### 2. case52_rpath_leak: add known_gap (examples/ground_truth.json)
Documents the tool limitation: compares RPATH for equality only, does not detect non-portable hardcoded build-directory paths.

### 3. Update expected case count (tests/test_validate_examples_unit.py)
Bumped `_EXPECTED_CASE_COUNT` from 48 → 63 to match cases added in #149.


napetrov:
Closing — wrong base branch. Clean PR from main follows.

PR #257, opened 2026-05-30
refactor(diff): extract bool_transition + diff_by_key detector helpers (Tier 1)

napetrov:
## What & why

Follow-up to an architecture evaluation of the detection logic. The
finding: the *top-level* design is already sound (decorator-based detector
registry + single-declaration `ChangeKind` registry + set-based verdicts).
The real smell the "tree of if cases" impression points at is **leaf-level
boilerplate duplication** inside individual detectors — not a fragile
dispatch tree.

This is **Tier 1**: extract the two repeated shapes into small,
behavior-preserving building blocks. No policy or classification change.

## Changes

New module `abicheck/diff_helpers.py`:

- **`bool_transition(...)`** — collapses the hand-rolled `if/elif` pairs
  around near-identical `Change` constructions (a flag going off→on / on→off).
  Preserves each site's bespoke wording and the tri-state rule (`None` =
  "not recorded in this snapshot") via `skip_none`. Applied to the
  `noexcept`, `virtual`, hidden-friend, `explicit`, and variable const-ness
  transitions in `diff_symbols.py`.
- **`diff_by_key(...)`** — factors out the removed / added / common map-diff
  scaffold so a detector only supplies per-bucket callbacks. The
  `variables` detector now uses it, with matched-pair logic moved to a small
  `_check_variable` helper. Iteration order is unchanged (removed/common in
  old-map order, added in new-map order), so emitted change ordering is
  identical.

## Verification

- Full fast suite: **6370 passed**, 16 skipped (`pytest -m "not integration
  and not libabigail and not abicc and not slow and not golden"`).
- New `tests/test_diff_helpers.py`: **13 tests** covering both helpers,
  including the tri-state skip and the falsey-value-as-present-key edge case.
- `ruff check`, `mypy abicheck/`, and `scripts/check_ai_readiness.py` all
  clean (0 errors / 0 warnings).

## Scope / non-goals

Deliberately stops at Tier 1. The map-diff scaffold can later extend to the
`functions` and `types` detectors, and a per-record field-comparator table
was considered but left out to avoid trading readability for uniformity.

https://claude.ai/code/session_01XMfP2fSXzyREMoZFUuRJ6S

---
_Generated by [Claude Code](https://claude.ai/code/session_01XMfP2fSXzyREMoZFUuRJ6S)_

### DIFF
--- a/abicheck/diff_helpers.py
+++ b/abicheck/diff_helpers.py
@@ -1,0 +1,124 @@
+# Copyright 2026 Nikolay Petrov
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Reusable building blocks for diff detectors.
+
+Detectors repeat two structural patterns:
+
+* **Boolean attribute transitions** — "flag went from off→on / on→off"
+  (e.g. ``noexcept`` added/removed, ``virtual`` added/removed). Each site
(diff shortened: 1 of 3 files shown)


napetrov:
@coderabbitai review

napetrov:
@coderabbitai review

PR #274, opened 2026-06-01
docs: sync example catalog counts

napetrov:
## Summary
- update README and docs to distinguish the current 121-example catalog from the release-pinned 74-case benchmark subset
- clarify that the generated examples encyclopedia covers 116 single-library cases, while bundle/release-level cases live under examples/
- refresh related development/reference docs that still described the old 74-case catalog as current

## Validation
- pytest -q tests/test_examples_docs.py
- mkdocs build --strict
- git diff --check


napetrov:
@coderabbitai review

PR #282, opened 2026-06-03
Fix stale SONAME bump verdict after rename filtering

napetrov:
## Summary

Fixes a real-world false-positive BREAKING verdict where raw removed/added symbols were downgraded to `func_likely_renamed`, but still left behind a stale SONAME bump recommendation and exit 4.

Changes:

- run SONAME bump policy after the post-processing pipeline, not before it
- keep rename-suppressed removed/added findings auditable, but exclude `rename:` redundant findings from verdict-driving redundant changes
- add a regression covering the postprocessed rename path

## Real-world evidence

Found during the recurring real-world validation campaign:

- c-ares 1.14.0 -> 1.18.1 (`libcares.so`)
- libssh2 1.9.0 -> 1.10.0 (`libssh2.so`)

Pre-fix, both reported:

- `verdict: BREAKING`
- `summary.breaking: 0`
- `binary_compatibility_pct: 100.0`
- release recommendation `version_bump: major`, `soname_action: bump_missing`

Independent cross-checks showed one removed internal-looking exported symbol in each case, matching abicheck's own `func_likely_renamed` risk downgrade rather than a report-visible breaking finding.

Post-fix reruns report `COMPATIBLE_WITH_RISK`, exit 0, and `soname_action: no_bump_needed`.

Artifacts are recorded in the local campaign ledger under `reports/abicheck-realworld-cron/artifacts/20260603-0008/`.

## Tests

- `ruff check abicheck/checker.py tests/test_elf_version_policy.py`
- `pytest -q tests/test_elf_version_policy.py::TestCheckerIntegration::test_compare_soname_policy_uses_postprocessed_changes tests/test_elf_version_policy.py::TestCheckerIntegration::test_compare_soname_bump_recommended tests/test_semver_recommendation.py::test_breaking_with_soname_bump_recommended_flags_missing_bump tests/test_exit_codes.py::test_exit_0_compatible_with_risk`
- `pytest -q tests/test_elf_version_policy.py tests/test_semver_recommendation.py tests/test_exit_codes.py tests/test_real_world_false_positives.py`

PR #281, opened 2026-06-03
fix(dwarf): filter transitive stdlib subprograms

napetrov:
## Summary
- share the ELF ABI-relevance symbol filter with DWARF snapshot extraction
- filter DWARF export indices and DW_AT_deleted subprograms so transitive stdlib/runtime subprograms do not enter non-runtime libraries as public ABI
- add regression coverage for DWARF export-index filtering

## Real-world evidence
First abicheck-realworld-cron run found oneTBB libtbbmalloc 2021.5.0 -> 2021.9.0 as a false-positive candidate: abicheck reported BREAKING while abidiff was clean.

This patch removes the bogus std:: deleted-subprogram roots: a fresh oneTBB dump now has 0 std:: DWARF functions (was 62), and the compare no longer reports the four std:: FUNC_REMOVED findings.

The broader oneTBB internal-type reachability issue remains open and is documented in reports/abicheck-realworld-cron/finding-analysis-20260602-2138.md.

## Tests
- python -m ruff check abicheck/elf_symbol_filter.py abicheck/dumper.py abicheck/dwarf_snapshot.py tests/test_elf_symbol_filters.py
- python -m pytest tests/test_elf_symbol_filters.py tests/test_real_world_false_positives.py tests/test_dwarf_coverage_gaps.py -q

## Summary by CodeRabbit

* **Bug Fixes**
  * Centralized ELF-symbol ABI-relevance filtering and applied it across DWARF extraction, dumping, and diffing to reduce false-positive ABI change reports and better reconcile stripped vs DWARF views.
  * Public/function-surface detection now consults ELF export evidence and preserves deleted-but-still-relevant entries.

* **Behavior**
  * ELF-only function removals (`func_removed_elf_only`) are now classified as BREAKING under strict policy.

* **Tests**
  * Expanded tests covering ELF export filtering, DWARF interplay, and stripped-side suppression.

* **Documentation**
  * Updated ADRs, examples, and reference docs to reflect the new classification and guidance.

PR #283, opened 2026-06-03
Fix surface scoping for field-level type changes

napetrov:
## Summary

Fixes a real-world false-positive BREAKING verdict where field-level DWARF type findings such as `p11_virtual::funcs` were classified against the full field spelling instead of the owning type. Because `p11_virtual::funcs` is not a known type name, the surface classifier kept it as an unknown in-surface finding even though the owner type `p11_virtual` is internal/private.

Changes:

- classify field-level type findings by their owner type before public-surface reachability checks
- keep public/unknown anti-hiding behavior for non-field type findings unchanged
- add a regression test for private field-level owner demotion

## Real-world evidence

Found during the recurring real-world validation campaign:

- p11-kit 0.25.10 -> 0.26.2 (conda-forge `libp11-kit.so.0`)

Pre-fix abicheck reported:

- `verdict: BREAKING`
- one visible `struct_field_type_changed` for `p11_virtual::funcs`
- no removed dynamic exports
- `abidiff` exit 0

ELF evidence shows the related `p11_virtual_*` functions/data are LOCAL symbols, not dynamic exports. Post-fix reruns move the internal field/type churn into the out-of-surface audit ledger and report `NO_CHANGE`, exit 0.

Local artifacts are recorded under `reports/abicheck-realworld-cron/artifacts/20260603-0241/`:

- initial JSON: `runs/P11KIT_0_25_10_to_0_26_2__libp11-kit.json`
- libabigail: `libabigail/P11KIT_0_25_10_to_0_26_2__libp11-kit.abidiff.txt`
- export diff: `exports/P11KIT_0_25_10_to_0_26_2__libp11-kit.removed.txt`
- post-fix rerun: `rerun-after-field-surface-fix-main/P11KIT_0_25_10_to_0_26_2__libp11-kit.json`

## Tests

- `pytest -q tests/test_surface_confidence.py tests/test_report_filtering.py`
- p11-kit real-world rerun: exits 0 and reports `NO_CHANGE`

Verification gap:

- `tests/test_surface_property.py` could not be collected in this environment because `hypothesis` is not installed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced field-level type change detection to properly resolve types through their owning context for accurate public surface classification.

* **Tests**
  * Added regression test validating field-level type change handling for non-public types.

### DIFF
--- a/abicheck/surface.py
+++ b/abicheck/surface.py
@@ -105,6 +105,23 @@ _TYPE_LEVEL_KIND_NAMES: frozenset[str] = frozenset(
     }
 )
 
+_FIELD_LEVEL_TYPE_KIND_NAMES: frozenset[str] = frozenset(
+    {
+        "type_field_removed",
+        "type_field_added",
+        "type_field_offset_changed",
+        "type_field_type_changed",
+        "type_field_added_compatible",
+        "field_bitfield_changed",
+        "union_field_added",
+        "union_field_removed",
+        "union_field_type_changed",
+        "struct_field_offset_changed",
+        "struct_field_removed",
+        "struct_field_type_changed",
+    }
+)
+
 # Tokens that are type qualifiers / keywords, not type names.
 _TYPE_NOISE: frozenset[str] = frozenset(
     {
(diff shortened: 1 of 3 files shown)


napetrov:
Combined #281/#282/#283 review result for this PR.

I did not find a blocking correctness issue in #283. The `Type::field` owner extraction is a narrow and reasonable fix for the p11-kit case: field-level type findings should be scoped by the owner type (`p11_virtual`), not by the full field spelling (`p11_virtual::funcs`) as if it were an independent type.

Local validation:
- merged #281 + #282 + #283 together in a temp worktree: clean merge
- focused combined suite: `175 passed` across real-world false-positive regressions, ELF symbol filters, SONAME policy tests, and surface-confidence tests

Follow-up testing I would add after this PR:
- cover every kind in `_FIELD_LEVEL_TYPE_KIND_NAMES`, not just one field-change kind
- add public-owner and private-owner variants so we pin both keep and demote behavior
- add nested owners such as `ns::Type::field`, `Outer::Inner::field`, and template owners like `Box<int>::value`
- add caused-by-only / mismatched `symbol` + `caused_by_type` cases

Broader architecture note: this should eventually move off string parsing. Detectors should emit structured subject metadata (`kind=field`, `owner_type=p11_virtual`, `field_name=funcs`), and `surface.classify_change_surface()` should consume that instead of recovering ownership with `rsplit("::", 1)`.

PR #284, opened 2026-06-03
docs: clarify benchmark suite coverage

napetrov:
## Summary
- clarify in README that the project has a 121-case full examples catalog, while the published validation table is the historical 74-case cross-tool benchmark subset
- add `--suite {all,pinned74}` to `scripts/benchmark_comparison.py` and record the selected suite in `benchmark_report.json`
- make release publishing explicitly run `--suite pinned74`
- align the tool-comparison reference page and add smoke coverage for the suite option and 74-case selector

## Why
The README looked stale because the validation snapshot still talked only about 74 cases even though the examples catalog has 121 cases. The 74-case table is still useful as a stable release-to-release cross-tool benchmark, but it needed to be labeled that way and the full-catalog command needed to be explicit.

## Validation
- `python -m pytest tests/test_benchmark_smoke.py`
- `python -m ruff check scripts/benchmark_comparison.py tests/test_benchmark_smoke.py`
- `git diff --check`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -83,7 +83,7 @@ jobs:
         run: pip install -e ".[dev]"
 
       - name: Run benchmark
-        run: python scripts/benchmark_comparison.py --abicc-timeout 60
+        run: python scripts/benchmark_comparison.py --suite pinned74 --abicc-timeout 60
 
       - name: Upload benchmark report (workflow artifact)
         uses: actions/upload-artifact@v4
--- a/README.md
+++ b/README.md
@@ -144,13 +144,13 @@ See `abicheck.service` for the full signature, plus the [MCP server integration]
 
 ## Examples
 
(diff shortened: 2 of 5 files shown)


napetrov:
Updated after review: the main README no longer presents the 74-case pinned table as the primary validation snapshot. It now points at the full 121-case catalog as the main validation target and keeps the historical pinned74 cross-tool comparison in the reference docs/release report path.

Also fixed a real `--suite all` tooling issue found while validating this: bundle cases with `expected: null` now render as unscored `?` instead of crashing the benchmark runner.

Local validation:
- `python scripts/benchmark_comparison.py --suite all --tools abicheck --skip-abicc --cases case84` passes and reports `? / NO_SOURCE`
- `pytest tests/test_benchmark_smoke.py -q`: 19 passed
- `ruff check scripts/benchmark_comparison.py tests/test_benchmark_smoke.py`: passed
- `git diff --check`: passed

Commit: `1c48cba docs: make readme validation target full catalog`

PR #291, opened 2026-06-04
Avoid ELF-only false positives for retained wrappers and linker artifacts

napetrov:
## Summary
- avoid ELF-only fingerprint rename candidates when the old function name is still exported by the new library
- filter exact ELF linker artifacts (`_init`, `_fini`, `__bss_start`, `_edata`, `_end`) from ELF-only public symbol surfaces
- apply the shared ABI-relevance filter to snapshot-only ELF function/variable maps and visibility-leak reporting
- add regressions for retained compatibility wrappers and linker boundary artifacts
- guard optional ELF metadata in the rename export lookup so type checks stay clean

## Real-world evidence
- libssh2 1.11.0 -> 1.11.1 kept `libssh2_session_callback_set` exported as a 6-byte wrapper and added `libssh2_session_callback_set2`; before the fix abicheck reported a false `func_likely_renamed`, and after the fix it reports `set2` as additive
- libev 4.25 -> 4.33 dropped linker boundary symbols `__bss_start`, `_edata`, and `_end`; before the fix abicheck reported `BREAKING`, while `abidiff` reported no ABI change; after the fix the real libev rerun reports `COMPATIBLE`

## Tests
- `ruff check abicheck tests`
- `mypy abicheck`
- `python scripts/check_ai_readiness.py` (0 errors, existing warnings only)
- `pytest -q tests/test_binary_fingerprint.py`
- `pytest -q tests/test_elf_symbol_filters.py tests/test_sprint2_elf.py tests/test_binary_fingerprint.py`

### DIFF
--- a/abicheck/diff_platform.py
+++ b/abicheck/diff_platform.py
@@ -36,6 +36,7 @@ from .diff_platform_templates import (
 from .diff_symbols import _public_functions
 from .diff_types import _RESERVED_FIELD_RE
 from .elf_metadata import SymbolBinding, SymbolType
+from .elf_symbol_filter import is_abi_relevant_elf_symbol
 from .model import (
     AbiSnapshot,
     Visibility,
@@ -281,7 +282,11 @@ def _diff_visibility_leak(old: AbiSnapshot, new: AbiSnapshot) -> list[Change]:
 
     leaked = [
         f for f in old.functions
-        if f.visibility == Visibility.ELF_ONLY and _looks_internal(f.name)
+        if (
+            f.visibility == Visibility.ELF_ONLY
+            and is_abi_relevant_elf_symbol(f.name)
+            and _looks_internal(f.name)
+        )
     ]
     if not leaked:
         return []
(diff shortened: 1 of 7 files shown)


napetrov:
Addressed the remaining rename-fingerprint review feedback in `bd604e8`.

What changed:
- `_fingerprints_from_elf()` now skips non-ABI ELF symbols via `is_abi_relevant_elf_symbol()`, so `_init`/`_fini`/linker artifacts cannot become rename candidates.
- Retained ABI exports are removed from both old-side and new-side fingerprint maps before rename matching, preserving the matcher's old-only/new-only distinction.
- Added regressions for `_init -> lib_init` and old `{foo_v1, foo_v2}` vs new `{foo_v1}`.

Local proof:
- `pytest -q tests/test_elf_symbol_filters.py tests/test_binary_fingerprint.py tests/test_sprint2_elf.py` -> 216 passed
- `ruff check abicheck/diff_symbols.py tests/test_binary_fingerprint.py tests/test_elf_symbol_filters.py tests/test_sprint2_elf.py` -> passed

PR #353, opened 2026-06-11
fix(probe_harness): validate probe manifest inputs

napetrov:
### Motivation

- The probe harness parsed manifest-provided `compiler`, `flags`, `id`, and `probe.name` without validation, enabling manifest-driven command execution and filesystem writes outside the intended `work_dir`.
- Malicious manifests could set `compiler` to `/bin/sh` with `-c` payloads, or use bare shell names like `sh`/`bash` already on the PATH, or pass joined flag forms like `-MF/tmp/evil.d` to write files outside `work_dir`. Inputs must be constrained before use.
- The change aims to close that security hole while preserving the existing compile-and-snapshot behavior for trusted manifests.

### Description

- Add a safe-component regex `_SAFE_COMPONENT_RE` and a set `_DISALLOWED_COMPILER_FLAGS` to define acceptable path components and forbidden flags respectively.
- Add `_ALLOWED_COMPILER_BASENAMES` (a frozenset of `g++`, `gcc`, `clang++`, `clang`, `c++`, `cc`) so that `_validate_compiler_name` rejects bare shell/tool names like `sh`, `bash`, and `rm` while still permitting versioned suffixes such as `g++-13` or `clang++-15`.
- Add `_DISALLOWED_FLAG_PREFIXES` (`-o`, `-x`, `-MF`, `-MT`, `-MQ`) and update `_validate_flags` to extract the flag name via `split("=", 1)[0]` before checking both exact membership and prefix matches, blocking joined forms like `-MF/tmp/evil.d` and `-o/path` that bypassed the previous exact-match check.
- Extend `_DISALLOWED_COMPILER_FLAGS` with `-MD` and `-MMD` to block dependency-file-generation flags that write files outside `work_dir` during compilation.
- Implement `_validate_safe_component`, `_validate_compiler_name`, and `_validate_flags` helper functions and wire them into `parse_probe_spec` so manifest `id`, `compiler`, `flags`, and `probe.name` are validated when parsing into `ProbeConfiguration`/`Probe` objects.
- Add unit tests in `tests/test_probe_harness.py` that assert invalid configuration IDs, probe names, compiler values (including bare interpreter names), and disallowed flags (including joined forms and dependency-generation flags) are rejected.

### Testing

- Ran `pytest -q tests/test_probe_harness.py` — 39 tests passed (up from 30).
- The new unit tests exercise the hardened validation logic and will prevent regressions for the parsed manifest surface (`load_probe_spec`/`parse_probe_spec`).

------
<a href="https://chatgpt.com/codex/cloud/tasks/task_e_6a0f579afa0c832bb0a12420d99099c3">Codex Task</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input validation: invalid configuration IDs, probe names, compiler identifiers, and disallowed compiler flags are now rejected earlier with clearer errors to prevent downstream failures.
* **Tests**
  * Added tests to ensure malformed configurations and unsupported flags raise appropriate parsing errors.

### DIFF
--- a/abicheck/probe_harness.py
+++ b/abicheck/probe_harness.py
@@ -50,6 +50,7 @@ from __future__ import annotations
 
 import json
 import os
+import re
 import shutil
 import subprocess
 import tempfile
@@ -198,6 +199,8 @@ class MatrixSnapshot:
 
 
 _CXX_STD_FLAG = "-std=c++"
+_SAFE_COMPONENT_RE = re.compile(r"^[A-Za-z0-9_.-]+$")
+_DISALLOWED_COMPILER_FLAGS = frozenset({"-c", "-o", "-x", "-E", "-S", "-M", "-MF", "-MT", "-MQ", "-pipe", "--"})
 
 
 def _parse_cxx_std(flags: list[str]) -> int | None:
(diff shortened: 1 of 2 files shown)


napetrov:
@copilot  - please address recommendations from code review

PR #362, opened 2026-06-12
feat: detect runtime-model build-mode flips and struct-return convention

napetrov:
## What

Implements the genuine detection gaps found while analysing the "AST and build-aware abicheck" proposal. The proposal's architecture (compile-command capture, source/call/include graphs, clang/castxml replay, stdlib-ABI and toolchain-drift diffing) is **already shipped** in `abicheck/buildsource/` (ADR-028..033) with 238 ChangeKinds and 129 example cases. The remaining gaps were the runtime-model build flags and the struct-return convention — this PR closes them.

## New ChangeKinds (238 → 243)

| Kind | Verdict | Source |
|------|---------|--------|
| `exceptions_mode_changed` | RISK | L3 build-evidence diff (`-fexceptions` ↔ `-fno-exceptions`) |
| `rtti_mode_changed` | RISK | L3 build-evidence diff (`-frtti` ↔ `-fno-rtti`) |
| `tls_model_changed` | RISK | L3 build-evidence diff (`-ftls-model=`, `-fextern-tls-init`) |
| `threadsafe_statics_mode_changed` | RISK | L3 build-evidence diff (`-fno-threadsafe-statics`) |
| `struct_return_convention_changed` | **BREAKING** | DWARF value-ABI return-trait flip (in-register ↔ hidden sret) |

The four mode kinds default to RISK per the ADR-028 D3 authority rule (build-context findings never decide a shipped-ABI break on their own; the artifact diff confirms). The struct-return kind is artifact-proven from DWARF, so it is BREAKING — as agreed.

## How

- **Canonical normalization** (`adapters/base.py`): runtime-model flags map to canonical `exceptions`/`rtti`/`tls_init`/`tls_model`/`threadsafe_statics` keys with `on`/`off` values, so a flip is order-independent and an explicit-on vs an omitted default never reads as a change.
- **Routing** (`build_diff.py`): those canonical keys route to their dedicated finding (with a default-aware no-op guard) instead of the generic `abi_relevant_build_flag_changed`.
- **Return-trait split** (`dwarf_advanced.py` + `checker.py`): when the `ret:` component of the value-ABI trait flips, emit `struct_return_convention_changed`; a parameter-position flip stays `value_abi_trait_changed`. Mirrored in the reporter / surface / report-classification lists.
- Documented all five in the Change Kind Reference; bumped the documented total to 243 across README/docs/CLAUDE.md.

## Tests

- New unit coverage for each kind: mode flips, the on-vs-absent no-op guard, the off-vs-absent flip, canonical key normalization, and the return-vs-parameter trait split.
- Updated the three existing value-ABI tests whose fixtures were return-position flips (now correctly the struct-return refinement) and added explicit parameter-position cases so both kinds stay covered.
- Green: `ruff check`, `mypy` (0 issues), `check_ai_readiness.py` (0 errors), full fast suite (9904 passed).

## Still to come on this branch

Example cases (`caseNNN_exceptions_mode_flip`, `…_rtti_mode_flip`, `…_threadsafe_statics`, `…_tls_model_change`, `…_struct_return_convention`, `…_return_type_silent_itanium`, `…_macro_variant_matrix`), FP-rate corpus additions, and a CI variant-matrix lane — tracked as follow-up commits.

https://claude.ai/code/session_01URKM3XH2W8y8V9oHjE1mVA

---
_Generated by [Claude Code](https://claude.ai/code/session_01URKM3XH2W8y8V9oHjE1mVA)_

## Release Notes

* **New Features**
  * Expanded ABI/API change detection from 238 to 243 types
  * Added struct-return calling convention change detection (register vs. hidden-pointer returns)
  * Added build-mode flip detection for exceptions, RTTI, TLS model, and thread-safe statics flags
  
* **Examples**
  * Increased example catalog from 129 to 134 validation cases
  * Added 5 new scenarios demonstrating struct-return conventions and build-mode compatibility risks


napetrov:
Addressed the runtime-mode review comments in 44e7f7f:

- **Per-TU last-one-wins** (line 296): runtime-mode flags are now resolved per compile unit with last-wins semantics, so `-fno-exceptions -fexceptions` in one TU records only the effective `on`.
- **Mixed-build default masking** (the earlier exceptions P2 and the TLS line-80 P2): rather than extend the default-on fill to TLS, I removed it entirely. Recording an implicit default for every C++ TU polluted normal `build_options` and was inconsistent across modes (and TLS has no single portable default to fill). The rare case — one library built half default-on and half `-fno-*`, then partially flipped — is now a documented limitation in `derive_build_options`; the artifact diff still proves any concrete break, and all-or-nothing flips across versions are detected. This keeps the L3 signal advisory and noise-free, consistent with ADR-028 D3.

---

PR #368, opened 2026-06-13
Validate qualified-name enrichment per snapshot to prevent namespace spoofing for extern "C" symbols

napetrov:
### Motivation
- The prior enrichment built a single mangled→qualified-name map from both old and new snapshots and attached the recovered name to every `Change` based solely on the exported symbol, which allowed a new-side C++ namespace to suppress breaking changes to an existing global `extern "C"` export.
- This change prevents an attacker-controlled header/artifact from spoofing a trusted namespace onto a stable C-linkage export and hiding real ABI breaks from suppression rules.

### Description
- Add ` _qualified_functions_by_mangled` and `_qualified_name_for_change` to `abicheck/diff_filtering.py` and use them in `_enrich_source_locations` so old/new qualified-name lookups are kept side-specific and the recovered qualified name is trusted only when appropriate (one-sided names for `FUNC_ADDED`/`FUNC_REMOVED`, and agreement required for existing-symbol changes).
- Update `EscalateFrozenNamespaceViolations` in `abicheck/post_processing.py` to use the side-aware lookup and to consult the side-validated qualified name when matching frozen-namespace patterns (matching now passes the `Change` into the matcher to apply the per-change logic).
- Add a regression test `test_extern_c_namespace_suppression_rejects_one_sided_spoof` in `tests/test_frozen_namespace.py` that verifies a new-side namespace alone does not suppress ABI breakage on a previously-global `extern "C"` symbol.

### Testing
- Ran `pytest tests/test_frozen_namespace.py -q`: all tests in that file passed (`24 passed`).
- Ran `python -m compileall abicheck tests/test_frozen_namespace.py`: compilation succeeded.
- Ran full `pytest -q`: test collection failed due to a missing test dependency (`hypothesis`), so full-suite results were not produced.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1a35b4c664832b82e8a7bbb04786de)

### DIFF
--- a/abicheck/diff_filtering.py
+++ b/abicheck/diff_filtering.py
@@ -127,6 +127,37 @@ def _safe_index(snap: AbiSnapshot) -> bool:
     return True
 
 
+def _qualified_functions_by_mangled(snap: AbiSnapshot | None) -> dict[str, str]:
+    """Return mangled/exported function names that have C++ qualification."""
+    if snap is None or not _safe_index(snap):
+        return {}
+
+    qualified: dict[str, str] = {}
+    for mangled, fn in (getattr(snap, "_func_by_mangled", None) or {}).items():
+        fname = getattr(fn, "name", None)
+        if fname and "::" in fname and mangled not in qualified:
+            qualified[mangled] = fname
+    return qualified
+
+
+def _qualified_name_for_change(
+    c: Change,
+    old_qualified: dict[str, str],
+    new_qualified: dict[str, str],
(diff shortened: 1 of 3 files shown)


napetrov (commented):
Reviewed after rebase/conflict resolution. Addressed bound-method regression in EscalateFrozenNamespaceViolations; local proof: ruff on changed files, pytest tests/test_frozen_namespace.py (24 passed).

PR #369, opened 2026-06-13
fix(probe): validate probe manifest toolchain inputs to prevent RCE

napetrov:
### Motivation
- The probe harness accepted manifest-controlled `compiler` and `flags` which made `abicheck probe run` able to execute arbitrary local commands when fed a malicious spec.

### Description
- Add validators and sanitisation in `abicheck/probe_harness.py` to validate configuration/probe identifiers, compiler executable names, flag strings, and string sequences before they are used.
- Introduce safe-name and safe-compiler regexes and an allowlist for flag prefixes/exact flags (`_SAFE_ID_RE`, `_SAFE_COMPILER_RE`, `_SAFE_FLAG_PREFIXES`, `_SAFE_FLAG_EXACT`) and helper functions `_validate_manifest_id`, `_validate_compiler_name`, `_validate_flag`, and `_validate_string_sequence`.
- Wire the validators into `parse_probe_spec` so `id`, `compiler`, `flags`, `include_dirs`, probe `name`, and `headers` are checked while preserving valid manifest behaviour.
- Add unit and CLI regression tests in `tests/test_probe_harness.py` and `tests/test_cli_probe.py` that assert rejection of interpreter/compiler paths, dangerous flags, and path-like identifiers.

### Testing
- Ran `python3 -m pytest tests/test_probe_harness.py tests/test_cli_probe.py -q` and the modified tests passed (`42 passed`).
- Ran linting and static checks with `python3 -m ruff` and `python3 -m mypy` against the changed files and they reported no issues.
- Ran the full test suite with `python3 -m pytest -q` where a pre-existing performance threshold test failed due to timing (one long-running performance test exceeded its asserted 5s limit), which is unrelated to the manifest validation changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1b4f132008832b83969335de7b4e25)

### DIFF
--- a/abicheck/package.py
+++ b/abicheck/package.py
@@ -41,7 +41,7 @@ import tempfile
 import zipfile
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import IO, Protocol, runtime_checkable
+from typing import IO, Any, Protocol, runtime_checkable
 
 from .errors import ExtractionSecurityError, SnapshotError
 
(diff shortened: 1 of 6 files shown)


napetrov (commented):
Reviewed after rebase/conflict resolution. Addressed manifest validation test/lint feedback; local proof: ruff on changed files, pytest tests/test_probe_harness.py tests/test_cli_probe.py (65 passed).

PR #375, opened 2026-06-13
fix(source-abi): preserve API break for generated constexpr changes

napetrov:
### Motivation
- A recent change marked constants from generated public headers as `generated`, which caused value changes to be classified as `generated_header_changed` (lower-severity) instead of `constexpr_value_changed` (an `API_BREAK`), weakening legacy exit-code gating for baked-in constexpr values.
- The intent is to keep removals of generated entities handled by the generated-header path while preserving the stronger API-break semantics for changed `constexpr` values so CI gates behave correctly.

### Description
- In `abicheck/evidence/source_diff.py` special-case `constexpr` in `_diff_generated` so generated `constexpr` value changes are not consumed by the generated-header path and can still be reported as `CONSTEXPR_VALUE_CHANGED`.
- In `abicheck/evidence/source_diff.py` move the generated-entity `continue` in `_diff_declarations` to after the `constexpr` branch so `constexpr` changes are handled first and other generated declarations are still skipped for default-arg handling.
- Add a regression unit test `test_diff_generated_constant_value_change_is_api_break` in `tests/test_source_abi.py` to assert that a generated `constexpr` value change produces `ChangeKind.CONSTEXPR_VALUE_CHANGED` with the expected old/new values.

### Testing
- Ran the targeted tests: `pytest -q tests/test_source_abi.py::test_diff_generated_constant_value_change_is_api_break tests/test_source_abi.py::test_diff_generated_constant_removed_detected tests/test_source_abi.py::test_diff_generated_header_changed tests/test_source_abi.py::test_diff_constexpr_value_changed` and all 4 passed.
- Ran the package-level tests: `pytest -q tests/test_source_abi.py tests/test_source_extractors.py` and observed `70 passed, 1 skipped`.
- Running the full test suite `pytest -q` aborts during collection due to a missing optional test dependency (`hypothesis`) in the environment, which is unrelated to this fix.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a29d65ec61c832ba0fbcda52ff727ba)

### DIFF
--- a/abicheck/buildsource/source_diff.py
+++ b/abicheck/buildsource/source_diff.py
@@ -263,6 +263,11 @@ def _diff_generated(old: SourceAbiSurface, new: SourceAbiSurface) -> list[Change
         new_b = _by_identity(new_bucket)
         for key in sorted(set(old_b) & set(new_b)):
             ov, nv = old_b[key], new_b[key]
+            # A generated constexpr value change is still a baked-in public
+            # constant change, so keep the stronger constexpr_value_changed
+            # finding instead of downgrading it to generated_header_changed.
+            if nv.kind == "constexpr":
+                continue
             if _is_generated(nv) and _entity_changed(ov, nv):
                 name = nv.qualified_name
                 changes.append(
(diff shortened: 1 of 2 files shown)


napetrov (commented):
Reviewed after rebase. Generated constexpr/source ABI classification remains breaking where expected; local proof: ruff on changed files, pytest tests/test_source_abi.py (42 passed).

PR #376, opened 2026-06-13
fix(evidence): sanitize clang call graph replay args

napetrov:
### Motivation
- The Clang call-graph extractor previously appended each compile unit's raw `argv` into a `clang++` invocation, which allowed attacker-controlled compile DB entries to inject compiler options (e.g. `-Xclang -load`, `-fplugin=...`) that can load and execute arbitrary shared libraries during extraction.
- The intent is to preserve AST-parsing functionality while eliminating the ability for untrusted build evidence to trigger compiler-side effects.

### Description
- Rebuilds the Clang parse invocation from normalized/allowlisted fields instead of replaying raw `argv`: added safe replay configuration and language mapping in `abicheck/evidence/call_graph.py` (`_SAFE_REPLAY_FLAG_PREFIXES`, `_LANGUAGE_TO_CLANG_X`).
- Introduced helpers to construct sanitized parse args: `_safe_replay_flags_from_context`, `_safe_clang_args_from_argv`, and `_safe_clang_args_from_compile_unit` in `abicheck/evidence/call_graph.py`.
- Reworked `ClangCallGraphExtractor` so `extract_from_args()` sanitizes input via `_safe_clang_args_from_argv` and added `_extract_from_safe_args()` which actually shells out to `clang++`, and `extract_from_build()` now uses fields from each normalized `CompileUnit` (via `_safe_clang_args_from_compile_unit`) rather than `cu.argv`.
- Added regression tests in `tests/test_call_graph.py` that assert malicious plugin-loading flags are not replayed and that necessary parse flags (includes, defines, language, standard, ABI-relevant flags) are preserved for both raw-argv and compile-unit paths.

### Testing
- Ran linter/type checks: `ruff check abicheck/evidence/call_graph.py tests/test_call_graph.py` (fixed style issues) and `mypy abicheck/evidence/call_graph.py`, both succeeded.
- Ran the focused unit tests: `pytest -q tests/test_call_graph.py` — `28 passed`.
- Attempted full test run `pytest -q` but collection failed due to a missing dev dependency (`hypothesis`), so the full suite was not executed in this environment (collection error from property-test modules).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2a2d4dcb44832ba7db44c67cc3752d)

### DIFF
--- a/abicheck/buildsource/call_graph.py
+++ b/abicheck/buildsource/call_graph.py
@@ -37,13 +37,18 @@ from __future__ import annotations
 import json
 import shutil
 import subprocess  # noqa: S404 - call-graph extraction shells out to clang (never shell=True)
+from collections.abc import Mapping
 from dataclasses import dataclass, field
+from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
+from ..build_context import _extract_flags
+from .adapters.base import source_from_argv
 from .source_graph import CONF_HIGH, CONF_REDUCED, CONF_UNKNOWN, GraphEdge, GraphNode
 
 if TYPE_CHECKING:
     from .build_evidence import BuildEvidence
+    from .build_evidence import CompileUnit as BuildEvidenceCompileUnit
     from .source_graph import SourceGraphSummary
 
 # ── call-edge labels (ADR-031 D4) ───────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


napetrov (commented):
Reviewed after rebase. Raw compile argv replay is avoided; updated tests for current buildsource module path; local proof: ruff on changed files, pytest tests/test_call_graph.py (28 passed).

PR #394, opened 2026-06-14
perf: track ICU/OpenSSL versioned-symbol churn + linearize affected-symbol enrichment (was O(n²))

napetrov:
## Summary

Started as a perf-**tracking** gap assessment; the new tracking immediately surfaced a real **O(n²)** bug, now fixed.

## 1. New tracking — `versioned_rename_churn` scenario
Reproduces field-eval **P08 ICU 75→78** (16k changes): every export carries a version token (`u_strlen_75` → `u_strlen_78`), so `2×n` removed/added churn + the `versioned_symbol_scheme` collapse pass — a path no prior scenario reached (`rename_churn` uses disjoint names; the ELF fingerprint cap short-circuits dense churn). Profiling it fixed a per-finding `diff_namespaces._segments` re-tokenization (plain-name fast path).

## 2. Real bug found via the sweep — quadratic affected-symbol enrichment (fix #8)
Sweeping `typedef`/`union`/`vtable` churn **across sizes** (the old perf table only sampled n=500, so no exponent was ever computed and the rows were mislabelled "linear") exposed a genuine **≈O(n²)**:

```python
# diff_filtering._build_type_to_funcs / _build_type_embed_index
for func in old_pub:               # O(functions)
    for tname in affected_types:   # O(types)
        if any(tname in ft for ft in refs):   # substring scan
```

O(types × functions × refs) — quadratic when many distinct types change (header refactor / versioned upgrade). **Fix:** one Aho-Corasick `_SubstringMatcher` over affected type names, built once, shared by both builders; each ref matched in O(len) with **byte-for-byte identical** substring semantics (cross-token matches like `Type_5` in `Type_50` preserved).

### Measured (n=4000)
| scenario | before | after | speedup |
|---|---|---|---|
| typedef_churn | 6.30 s | 0.73 s | 8.7× |
| union_churn | 9.15 s | 1.10 s | 8.3× |
| vtable_churn | 7.79 s | 1.03 s | 7.6× |

All three drop from tail exponent ≈1.85 → **~1.0 (linear)**. `type_churn`'s 1.67 residual is the separate, documented opaque-filter path — unchanged.

## Validation
- `_SubstringMatcher` fuzzed 1500 random cases against naive `{n for n in needles if n in hay}` — identical.
- Full fast suite **10,313 passed**; FP-rate gate **0/0**; `ruff` ✅, `mypy` ✅, AI-readiness 0 errors.
- Docs: corrected the scaling table + coverage row, added fix #8; perf workflow PR-paths gain `diff_namespaces.py` / `versioned_symbol_scheme.py` (`diff_filtering.py` already matched `diff_*.py`).

## Out of scope (assessment follow-ups, not in this PR)
- Synthetic L3/L4/L5 build-source guard (currently only the live `eval/scaling.py` lane).
- Changed-scope + per-TU cache cold/warm-ratio tracking.
- DWARF/PE/PDB parser benchmarks.

https://claude.ai/code/session_01N8VmmxBPVPsBqa7Ypi6zma

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -37,6 +37,8 @@ on:
       - "abicheck/post_processing.py"
       - "abicheck/demangle.py"
       - "abicheck/binary_fingerprint.py"
+      - "abicheck/diff_namespaces.py"
+      - "abicheck/versioned_symbol_scheme.py"
       - "abicheck/surface.py"
       - "abicheck/suppression.py"
       - "abicheck/html_report.py"
--- a/abicheck/diff_filtering.py
+++ b/abicheck/diff_filtering.py
@@ -883,6 +883,72 @@ def _has_public_pointer_factory(
     return False
 
 
+def _opaque_usage_index(
+    candidates: set[str],
+    snap: AbiSnapshot,
+    bare_re_cache: dict[str, re.Pattern[str]],
+    factory_re_cache: dict[str, re.Pattern[str]],
+) -> tuple[set[str], set[str]]:
+    """Single pass over the public surface → ``(used_by_value, has_pointer_factory)``.
+
+    ``_filter_opaque_size_changes`` previously called ``_is_pointer_only_type`` and
+    ``_has_public_pointer_factory`` *per candidate*, each rescanning every public
+    function/variable with a word-boundary regex — O(candidates × functions) with
+    a regex per pair (``type_churn`` n=4000: ~3.2 M regex searches). This walks the
+    surface once and uses an Aho-Corasick prefilter (:class:`_SubstringMatcher`)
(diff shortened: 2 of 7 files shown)


napetrov:
Addressed the Codex P2 (`scripts/benchmark_scaling.py:898`) in `f00d622`: confirmed `_run_compare` left `collapse_versioned_symbols=False`, so the scenario only emitted the advisory and never the suppression/reclassification branch. Added `_run_compare_collapse` and registered `versioned_rename_churn` with it, so detection **and** collapse are now timed over the `2×n` churn set (still linear, ~1.1). The unit test now asserts the registered run is the collapse variant and that the collapse branch actually folds the churn away.

---

PR #399, opened 2026-06-14
feat: intra-version cross-source validation engine (ADR-035 D4, G19.2)

napetrov:
## Phase 2 of ADR-035 / G19 — Cross-source validation engine

Implements **G19.2**: a new `buildsource/crosscheck.py` engine that consumes **one** merged `AbiSnapshot` and diffs its evidence sources *against each other within a single version* (no baseline compare), surfacing a class of "bad ABI hygiene" findings that are only visible when the binary export table, the public-header AST, the build flags, and header provenance are checked for **mutual consistency**.

### New `ChangeKind`s (ADR-035 D4 table)

| Check | Inputs | Tier |
|---|---|---|
| `exported_not_public` | binary exports ↔ L2 header decls | RISK |
| `public_not_exported` | L2 header decls ↔ binary exports | RISK |
| `header_build_context_mismatch` | L2 header context ↔ L3 build flags | API_BREAK |
| `private_header_leak` | public API ↔ private-header provenance | RISK |

Per ADR-035 D1/D4 (the authority rule) **none are `BREAKING` on their own** — an artifact diff still proves a shipped break. They default to RISK/API_BREAK and stay advisory/suppressible until a check earns its FP-rate-gate corpus.

### Design highlights

- **Coverage honesty.** Each check emits a coverage row; when its required evidence is absent (e.g. no public-header provenance, no L3 build evidence) it is reported **skipped** — never counted as clean and **never emitting a false positive**. An ELF-only snapshot simply reports every check as skipped.
- **Provider-agreement matrix (§6.8).** Every run check records which evidence sources (`binary_exports`, `public_header_ast`, `build_config`, `source_index`) corroborate it, driving the confidence tag.
- **Reuses existing machinery** — leans on the ADR-024/ADR-015 `ScopeOrigin` provenance already on every declaration (`EXPORT_ONLY` *is* `exported_not_public`), the `BuildEvidence.abi_relevant` flag set, and the L5 source graph when present.
- `public_not_exported` is intentionally narrow: only declarations with a real export obligation (default-visibility, non-inline, non-template, non-pure-virtual, mangle-bearing functions / extern data) are compared, so a healthy header-only API stays quiet.

### Tests & gates

- `tests/test_crosscheck.py` — 27 cases: positive/negative per check, the coverage-honesty contract, provider matrix, config-disable, and the per-check cap. All pass.
- Full fast suite: **10485 passed**. `mypy` clean (160 files). `ruff check` clean. AI-readiness 0 errors. FP-rate gate 0/0. Detector spec regenerated & in sync (250 kinds).

### Docs / tracking

- `concepts/build-source-data.md` — new intra-version cross-source findings section.
- Headline `ChangeKind` count 246 → 250 across docs (doc-count-sync gate).
- `usecase-registry.yaml`: `UC-CHANGE-crosscheck-hygiene` → **partial** with module/test evidence; G19 plan Phase 2 marked **DONE**.

### Not in this PR (follow-ups, per the plan)

`odr_type_variant` and `public_to_internal_dependency` checks; wiring the engine into the Phase-3 `scan`/`audit` orchestrator + `crosschecks:` severity config; FP-rate-gate corpus cases before any check is promoted to gate.

https://claude.ai/code/session_016BRYWg3YvkBEFRQkXorxeu

---
_Generated by [Claude Code](https://claude.ai/code/session_016BRYWg3YvkBEFRQkXorxeu)_


napetrov:
Thanks @chatgpt-codex-connector — all four review points were valid and are addressed in 740f3a1, each with a regression test:

- **`exported_not_public` / ELF_ONLY visibility** — confirmed against `provenance.tag_provenance`: `EXPORT_ONLY` origin is only set for `Visibility.ELF_ONLY` symbols, so the check now gates on `origin == EXPORT_ONLY` alone (dropped the `PUBLIC` visibility requirement that masked every real export-only symbol).
- **`public_not_exported` / static helpers** — `_has_export_obligation` now excludes `fn.is_static` (internal linkage, no dynamic symbol).
- **Mach-O export normalization** — `_exported_symbol_names` now strips the platform's leading underscore for Mach-O exports, matching how the dumper normalizes `Function.mangled` (`_foo` → `foo`, `__Z…` → `_Z…`).
- **PIMPL false positive** — `_private_type_names` now subtracts any type that also has a `PUBLIC_HEADER`/`GENERATED` declaration, so the opaque-handle forward-decl + private-definition pattern no longer flags `Impl *` public APIs.

---

napetrov:
Second review round — all three new points were valid and are addressed in bad5fe2, each with a regression test:

- **Non-default ELF version aliases** — `_exported_symbol_names` now restricts the ELF set to `s.is_default`, so a symbol exposed only as a non-default alias (`foo@LIB_1`) no longer satisfies an unversioned header declaration of `foo()`.
- **Operators vs. template heuristic** — replaced `"<" in name` with `_looks_templated()`, which keys off the token before the first `<`; `operator<` / `operator<<` / `operator<=>` are no longer treated as templates, so a missing exported operator is still flagged.
- **Basename alias collision** — `_private_type_names` now collects public-type tokens (canonical + trailing segment) and skips any private canonical/alias that collides, so a public `Impl` alongside a private `detail::Impl` no longer mis-flags a public `Impl *` signature (only an explicit `detail::Impl` reference leaks).

---

napetrov:
Third review round — all three points were valid and are addressed in 91a6a94, each with a regression test:

- **Visibility-gated obligation** — confirmed `_CastxmlParser._visibility` derives `Visibility.PUBLIC` from the export table, so a declared-public-but-unexported symbol is `HIDDEN`/`ELF_ONLY` before the helper runs. Dropped the `visibility == PUBLIC` gate from `_has_export_obligation`/`_var_has_export_obligation`; the obligation now rests on the public-header source declaration (origin/access/static/inline/template/mangle).
- **Private-header exports** — `exported_not_public` now flags an exported symbol declared only in a private or non-public generated header (origin `PRIVATE_HEADER`/`GENERATED`), gated on actual export-table membership, in addition to `EXPORT_ONLY`.
- **GENERATED ≠ public** — provenance maps a public generated header to `PUBLIC_HEADER` first, so a bare `GENERATED` origin is a non-public generated header. Removed `GENERATED` from the public-type set and added it to the leak-candidate origins, so a public API exposing e.g. a `build/generated/internal_config.h` type is now reported as `private_header_leak`.

---

napetrov:
Fourth review round — both points valid and addressed in e63343a, each with a regression test:

- **Export-only symbols with no decl object** — confirmed the with-headers path sets `functions=parser.parse_functions()` (header-declared only), so a symbol that lives solely in the export table has no `Function` to iterate. `exported_not_public` is now **driven by the export table**: it compares the export set against public-declared symbols and flags any export no public header declares (enriching the message when a private/generated-header decl exists). Itanium ctor/dtor exports are skipped to avoid the unmangled-member false positive.
- **C++ member mangle fallback** — `_has_export_obligation` now requires a real mangled symbol (`_looks_mangled`: `_Z…`/`?…`, or any non-empty name for `extern "C"`/data). A castxml ctor fallback (`mangled == "Widget"`) no longer gets compared against the binary's `_ZN6WidgetC1Ev` and mis-flagged.

---

napetrov:
Fifth review round — both points valid and addressed in 5d99757, each with a regression test:

- **C++ variable bare name** — `_candidate_symbols` is now keyed on whether `mangled` is a real mangling (`_Z…`/`?…`). A C++ namespace/global variable contributes only its mangled symbol (`_Z1g`), not the spelling `g`, so an unrelated accidental export named `g` is no longer treated as documented and is reported. Only un-mangled decls (C / extern-C / C data) contribute their bare name.
- **MSVC structors** — `_is_cxx_structor` now recognizes MSVC ctor (`??0…`) / dtor (`??1…`) manglings in addition to Itanium `_ZN…C/D…E`, so exported constructors/destructors on PE/MSVC public libraries are skipped instead of false-flagged.

---

napetrov:
Sixth review round — both points valid (confirmed `is_abi_relevant_elf_symbol` keeps own-class `_ZTV`/`_ZTI`/`_ZTS`), addressed in d822e12 with regression tests:

- **RTTI/vtable exports** — `exported_not_public` now exempts compiler-generated C++ class artifacts (Itanium `_ZTV`/`_ZTI`/`_ZTS`/`_ZTT`/`_ZTh`/`_ZTv`/`_ZTc`, MSVC `??_`) in addition to ctors/dtors. The owning class is a `RecordType`, never in the documented symbol set, so these were false-flagged on any public polymorphic class. `_is_cxx_structor` → `_is_cxx_implementation_symbol`.
- **Parsed const constants** — `public_not_exported` no longer requires `Variable.value` to exclude a constant. castxml stores the initializer in `snapshot.constants`, leaving `Variable.value` `None`, so a public `const int k = 42;` was wrongly flagged. A const/constexpr namespace-scope variable has internal linkage in C++ and emits no symbol, so the exclusion now keys off `is_const` alone.

---

PR #421, opened 2026-06-15
feat: add clang as an alternative L2 header-scan backend

napetrov:
## Why

L2 (header AST) had a single producer — **castxml** — so `dump --headers`, `compare -H`, and `scan -H` **hard-failed** on the many dev/CI images that ship clang but not castxml, taking header-aware public-surface scoping and the ADR-035 D4 cross-source checks down with them. This is **P1** from the UXL field run (`validation/uxl-scan-levels-timing-2026-06.md`) and the long-planned ADR-003 extension *"clang as an alternative L2 frontend"*.

## What

A `clang -ast-dump=json` → `AbiSnapshot` parser, **`abicheck/dumper_clang.py`** (`_ClangAstParser`), a sibling to `dumper_castxml._CastxmlParser` exposing the **identical** `parse_functions` / `parse_variables` / `parse_types` / `parse_enums` / `parse_typedefs` / `parse_constants` surface, selected behind a backend knob:

- `dumper.dump(..., header_backend=...)` threaded through `service.run_dump` / `resolve_input` and the per-format `_dump_elf` / `_dump_macho` / `_dump_pe` builders via a single `_header_ast_parser` factory both frontends sit behind;
- **`--header-backend auto|castxml|clang`** on the `dump` and `compare` CLIs;
- **`ABICHECK_HEADER_BACKEND`** env var for a global default (flip a clang-only CI image with no flag);
- `auto` prefers castxml (the schema reference), then falls back to clang when only clang is on `PATH` — closing the P1 hard-fail.

The clang-JSON cache lives beside the castxml-XML cache (`~/.cache/abi_check/clang/*.json`), keyed on the backend so the two never collide.

## Coverage trade-off

clang's JSON AST is **syntactic** — it does not compute record layout — so a clang-derived `RecordType` carries field names/types, bases, and access but **not** `size_bits` / `offset_bits` / vtable slots (left `None`/empty; the layout detectors skip an unknown-vs-unknown comparison and **DWARF (L1) stays the layout authority**). Everything the source-API and public-surface-scoping detectors need — signatures, `noexcept`/`const`/`explicit`, enum values, typedef targets, public constant values — is produced, which is exactly what the D4 cross-source checks consume.

## Tests

- `tests/test_dumper_clang.py` — pure-parser unit coverage (signatures, qualifiers, member access, enum auto-increment, public-header-scoped constants, bitfields, bases, forward-decl skip, sticky `loc.file`) plus backend-resolver / command-builder tests. **28 tests.**
- `tests/test_clang_header_backend_integration.py` — a **live** `clang -ast-dump=json` end-to-end test over a real ELF `.so` that asserts a header-aware snapshot and clang↔castxml parity on the public surface. Deliberately *not* `integration`-marked (that gate requires castxml — the opposite of the case under test); each test self-skips on its own tool requirement.

The first live test **passes against real clang** in this environment (castxml absent).

## Docs

- ADR-003 extension marked **implemented** with the landed design + next steps.
- `docs/user-guide/tool-modes.md` requirements row, root `CLAUDE.md` module map, and the UXL field-run P1 status updated.

## Verification

- `ruff check abicheck/ tests/` ✅
- `mypy abicheck/` ✅ (0 errors)
- `python scripts/check_ai_readiness.py` ✅ (0 errors; one pre-existing soft size WARN on `dumper.py`, well under the 2000-line hard cap)
- Full fast unit lane ✅ (11k+ passed)

https://claude.ai/code/session_0196E5VyDpsG68Q7RVGLXYZ4

---
_Generated by [Claude Code](https://claude.ai/code/session_0196E5VyDpsG68Q7RVGLXYZ4)_

## Release Notes

* **New Features**
  * Added a clang-based header AST backend alongside castxml to produce header-aware snapshots for `dump` and `compare`.
  * Added `--header-backend {auto,castxml,clang}` with `ABICHECK_HEADER_BACKEND` defaulting support (auto prefers castxml, then falls back to clang).

* **Bug Fixes**
  * Improved header-AST cache separation to avoid backend collisions.

* **Documentation**
  * Updated user guide requirements and selection behavior; refreshed scope-confidence fallback wording.

* **Tests**
  * Added ELF/Linux integration coverage and extensive clang backend unit tests, including castxml↔clang public-surface parity.


napetrov:
Thanks for the thorough reviews (@chatgpt-codex-connector, @coderabbitai). Pushed `b8ae32d` addressing the substantive findings.

**Fixed**
- **`header_backend` ignored for PE/Mach-O** (Codex P2; CodeRabbit, service.py + cli.py) — now threaded through `run_dump` → `_dump_pe`/`_dump_macho` → `_try_header_scoped_dump`, and the CLI `_handle_non_elf_dump` → `_dump_native_binary` → `run_dump`.
- **Provenance lost when clang omits `loc.line`** (Codex P2) — `_source_location` keeps the declaring header when the line is sticky, so a `static constexpr` member on a one-line `struct` is no longer dropped from `constants`.
- **Walk descended into function bodies** — now stops at function decls, so block-scope locals can't leak into `variables`/`constants`.
- **`extern "C"` flag discarded** — threaded into `_Decl` and used as the authoritative `is_extern_c` (mangled==name kept as the C-mode fallback).
- **`noexcept(false)`/`noexcept(0)`** now parse as *throwing*.
- **Folded constant values** (`1<<3`) — read off the `ConstantExpr` wrapper before unwrapping, recovering enum values and bitfield widths.
- **Mach-O visibility** — clang's underscore-prefixed mangled names now match the underscore-stripped export set (this was the macOS unit-lane failure).
- Docs: `--header-backend auto` clarified as the default selector; `dump --headers` → `dump -H`.

**Intentionally not changed (parity / scope), happy to revisit if you disagree:**
- **Qualified vs. leaf model names** — the castxml parser emits *unqualified* `Function.name`/record/enum/typedef names (functions are keyed by mangled name); the clang backend matches that on purpose so the two stay a parity oracle. Qualifying only clang would diverge the snapshots.
- **`source_header`/`origin` in the parser** — these are populated by `apply_provenance()` in `dump()` for *both* backends identically (the castxml parser doesn't set them either), so clang already gets the same scoping data.
- **`__future__` import after the docstring** — matches the existing convention (`dumper_castxml.py`, `source_extractors/clang.py`); a module docstring legally precedes the future import.
- **Anonymous `typedef struct {…} T;` field capture** — left as the already-documented partial-coverage limitation; clang's JSON lacks castxml's clean record→typedef id link, and symbols are still captured. Tracked for a follow-up.
- **Integration-test markers / fixture location** — the build-and-dump test is deliberately Linux-gated and unmarked so it runs in the Linux unit lane (real-clang coverage that protects the 95% gate) while self-skipping elsewhere; the `integration` marker gates on **castxml**, which is exactly the host the clang backend exists to support.

CI re-running now.

---

PR #429, opened 2026-06-16
feat: per-side L2 header backend on compare/deep-compare

napetrov:
## What

Adds `--old-header-backend` / `--new-header-backend` to `compare` and `deep-compare`, each defaulting to (inheriting) `--header-backend`. This is the **backend mirror** of the existing `--old-header` / `--new-header` asymmetric-input flags.

## Why

Header-scoped scans are the only way abicheck separates *public source API* from *private/internal* surface, but the castxml L2 path can abort on a stock host toolchain (sized-float `_FloatN`, GCC `__assume__`, newer C++ constructs) — see gap **G16**. The clang backend parses against the host toolchain and is the structural recovery path.

Until now `--header-backend` applied to **both sides uniformly**. When an *old* release parses cleanly on castxml (the schema reference) but a *new* release uses constructs castxml's bundled Clang chokes on, you were forced to move **both** sides to clang. Per-side selection lets the new side use clang while the old side keeps castxml.

Each side already resolves and G16-auto-falls-back independently (separate `_header_ast_parser` calls), so no fallback rewrite was needed.

## Changes

- `cli_resolve._resolve_compare_snapshots` — take `old_header_backend` / `new_header_backend` overrides (`None` = inherit `header_backend`), thread each into its side's `_resolve_input`.
- `cli.compare_cmd` and `cli_max` deep-compare — two new options + wiring.
- `tests/test_cli_new_features.py::TestPerSideHeaderBackend` — asserts independent per-side routing, inheritance of the global default, and one-sided override.
- Docs — `tool-modes.md` backend note + `usecase-registry.yaml` G16 entry.

## Not in scope

- Cross-frontend **corroboration** (run both frontends on the *same* side, agreement matrix) — a new provider axis.
- Coverage **union** — blocked on the G4 libclang extractor.

## Validation

- New tests pass (3); broader CLI/compare/deep slice green (1061 passed).
- `mypy abicheck/` clean (191 files); `ruff check` clean; AI-readiness gate 0 errors.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QWSt5jaw1Q4qtUSHWqBk8z)_

## Summary by CodeRabbit

* **New Features**
  * Added `--old-header-backend` and `--new-header-backend` to `compare` and `deep-compare`, letting you choose header parsing independently for old vs new inputs (each defaults to `--header-backend` when omitted).
  * Crosscheck and rendered scan layers now expose additional boundary-integrity counters derived from coverage.

* **Documentation**
  * Updated the user guide and use-case notes to describe the per-side header backend options and expected default/override behavior.

* **Tests**
  * Added CLI tests for correct per-side routing and inheritance, plus crosscheck/integrity tests validating counter propagation and corroboration logic.


napetrov:
Thanks for the review.

- **`None` reachable_declarations guard (major)** — valid, fixed in 746435d (`len(surface.reachable_declarations or [])`).
- **crosscheck.py > 1500 lines (nitpick)** — the >1500 threshold is an advisory WARN in this repo (the hard cap is 2000; `scripts/check_ai_readiness.py` passes), and this PR only nudged the file ~19 lines over by adding the D4 integrity-counter plumbing. A check/helper module split is a worthwhile refactor but a large, cross-cutting change that's out of scope for this focused plumbing PR — doing it here would hurt, not help, change isolation. I've trimmed the new comments to minimise the delta; tracking the split as separate follow-up work.

---

PR #440, opened 2026-06-18
fix(scan): clarify L2 backend hint and advise when a deep level lacks an L3 compile DB

napetrov:
## What & why

Two UX-clarity fixes for the `scan` evidence-level flow, surfaced while planning a real-world oneAPI (oneTBB/oneDNN/oneDAL/oneCCL) scan validation. Both address the same gap: when a user asks for a deep evidence level but the inputs can't deliver it, the tool degraded silently or pointed at the wrong remedy.

### 1. Stale L2-header hint
The L2-header coverage row (and the `--estimate` cost row) said **"needs castxml"**, but the clang L2 backend now covers header parsing (ADR-003 extension). Updated both `cli_scan` and `service_scan` to read **"needs castxml or clang"** so a clang-only host isn't told it's missing a dependency it doesn't need.

### 2. Pointed advisory when a deep `--source-method` has no L3 compile DB
A deep `--source-method` (`s1/s2/s4/s5/s6` → `collect_mode != "off"`) needs an L3 compile database. Over a pristine source checkout with no `compile_commands.json`, the L3/L4/L5 layers are skipped and the coverage rows just say `not_collected` — leaving the user to infer *why*.

Now, when the user actually supplied a source input (`--sources`/`--build-info`) yet L3 still came back empty, the scan emits a pointed advisory naming the requested level and the exact remedy:

```
note: requested source-method s5 (depth source) needs an L3 compile database, but
none was found — L3/L4/L5 were skipped. Provide one with --build-info/--compile-db
(a compile_commands.json or build dir), or a trusted --build-config plus
--allow-build-query to generate it.
```

The advisory rides the structured result, so it appears in the text report **and** the `advisories` array under `--format json` (unlike the pre-existing stderr warning, which would pollute JSON stdout). It's gated on an actual source input so a plain binary-only scan at the default deep mode isn't nagged.

## Tests
Adds four tests to `tests/test_cli_scan.py`:
- advisory fires for a deep method over a DB-less source tree;
- stays silent when a compile DB is present (seeded);
- stays silent when no source input is given (binary-only);
- the L2 hint mentions clang.

All `tests/test_cli_scan.py` (44) + scan/service suites (364) pass; `ruff`, `mypy`, and the AI-readiness gate are clean.

https://claude.ai/code/session_01WmCKUCjSMr831W6pETGmfd

---
_Generated by [Claude Code](https://claude.ai/code/session_01WmCKUCjSMr831W6pETGmfd)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified that “public-header AST” evidence may require **castxml or clang** when headers are involved.
  * Improved deep-scan advisories when L3 evidence is missing, including clearer “honest level-vs-evidence” messaging.
  * Refined build-query auto-enabling to require **explicit deep-level consent** plus a trusted configuration.
* **New Features**
  * Added primary **`--config`** flag for selecting project build configuration; **`--build-config`** remains as a deprecated alias.
* **Documentation**
  * Updated scan-level requirements, input-missing behavior, and trusted-config/deeper-mode guidance.
* **Tests**
  * Extended end-to-end and unit coverage for deep scans, advisories, and config aliasing.


napetrov:
Thanks for the reviews — addressed in `bac1c44`:

- **Codex P2 (security):** level-implies-query no longer fires for the default mode preset. The auto-enable of `build.query` is now gated on an **explicit** `--source-method`/`--depth` (`level_explicit`), so a `--config` passed purely for project settings never silently runs a subprocess. Added a regression test for the default (`--audit`) case.
- **Advisory flag name:** the missing-L3 advisory now says `--config` (not `--build-config`).
- **Docs:** corrected the unseeded-`s5` note — it falls back to a headers-only public-API replay with an advisory, not a full per-TU replay.
- **Test style:** `test_config_alias_on_dump_and_scan` is now parametrized.

One suggestion I **kept as-is**: `--depth symbols` in the use-case table. It's a valid user depth (`scan_levels.USER_DEPTHS` / `cli_params.DepthParam` accept `{symbols,headers,build,source,full}`) and is the correct *precise* level for a binary-only gate — `symbols` = L0/L1 only, whereas `headers` would add the L2 AST. So `s0` / `--depth symbols` is intentional for that row.

---

PR #463, opened 2026-06-28
Fix AST cache fallback for read-only home

napetrov:
## Summary
- Respect XDG_CACHE_HOME for POSIX AST cache paths.
- Fall back to the system temp directory when the normal AST cache directory is unavailable, instead of failing header-mode dumps before castxml/clang can run.
- Keep castxml dump successful if writing the cache copy fails after AST extraction.

## Proof
- ../.venv/bin/python -m pytest tests/test_dumper_unit.py::TestCachePath tests/test_dumper_coverage.py::TestCastxmlDumpBranches tests/test_unknown_type_warning.py -q
- HOME=/proc XDG_CACHE_HOME=/tmp/abicheck-xdg-test ../.venv/bin/python - <<'PY' ... _cache_path('k') ... PY => /tmp/abicheck-xdg-test/abi_check/castxml/k.xml
- ../.venv/bin/python -m pytest tests/test_abicc_scenario_parity.py::TestFunctionPointerChanges::test_std_function_field_signature_changed_after_rename_is_breaking -q
- CLI smoke: compare same generated libpvxs_like.so with same headers under HOME=/proc and PATH=/usr/bin:/bin => rc 0, verdict NO_CHANGE, changes 0; stderr showed cache fallback to /tmp, not fatal.

## Gap
- I cloned pvxs and EPICS Base to try a full libpvxs.so validation, but EPICS Base build was still in modules/CA after a long run and I stopped it before pvxs artifacts existed. The cache failure mode is covered by the CLI smoke above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AST cache path resolution with per-backend filenames and standard cache locations, with automatic fallback to a temporary cache when the preferred directory can’t be used.
  * Made cache persistence best-effort for both replay and debuginfo downloads, handling filesystem errors via warnings and safe cleanup instead of aborting.
* **Documentation**
  * Clarified snapshot cache location to reflect standard user cache directories.
* **Tests**
  * Added regression tests for POSIX cache path selection, temp-cache fallback when home is unavailable/unwritable, and best-effort behavior for unwritable cache directories.

### DIFF
--- a/abicheck/buildsource/source_replay.py
+++ b/abicheck/buildsource/source_replay.py
@@ -764,13 +764,17 @@ class SourceAbiCache:
             if digest is not None:
                 deps[dep_path] = digest
         entry = {"tu": tu.to_dict(), "deps": deps}
-        self.cache_dir.mkdir(parents=True, exist_ok=True)
         tmp = self._path(key).with_suffix(".json.tmp")
-        tmp.write_text(
-            json.dumps(entry, indent=2, sort_keys=True) + "\n", encoding="utf-8"
-        )
-        # Atomic publish so a concurrent reader never sees a partial file.
-        tmp.replace(self._path(key))
+        try:
+            self.cache_dir.mkdir(parents=True, exist_ok=True)
+            tmp.write_text(
+                json.dumps(entry, indent=2, sort_keys=True) + "\n", encoding="utf-8"
+            )
+            # Atomic publish so a concurrent reader never sees a partial file.
+            tmp.replace(self._path(key))
+        except OSError:
+            # Cache writes are best-effort; replay already produced the facts.
+            tmp.unlink(missing_ok=True)
 
 
 def _dep_digest(path: str, memo: dict[str, str | None] | None = None) -> str | None:
--- a/abicheck/debug_resolver.py
+++ b/abicheck/debug_resolver.py
@@ -532,7 +532,14 @@ class DebuginfodResolver:
         xdg = os.environ.get("XDG_CACHE_HOME", "")
         if xdg:
             return Path(xdg) / "abicheck" / "debuginfod"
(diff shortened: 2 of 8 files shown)


napetrov:
Disk I/O audit follow-up pushed in 49afd5f6.\n\nChecked broad non-explicit disk writes via rg and tests. Internal caches are now best-effort/fail-open under read-only HOME or unwritable cache dirs:\n- AST/castxml cache: XDG_CACHE_HOME, temp fallback, cache-copy warning only.\n- L4 SourceAbiCache: put() no longer aborts replay if cache dir/write/replace fails.\n- debuginfod cache: no-home default falls back to temp; downloaded artifact falls back to temp cache if configured cache is unwritable.\n- snapshot cache and BuildEvidenceCache were already fail-open; snapshot docstring updated for XDG behavior.\n\nIntentionally still hard-fail: explicit user outputs/targets such as --output, --output-dir, baseline/evidence output paths, and package extraction target dirs. Those are user-requested writes, not internal cache.\n\nLocal proof:\n- pytest targeted disk/cache suite: 48 passed\n- git diff --check: clean\n- HOME=/proc header-mode CLI smoke on pvxs-like shared library: rc=0, verdict=NO_CHANGE, changes=0\n\nRemaining caveat: full real pvxs lib build was attempted earlier but EPICS Base build did not finish in the local time window, so this PR proves the reported environment class with a pvxs-shaped smoke rather than full pvxs artifacts.

napetrov:
Follow-up 2422299e fixes the CI readiness fallout from the audit change: moved AST cache path helper out of dumper.py so the file-size hard gate stays under 2000 lines, added SPDX on the new helper, and made the L4 cache smoke assert explicitly.\n\nLocal proof after this commit:\n- targeted disk/cache/readiness suite: 51 passed\n- git diff --check: clean

PR #464, opened 2026-06-28
Auto-query Make builds for scan evidence

napetrov:
## Summary
- auto-run a fixed Make dry-run query (`make -B -n -k`) when `scan --sources` needs L3 and no compile DB exists
- ingest Make dry-run transcripts through the existing reduced-confidence MakeAdapter, including partial evidence when make exits non-zero after printing compile commands
- update scan docs and build-source notes for the new default Make behavior

## Validation
- `python3 -m pytest tests/test_build_query_inference.py tests/test_make_adapter.py -q`
- `python3 -m ruff check abicheck/buildsource/build_query.py tests/test_build_query_inference.py`
- `git diff --check`
- pvxs no-compile-db proof: `scan --depth build --sources tmp/pvxs-scan/pvxs-1.5.1-build ...` produced `L3_build present: make, 110 compile units, 0 targets` with no `compile_commands.json` in that tree

## Risk
- Make dry-run can still execute recursive/`+` recipes on some Makefiles; this is now documented as part of the default source-query trust boundary.

## Summary by CodeRabbit

* **New Features**
  * When no compile database is available, inferred builds can now automatically collect GNU Make evidence using a fixed, side-effect-safe dry-run and transcript scraping.
* **Bug Fixes**
  * Make transcript evidence is ingested even when the dry-run returns non-zero, with partial-status reporting and the exit code noted.
  * Improved recognition of “entering/leaving directory” markers across common Make executable variants, including nested directory cases.
  * If a GNU Make launcher can’t be found, evidence is safely marked as skipped.
* **Documentation**
  * Updated build-source and scan-level guidance to reflect the reduced-confidence Make transcript evidence and the compile-db preference.
* **Tests**
  * Expanded coverage for Make dry-run invocation, non-zero handling, and launcher/transcript parsing scenarios.

PR #470, opened 2026-06-30
fix(scan): require explicit --allow-build-query for build.query

napetrov:
### Motivation
- Prevent execution of arbitrary commands from a configured `build.query` when an operator only supplies `--config` plus a pinned source level or depth, restoring the explicit `--allow-build-query` action ceiling. 
- Close a race where a trusted `--config` was treated as implicit consent to run queries, which allowed attacker-controlled repo/config to run arbitrary local commands in CI. 

### Description
- Remove the `level_explicit` probe and the auto-enable path that set `effective_allow_query = True` in `run_scan_core`, so the scan core now forwards the real `allow_build_query` flag unchanged to snapshot/build-source collection (`abicheck/cli_scan.py`).
- Ensure the inline build-source layer still treats auto-discovered `.abicheck.yml` as untrusted and preserves the `--allow-build-query` opt-in semantics (`abicheck/buildsource/inline.py` behavior unchanged; gating now only from explicit `--allow-build-query`).
- Update documentation to require explicit `--allow-build-query` when running `build.query` in the example and explanatory text (`docs/user-guide/scan-levels.md`).
- Add regression tests that assert an explicit `--config` plus a pinned `--source-method` or `--depth` does not execute the configured query and reports the skipped-query diagnostic (`tests/test_cli_scan.py`).

### Testing
- Ran the targeted tests: `pytest -q tests/test_cli_scan.py -k 'scan_config_query_requires_explicit_allow_build_query or scan_depth_config_query_requires_explicit_allow_build_query'` and both tests passed (`2 passed`).
- Ran the full scan CLI test suite: `pytest -q tests/test_cli_scan.py` and all tests passed (`51 passed`).
- Verified the docs example and test assertions reflect the new requirement that `build.query` execution always requires `--allow-build-query` (no automated doc check beyond test-driven examples were required).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a336abeb2e8832bb03de22276cfe11b)

### DIFF
--- a/abicheck/cli_scan.py
+++ b/abicheck/cli_scan.py
@@ -832,16 +832,6 @@ def scan_cmd(
             severities=severities,
             budget=budget,
             budget_s=budget_s,
-            # A concrete explicit level is what consents to level-implies-query
-            # auto-running build.query: a non-auto --source-method, or --depth ONLY
-            # when no --source-method is given (resolve_level gives --source-method
-            # precedence and ignores --depth otherwise, so `auto`+`--depth` resolves
-            # via auto/the preset, not the depth — it must not count as consent;
-            # Codex review).
-            level_explicit=(
-                source_method is not None and source_method != SourceMethod.AUTO.value
-            )
-            or (source_method is None and depth is not None),
         )
     except _BudgetOverflow as bo:
         click.echo(bo.message, err=True)
@@ -913,7 +903,6 @@ def run_scan_core(
     severities: dict[str, str],
     budget: str | None,
     budget_s: float | None,
-    level_explicit: bool = False,
 ) -> ScanCoreResult:
     """The shared scan orchestration (classify → always-on tier → level → compare).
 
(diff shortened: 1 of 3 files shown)


napetrov:
## ⚠️ Conflicts are resolvable, but this PR now contradicts `main` — needs a human decision

I looked at rebasing/merging `main` into this branch to clear the merge conflict. The **textual** conflicts (`abicheck/cli_scan.py`, `docs/user-guide/scan-levels.md`, `tests/test_cli_scan.py`) are straightforward to resolve. The problem is a **semantic** one that a clean merge silently hides:

Since this PR was opened, `main` landed the "zero-config source scanning" work (#456 *Zero-config source scanning: auto build-query…* and #464 *Auto-query Make builds for scan evidence*). That work **deliberately removed the `--allow-build-query` action ceiling** this PR is trying to enforce. From the current `abicheck/buildsource/inline.py` docstring:

> `allow_build_query` is accepted only for backward compatibility and is ignored — `--allow-build-query` is a deprecated no-op.

…and the inferred CMake/Bazel/Make query now "runs whenever `--sources` needs L3, since pointing abicheck at a source tree is itself the request to analyse it."

Consequences if this branch is merged after only resolving the conflict markers:

1. The diagnostic this PR's test asserts — `"build.query configured but --allow-build-query not set"` — **no longer exists on `main`** (it was deleted along with the old gate in `inline.py`), so `test_scan_config_query_requires_explicit_allow_build_query` **fails**.
2. More importantly, `main` still **executes** a trusted-`--config` `build.query` without the opt-in (the inline layer runs it regardless of `allow_build_query`), so the exact code-execution surface this PR set out to close **stays open** — the `cli_scan.py` change here no longer gates anything downstream.

In other words, removing the `level_explicit` auto-enable in `cli_scan.py` is necessary but **no longer sufficient** — the real gate now lives in `inline.py`/`build_query.py`, which `main` rewrote to *not* gate on `--allow-build-query`.

**This is a design-direction conflict, not a merge mechanic.** Resolving it properly means re-arming the `--allow-build-query` ceiling on top of `main`'s new inline build-query path (i.e. partially reversing the zero-config auto-query behaviour) — which is a deliberate product decision, not something to slip in via conflict resolution. The alternatives are to redesign this PR against the new `inline.py`, or to close it as superseded by the zero-config direction.

Holding off on pushing a merge here until a maintainer decides which way to go. (#473 is the same theme and has the same issue.)


---

napetrov:
Closing as **superseded** by the zero-config source-scanning direction on `main` (#456, #464), which intentionally made `--allow-build-query` a deprecated no-op and auto-runs the build query when `--sources` needs deeper evidence. The maintainer has chosen to keep that behaviour, so this PR's opt-in gate is no longer the intended design. See the analysis in the comment above if the explicit action ceiling is ever revisited.

---

PR #527, opened 2026-07-11
fix: unify benchmark_comparison.py tool-call timeouts to 90s

napetrov:
## Summary
- `scripts/benchmark_comparison.py` benchmarks abicheck vs ABICC vs abidiff across the `examples/` catalog (currently 170 cases, matching `examples/ground_truth.json`), tracking per-tool accuracy against expected verdicts, false positives/negatives, and elapsed time.
- Several tool lanes had inconsistent, hardcoded per-call timeouts: `abidiff`, `abicheck compat`, and `abicheck strict` were hardcoded at 60s, while ABICC (the slowest tool, prone to hanging) defaulted to 120s.
- Unified all of these under a single `DEFAULT_TIMEOUT = 90` constant, exposed via a new `--timeout` flag (default 90s), and lowered the `--abicc-timeout`/`--abicheck-full-timeout` defaults from 120s to 90s so a single hung example run is bounded consistently across the whole tool matrix.
- Verified the harness already iterates every `examples/case*` directory (170), so it structurally covers the full current example set with no changes needed there — confirmed by running the fast lane end-to-end in this environment (ABICC/abidiff/castxml aren't installed here, so those columns report `SKIP`, but the abicheck/compat/strict lanes ran against the complete catalog).

## Test plan
- [x] `python3 scripts/benchmark_comparison.py --help` — new `--timeout` flag present, defaults render as 90s
- [x] `python3 -c "import ast; ast.parse(open('scripts/benchmark_comparison.py').read())"` — syntax check
- [x] Ran `scripts/benchmark_comparison.py --skip-abicc --tools abicheck abicheck_compat abicheck_strict` against the full 170-case catalog to confirm coverage and that timeout plumbing doesn't break any tool call

---
_Generated by [Claude Code](https://claude.ai/code/session_015AhNEkeBDcQYzPKXFeadSo)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized benchmark timeouts and TIMEOUT reporting; added a shared per-tool `--timeout`.
  * Tightened header provenance handling in dump/compare/audit/app-compat flows.
  * Reduced spurious reachability/owner-change findings by improving graph/header and path matching.
  * Corrected castxml elaborated-type type-name unwrapping and strengthened public-surface seeding/never-filter behavior.
  * Improved scoring to report full-catalog coverage accuracy with FP/FN counts.
* **New Features**
  * Added frozen competitor results: `--freeze/--no-frozen` to reuse prior verdicts and enable merge-style reruns.
* **Documentation**
  * Updated benchmark documentation with full-catalog results and strict miss/FP/FN definitions.
* **Tests**
  * Added regression coverage for header provenance, owner-change logic, castxml elaborated types, and public-surface behavior.


napetrov:
## Benchmark comparison — interim data dump (temporary, will be folded into `docs/reference/tool-comparison.md` once the full rerun finishes)

Posting this now as a durable record so the numbers survive a container/session restart. This comment will be superseded by a docs update once the full rerun below completes.

### Full 170-case catalog, all 6 modes (SKIP/ERROR/TIMEOUT/incapacity all counted as misses — nothing silently excluded)

This table was computed earlier in this session (all fixes from this PR already applied at the time):

| Mode | Correct / 170 | % | Total time |
|---|---:|---:|---:|
| **abicheck (L2, headers)** | 158 | **92.9%** | 155s |
| **abicheck source** (clang-plugin/full, L3-L5) | 102 | 60.0% | 686s |
| **libabigail (abidiff)** | 52 | 30.6% | 1.3s |
| **libabigail + headers** | 52 | 30.6% | 4.8s |
| **ABICC (abi-dumper)** | 82 | 48.2% | 1704s (~28 min) |
| **ABICC (xml)** | 79 | 46.5% | 2029s (~34 min) |

**ABICC is ~350-2000× slower than libabigail** for the same catalog (1704-2029s vs 1.3-4.8s) while scoring lower on accuracy — this is the load-bearing reason ABICC results are frozen (`--freeze`) rather than re-run on every iteration.

### False-positive / false-negative breakdown (libabigail, freshly re-verified — exact, not reconstructed)

Just re-ran (near-instant, ~1-2s total, so no reason not to verify exactly):

| Mode | Correct/170 | % | FP (over-called) | FN (under-called/incapacity) | Total time |
|---|---:|---:|---:|---:|---:|
| abidiff | 52 | 30% | 3 | 115 | 0.9s |
| abidiff+headers | 52 | 30% | 3 | 115 | 2.3s |

libabigail's misses are overwhelmingly false negatives (115/170) — it silently under-reports (e.g. layout/vtable/noexcept/access-level changes it has no AST to see), not false alarms. Its 3 false positives are cases where DWARF-only analysis over-calls a change abicheck's header+DWARF cross-check correctly clears.

Frozen to `scripts/frozen_competitor_results.json` (not yet committed — will commit alongside the docs update).

### In progress

Re-running `abicheck` (L2) + `abicheck_full` (L3-L5) + both ABICC modes now to get an exact, freshly-verified FP/FN split for every lane (not just libabigail) and a fully current accuracy/timing table reflecting every fix landed in this PR. ABICC alone will take ~30-40 min per mode. Will post final numbers here and fold everything into `docs/reference/tool-comparison.md` once done.


---

napetrov:
## Final benchmark results — full 170-case catalog, all 6 modes

Supersedes the interim data dump above (#issuecomment-4950371142). Freshly re-run end-to-end with all fixes from this PR applied, folded into `docs/reference/tool-comparison.md`.

| Tool | Correct / 170 | Accuracy | FP (over-called) | FN (under-called/incapacity) | Total time |
|---|---:|---:|---:|---:|---:|
| **abicheck (L2, headers)** | 160 | **94.1%** | **0** | 10 | 835s (~14 min) |
| abicheck (L3-L5, +sources) | 104 | 61.2% | 17 | 49 | 3977s (~66 min) |
| libabigail (`abidiff`) | 52 | 30.6% | 3 | 115 | **~1s** |
| libabigail + headers | 52 | 30.6% | 3 | 115 | **~2-5s** |
| ABICC (abi-dumper) | 73 | 42.9% | 2 | 90 | 2534s (**~42 min**) |
| ABICC (xml/legacy) | 80 | 47.1% | 1 | 84 | 2143s (**~36 min**) |

**ABICC is ~500-2500× slower than libabigail** on the identical catalog while scoring *lower* than abicheck's own L2 lane — the load-bearing reason ABICC/libabigail results are frozen (`scripts/frozen_competitor_results.json`, committed) rather than re-run every time abicheck itself changes.

**FP/FN reading:** libabigail's 115 misses are almost entirely false negatives (DWARF has no view into noexcept/static/const/layout-invisible changes) — it rarely cries wolf. ABICC's misses skew false-negative too, but a large share are outright SKIP/ERROR/TIMEOUT (case85, case09, case105, case109... routinely hit the 90s timeout in this environment) rather than a wrong-but-confident verdict. abicheck's L3-L5 lane is the one spot with a real false-positive rate (17) — it's intentionally more sensitive (RISK/API_BREAK source-replay findings the L2 lane doesn't attempt), tracked as a known, documented tradeoff rather than hidden.

abicheck L2's 10 remaining misses: `case20`, `case78`, `case97` (one shared root cause — enum-as-literal-constant / hidden-friend-adjacent gaps), `case105`, `case111` (documented detector gaps), and `case130`-`case133` (build-mode flips needing L3 `-p build/` context, structurally outside the L2-only lane).

Operational note for anyone re-running this: ABICC can leave orphaned `abi-compliance-checker` perl processes running past their subprocess timeout (already documented in the doc) — worth a `pkill -f abi-compliance-checker` check before/after a full run.


---

napetrov:
## Investigated "source scan is worse" — found and fixed two benchmark-harness bugs, not a product regression

The `abicheck_full` (L3-L5, +sources) lane scored much worse than L2 (104/170, FP=17, FN=49) — root-caused it to two bugs in `scripts/benchmark_comparison.py`, both confirmed by direct reproduction:

1. **Forced `-include <header>`**: `_build_plugin_side` unconditionally injected `-include <header>` into every plugin-instrumented compile ("compact fixtures often define the API without including their header"). But many fixtures (`case07`, `case08`, `case09`, `case14`, `case19`, `case21`-`23`, `case25`, `case26`, ...) legally redefine a type independently in their `.c` file without including the header — forcing the include caused a **hard redefinition compile error**, misreported as `ERROR`. The CMake macro already has a proper per-case opt-in for exactly this (`V{version}_FORCE_INCLUDE` in `examples/CMakeLists.txt`), so the harness's blanket duplicate was redundant for cases that need it and actively harmful for ones that don't. **Fix: removed it.**

2. **False "wrong release" rejection**: `case04_no_change`'s `CMakeLists.txt` deliberately points both `V1_SOURCES` and `V2_SOURCES` at the same file (`v1.c`) to guarantee zero ABI diff — but `find_sources()`'s naive `v1.c`/`v2.c` filename-guessing assumed distinct files per side, so `_plugin_pack_is_target_specific` rejected the pack as cross-contaminated. **Fix: added `_cmake_declared_source()`** to read the real compiled source from `CMakeLists.txt`, and the validator now tolerates a deliberately-shared source.

### Result

| | Before | After |
|---|---:|---:|
| Correct/170 | 104 (61.2%) | **116 (68.2%)** |
| False positives | 17 | 16 |
| False negatives | 49 | 38 |
| Total time | 3977s (~66 min) | **694s (~12 min)** — ~5.7× faster |

The speedup is because doomed cases no longer burn the full 90s timeout before failing. Both fixes are benchmark-tooling-only — no product code changed; abicheck's actual L3-L5 detectors were never the problem.

Regression tests added (`tests/test_benchmark_smoke.py`): `_cmake_declared_source` reads real CMakeLists.txt sources, `_plugin_pack_is_target_specific` tolerates a shared source, and `_build_plugin_side` no longer emits `-include`. Also addressed a CodeRabbit finding on the way (unguarded `CMakeLists.txt` read could abort the whole run on one bad file — now guarded).

`docs/reference/tool-comparison.md` updated with the corrected numbers and an inline note explaining the fix for future readers.


---

napetrov:
## L3-L5 benchmark: 116/170 → 144/170 (84.7%)

Two more fixes landed on top of the earlier harness-bug fixes:

1. The special-case dispatcher (audit/cross-source, bundles, BTF, L3-L5 fixture packs, snapshot-pairs, Python stubs — 28 cases with no compilable v1/v2 source at all) only ever credited the `abicheck` column, leaving `abicheck_full` at its `SKIP` default regardless of active tools. These fixtures never go through a build lane — several are deliberately hand-built *at* L3-L5 — so there's no L2-vs-full distinction to make. Now credits both columns.
2. `case16_inline_to_non_inline`'s `.cpp` is genuinely header-only (empty file, inline function lives entirely in the header) and needed the CMake macro's `V{version}_FORCE_INCLUDE` opt-in to produce any plugin facts — it never declared it. Added it, converting an `ERROR` into a real verdict.

| | 104/170 (start) | 116/170 (harness fixes) | 144/170 (this round) |
|---|---:|---:|---:|
| Accuracy | 61.2% | 68.2% | **84.7%** |
| FP | 17 | 16 | 17 |
| FN | 49 | 38 | 9 |

### What's left for 100% (26 remaining misses, categorized)

- **15 cases — systematic false positive**: a plain compatible addition (`case03`, `case05`, `case13`, `case16`, `case29`, `case47`, `case49`, `case52`, `case54`, `case61`, `case62`, `case99`, `case136`-`138`) scores `COMPATIBLE_WITH_RISK` instead of `COMPATIBLE`. Root cause: `PUBLIC_REACHABILITY_CHANGED` fires whenever a declaration enters the public-reachability closure — including a declaration that's *brand new* (didn't exist in the old version), which just duplicates the already-correct `var_added`/`func_added` finding at inflated severity. I drafted a fix (only fire for a decl present in both graphs) but **reverted it** — it directly conflicts with an existing, deliberately-authored test (`test_findings_reachability_entered_and_left`), so this needs a product-policy call, not a unilateral change. Flagging for a maintainer decision.
- **5 cases — documented detector gaps shared with the L2 lane** (`case20`, `case78`, `case97`, `case105`, `case111`) — pre-existing, not new, not L3-L5-specific.
- **3 cases (`case118`-`120`) — no `CMakeLists.txt` at all.** The plugin-build lane only compiles CMake targets, so these are structurally unreachable without a new direct-compile-with-plugin-flags fallback (mirroring the L2 lane's `compile_so()`). Not yet implemented.
- **3 cases — one-off** (`case83`, `case103`, `case122`) needing individual triage.

All fixes committed and pushed; `docs/reference/tool-comparison.md` updated with the full breakdown.


---

napetrov:
## Summary of this session's work

Investigated why the L3-L5 (`abicheck_full`, source-scan) benchmark lane scored far worse than the L2 (headers) lane, and closed most of the gap through a mix of benchmark-harness fixes and two real product fixes.

**Score progression on the full 170-case catalog:**

| Stage | Correct | Accuracy | FP | FN |
|---|---|---|---|---|
| Start of session | 104/170 | 61.2% | 17 | 49 |
| After harness fixes (forced `-include`, wrong-release pack validation, special-case crediting) | 144/170 | 84.7% | 17 | 9 |
| After this round | **153/170** | **90.0%** | **7** | 10 |
| L2 (headers) lane, for reference | 160/170 | 94.1% | 0 | 10 |

### Two real product fixes

1. **`PUBLIC_REACHABILITY_CHANGED` narrowing** (`abicheck/buildsource/source_graph.py`) — it was firing whenever a declaration entered the public-reachability closure, including brand-new/fully-removed declarations, duplicating the already-correct `var_added`/`func_added` findings at inflated severity. Narrowed to only fire for a declaration present in *both* graphs — a genuine "persisting decl crosses the public boundary" signal. This replaces behavior a previous test had locked in; flagged for your judgment before applying, and you confirmed the reasoning.

2. **`EXPORTED_SYMBOL_SOURCE_OWNER_CHANGED` false-firing on 14 examples** — these cases named their per-version source files inconsistently (`v1.c`/`v2.c`, `bad.c`/`good.c`) instead of the `old/lib.<ext>`+`new/lib.<ext>` convention already used by `case19` onward. With only one declaring file per side, `_common_prefix_len()` had no sibling to structurally compare against and fell back to comparing full absolute paths, which still differ between `old/` and `new/`. Fixed by (a) renaming all 14 cases to the shared-basename convention (mechanical — no source/verdict semantics changed) and (b) giving `_common_prefix_len()` a single-declaring-file fallback that reserves just the basename, matching the "unmoved" outcome multi-file sides already reach structurally.

### Also fixed along the way
- A CodeRabbit review caught 3 real issues introduced by the rename: stale "security-hardening flag" comments in `case137`/`case138` that no longer matched what their `CMakeLists.txt` actually does (RUNPATH change / forced libm dependency), and an accidental `lib_version` value change in `case61` that broke its "purely additive" fixture intent. All three fixed.
- Test fallout in `test_appcompat_examples.py` and `test_castxml_free_examples.py` (both hardcoded the old flat-file convention).
- Regenerated `docs/examples/*.md` for the 14 renamed cases.

### Verification
Full fast suite (13,860 tests), FP-rate gate, tier-accuracy gate, ruff, mypy, and the AI-readiness gate are all green. `docs/reference/tool-comparison.md` is updated with the final numbers and an honest breakdown of the remaining 17 misses (3 structural/no-CMakeLists, 6 shared with the L2 lane, 5 residual reachability over-calls on genuine boundary-crossing declarations, 3 one-offs still needing triage).

---

PR #606, opened 2026-07-19
fix(clang-plugin,dumper): LLVM 19-22 support + --gcc-path icx/icpx recognition

napetrov:
## Summary

Two independent, locally-verified fixes that came out of validating claims from the oneDAL PR's Intel/icpx integration work:

- **Clang facts plugin now builds against LLVM/Clang 19-22** (`contrib/abicheck-clang-plugin/AbicheckFactsPlugin.cpp`). `clang::FileEntry::getName()` was removed upstream (the `SourceManager::fileinfo_*` iteration key moved to `FileEntryRef`, which keeps its own `getName()`, back in LLVM 18), so the now-unreachable `const FileEntry *` overload of `fileEntryKeyName()` failed to *compile* on current LLVM majors even though nothing calls it there. Guarded `#if CLANG_VERSION_MAJOR < 18` instead of relying on overload resolution alone to keep it dead code.
- **`--gcc-path` now recognizes Intel's `icx`/`icpx`/`dpcpp`/`dpcpp-cl`** as clang-family binaries (`abicheck/dumper_clang.py`). Previously the check only matched a `"clang"` substring in the binary name, so `--gcc-path /path/to/icpx` was silently ignored and the `clang` `--ast-frontend` fell back to a plain `clang` on `PATH` — a different, possibly-mismatched toolchain than the one the real build used. This removes the need for a symlink workaround (e.g. `icpx-clang++ -> icpx`) to get `--gcc-path` honored. As part of this, the match switched from `Path(gcc_path).name.lower()` to `.stem.lower()` — extension-agnostic (e.g. `icpx.exe` on Windows), a minor incidental improvement riding along with the alias-list addition.

## Verification

- Clang plugin fix verified against **real** LLVM/Clang 16, 17, 18, and 22 pulled from apt.llvm.org (not mocked): builds clean on all four, and the C.6 differential-conformance test, the end-to-end scan-flow test, and the public-roots diagnostic test all pass unmodified on every major.
- `.github/workflows/clang-plugin.yml` matrix extended from 16-18 to 16-22 (gcc + clang host legs at the oldest and newest tested majors) so this class of upstream API churn is caught automatically going forward.
- `--gcc-path` fix covered by new unit tests (`tests/test_dumper_clang.py`): a parametrized `test_is_clang_family_binary` exercises the pure classifier directly (both the `"clang"`-substring branch and the vendor alias-set branch, plus negatives like `cl.exe`/`icc`), and the existing `_clang_header_dump`-level tests confirm the end-to-end wiring (icx/icpx/dpcpp/dpcpp-cl honored, a real GCC binary still rejected).
- `ruff check`/`ruff format --check`/`mypy abicheck/dumper_clang.py`/`python scripts/check_ai_readiness.py` all clean (0 errors). CI: all 71 checks green (`codecov/patch` 92.85% patch coverage, target 90%).

## Note on diff size

`abicheck/dumper_clang.py` and `tests/test_dumper_clang.py` carry a lot of `ruff format` reflow beyond the ~25-line functional change — those two files weren't fully `ruff format`-clean on `main` before this PR, and touching them triggered a full-file reformat. That noise is why `changed_files`/`additions` look large relative to the actual fix; the functional diff is the `_is_clang_family_binary` addition in `dumper_clang.py`, the `#if CLANG_VERSION_MAJOR < 18` guard in the plugin, and the CI matrix entries.

## Non-goals / explicitly out of scope

This does **not** add Intel icpx support to the Clang plugin producer. Intel's oneAPI apt packages (confirmed via `dpkg-deb -c` against the real `2026.1.0-235` package) ship no `LLVMConfig.cmake`/`ClangConfig.cmake` at all — only `IntelSYCL`/`IntelDPCPP` CMake helpers — so a green matrix leg here says nothing about building against icpx's own fork. `AGENTS.md` documents this explicitly. The portable path for icpx today is the `wrapper` producer (`abicheck-cc icpx ...`) or `--ast-frontend clang` with `--gcc-path` (now fixed by this PR), neither of which need any vendor CMake package.

## Test plan

- [x] `python contrib/abicheck-clang-plugin/tests/conformance.py --plugin <so> --clangxx clang++` — passes on LLVM 16, 17, 18, 22
- [x] `python contrib/abicheck-clang-plugin/tests/scan_flow.py --plugin <so> --clangxx clang++` — passes on LLVM 22
- [x] `python contrib/abicheck-clang-plugin/tests/test_public_roots_diagnostic.py --plugin <so> --clangxx clang++` — passes on LLVM 22
- [x] `pytest tests/test_dumper_clang.py -q` — 175 passed
- [x] `ruff check` / `ruff format --check` / `mypy abicheck/dumper_clang.py` — clean
- [x] `python scripts/check_ai_readiness.py` — 0 errors

## Summary by CodeRabbit

* **New Features**
  * Added support for building and validating the Clang plugin with LLVM/Clang versions 19–22.
  * Added recognition for Intel LLVM-family compiler binaries, including `icx`, `icpx`, `dpcpp`, and `dpcpp-cl`, when configuring compiler paths.

* **Bug Fixes**
  * Resolved compatibility issues with newer LLVM releases.
  * Prevented GCC binaries from being incorrectly treated as Clang-family compilers.

* **Documentation**
  * Updated plugin support coverage and compiler-path behavior documentation.

### DIFF
--- a/.github/workflows/clang-plugin.yml
+++ b/.github/workflows/clang-plugin.yml
@@ -51,17 +51,34 @@ jobs:
       #  * LLVM major — the plugin is ABI-locked to its LLVM (C.5), so parity is
       #    validated per-version. Each major is pinned to a runner whose
       #    apt.llvm.org channel publishes it (16 is jammy-only; noble-16 does not
-      #    exist; 17/18 are on noble).
+      #    exist; 17-22 are on noble).
       #  * Host C++ compiler — the plugin is a plain LLVM shared module, so it
       #    must build with either GCC or Clang as the *host* compiler. LLVM 18
-      #    is built both ways (g++ and clang++) to prove that; 16/17 use the
-      #    distro default (g++) to keep the matrix small.
+      #    (the lowest major whose fileinfo_iterator key is FileEntryRef) and 22
+      #    (the newest major tested, currently the highest apt.llvm.org
+      #    publishes) are each built both ways (g++ and clang++) to prove that
+      #    across the FileEntry-API split; 16/17/19/20/21 use the distro default
+      #    (g++) to keep the matrix from growing on every dimension at once.
+      #
+      # 19-22 exist to catch upstream Clang AST-JSON/API churn as soon as it
+      # lands, not because oneDAL/Intel's icpx needs them: icpx is a separate
+      # downstream LLVM fork, not vanilla apt.llvm.org, so passing here does
+      # NOT mean the plugin builds against icpx's own bundled toolchain (which
+      # ships no LLVM/Clang CMake package at all, vendor or not -- see
+      # actions/collect-facts/run.sh's $CMPLR_ROOT handling). This matrix
+      # covers "does the plugin still build clean against the current LLVM
+      # majors a distro/apt.llvm.org actually ships."
       matrix:
         include:
           - { llvm: "16", runner: "ubuntu-22.04", host_cc: "gcc",   host_cxx: "g++" }
           - { llvm: "17", runner: "ubuntu-24.04", host_cc: "gcc",   host_cxx: "g++" }
           - { llvm: "18", runner: "ubuntu-24.04", host_cc: "gcc",   host_cxx: "g++" }
           - { llvm: "18", runner: "ubuntu-24.04", host_cc: "clang", host_cxx: "clang++" }
+          - { llvm: "19", runner: "ubuntu-24.04", host_cc: "gcc",   host_cxx: "g++" }
+          - { llvm: "20", runner: "ubuntu-24.04", host_cc: "gcc",   host_cxx: "g++" }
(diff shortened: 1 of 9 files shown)


napetrov:
Addressed the `__slots__` nitpick (sorted alphabetically, in cab1af0).

Not addressing the "Docstring Coverage 10.26%" pre-merge warning: this repo's `CLAUDE.md` convention is explicitly to default to no docstrings/comments unless the *why* is non-obvious (private helpers throughout this codebase — `_pointer_depth`, `_categorize`, `_walk`, etc. — are undocumented by design, pre-dating this PR). Adding docstrings purely to satisfy an 80% coverage heuristic would go against the project's own style and add noise for no benefit here.

---

PR #629, opened 2026-07-25
fix: pair-wide C++20 dialect resolution + P1 toolchain-profile groundwork

napetrov:
## Summary

Closes out the last remaining P0 gap from the toolchain-profile audit (pair-wide C++20 dialect resolution) and starts on P1 (toolchain-profile config schema + structured provenance).

## Motivation

Follow-up to the toolchain-profile audit (previous commit on this branch fixed the C++20 `#error`-directive false positive, hidden-friend surface classification, the CastXML version gate, and SONAME evidence gating). That audit found one real remaining P0 gap — `compare`'s C++20 heuristic only ever inspects one side's headers at a time, so old/new could silently resolve to different dialects — plus several well-scoped, safe P1 items.

## Changes

**P0**
- `cli_helpers_compare._pair_wide_dialect_override` (wired into the CLI `compare` path) and a matching fix in `service.run_compare_request` (Python-API/MCP path) resolve the C++20 requires/concept heuristic once, over the union of both sides' headers, and pin the identical explicit `-std=gnu++20` onto both — instead of each side guessing independently. An explicit user-supplied `-std=` always still wins.
- `build_context.py`'s `_ABI_EXTRA_PREFIXES` now forwards `-m32`/`-m64`/`-march=`/`-mtune=`/`-mabi=`/`-mfloat-abi=`/`-stdlib=`/enum-packing/char-signedness flags from a matched `compile_commands.json` entry into the real `dump -p`/`compare -p` header parse (previously only a narrow subset reached the parse; the rest were captured only for the separate advisory build-evidence-drift diff).
- `--allow-unsupported-castxml` is now a real CLI flag (`dump`/`compare`/`scan`), matching `--allow-ast-frontend-fallback` — previously only the `ABICHECK_ALLOW_UNSUPPORTED_CASTXML` env var existed, despite `model.py`'s snapshot-provenance docstring already documenting a CLI flag.

**P1 (groundwork)**
- `profiles.<id>.compile`: an optional, declarative compile-context overlay for `.abicheck.yml` `profiles:` entries (`compiler_family`/`compiler_version`/`target`/`standard`/`stdlib`/`abi_macros`/`args`), validated with the same flag-injection-safe atom check `compile.std` already uses — no raw executable path/command (trust-boundary safe for untrusted auto-discovered config). Schema/validation only. G30 P1.4's run-plan generator (merged separately as #627 while this branch was in flight) resolves `(target, profile, check)` cells from `build-output.json` but does not yet read this `compile:` overlay — no consumer forwards it into an actual `dump`/`compare` invocation yet.
- Snapshot schema **v14 → v15**: `AbiSnapshot.ast_resolved_standard` / `ast_cplusplus_macro` / `ast_compile_args` / `ast_sysroot` record the compile context actually used for a header-AST parse. Purely additive; a pre-v15 snapshot loads all four as their conservative "not recorded" defaults. (Originally landed as v14, but this branch was rebased onto `main` after ADR-050 D1's `AbiSnapshot.contract` field — a separate, verdict-blocking addition — independently claimed schema v14 there first; renumbered to v15 rather than reusing the same integer for two unrelated additions.)

## Test plan

- [x] Full fast unit suite (`pytest -m "not integration and not libabigail and not abicc and not slow and not golden"`): **18906 passed**, 29 skipped, 4 xfailed, 0 failed (post-rebase, includes P1.4's new `run_plan`/`reusable_workflows` tests from `main`).
- [x] New unit tests: `tests/test_ast_compile_provenance.py` (21 tests), extensions to `tests/test_cli_helpers_compare_coverage.py`, `tests/test_project_targets.py`, `tests/test_build_context.py`, `tests/test_compile_context_parity.py`.
- [x] `tests/test_pvxs_regression.py` (new, `integration`-marked, uses `--ast-frontend clang` so it runs without castxml).
- [x] `ruff check`/`ruff format --check` clean on every file this PR touches (checked against the `ruff==0.9.0` pin from `.pre-commit-config.yaml`, not the newer version otherwise on `$PATH`).
- [x] `mypy abicheck/`: clean (0 errors).
- [x] `scripts/check_ai_readiness.py`: 0 errors.
- [x] `scripts/check_docs_contract.py`: 0 errors.
- [x] Regenerated the 11 committed G20 fixture snapshots (`scripts/gen_g20_fixtures.py`) for the schema v15 bump; diff is schema-version only.
- [x] Rebased onto `main` (which had moved 3 commits ahead, including the independent ADR-050 `contract`/v14 addition and G30 P1.4's run-plan generator) — resolved the `serialization.py` schema-version conflict by renumbering this branch's fields to v15, and fixed every downstream literal (docs, tests, docstrings) plus two pre-existing stale-`v12` / undocumented-`contract` gaps already on `main`.

## Checklist

- [x] Changelog fragment added (`changelog.d/20260725_010000_..._fixed.md`, `..._added.md`)
- [x] Docs updated (`docs/reference/snapshot-format.md`, `docs/reference/environment.md`)

---
_Generated by [Claude Code](https://claude.ai/code/session_017Yp1nERwwW8mjPbFNgAmYQ)_

## Summary by CodeRabbit

* **New Features**
  * Added optional per-profile compile settings for compiler, target, language standard, ABI macros, and additional arguments.
  * Added command-scoped `--allow-unsupported-castxml`.
  * Added header-AST compile-context provenance to ABI snapshots.
  * Automatically pins consistent C++20 settings across comparison inputs when needed.
* **Bug Fixes**
  * Prevented false C++20 detection from guarded error messages.
  * Forwarded ABI-relevant compiler flags into header parsing to avoid header-analysis drift.
* **Documentation**
  * Updated environment and snapshot-format docs for schema version **15** (provenance fields included).


napetrov:
Pushed `45e1ffe` (macOS CI fix) and `4885bfd` (review response) since the last automated review.

**CI**: `integration-tests (macos-latest)` was genuinely failing — `--scope-public-headers` can't match this test's minimal Mach-O export table to header declarations (an unrelated, platform-specific scoping fallback), which widens the verdict to `COMPATIBLE_WITH_RISK` and earns its own correct `PATCH` release recommendation. Took two pushes to get the assertions right; they now check the actual invariant a forced-C++20 regression would violate (no `BREAKING`/`API_BREAK`, no `MAJOR` bump, no SONAME-bump advisory) rather than an exact verdict/bump match. The "Full example matrix" / "Build-source smoke" failures reported against `4495e1a` were a cascading effect of that commit's in-flight CI run being superseded by the next push (exit 143 / runner shutdown signal) — not a real failure; the fresh run at the latest commit is green.

**CodeRabbit review** — addressed:
- 🟠 Major/security: `ast_compile_args`/`ast_sysroot` now go through `RedactionPolicy`, matching every L3 build-evidence adapter (`compile_db.py`, `make.py`, `bazel.py`, `ninja.py`, `cmake_file_api.py`) — a `--gcc-options` token could otherwise carry a secret-looking `-DTOKEN=...` or an absolute home path straight into a persisted snapshot. Added regression tests.
- Dedup: pair-wide C++20 dialect resolution was hand-written twice (CLI `compare` vs. `service.run_compare_request`). Extracted the shared core to `service_scan.pair_wide_cxx20_std_override` — a leaf module both sides already depend on, so this doesn't introduce a CLI-layer → service-layer dependency. Also broadened the explicit-`-std=` guard to check both compile contexts instead of relying on an unstated single-call-site invariant.
- Dedup: the two near-identical env-var-scoping Click callbacks → one `_scoped_env_flag_callback` factory.
- Dedup: the repeated "combine `gcc_option_tokens` + shlex-split `gcc_options`" logic → `_combined_option_tokens`.
- Docs: added a schema table for the four new v14 fields to `snapshot-format.md`.

**Not changed** (reasoned, not skipped): the schema-version-literal nitpick in `snapshot-format.md` — those hardcoded numbers already have a working sync guard (`test_docs_snapshot_schema_version_matches_constant`), and this PR follows that pre-existing, tested convention rather than an ad-hoc one. Replacing it with a generator is a separate, repo-wide documentation-architecture change out of scope here.

Full fast unit suite (18527 tests) and `mypy`/`ruff` (pinned to the `0.9.0` `.pre-commit-config.yaml` version, not whatever's newer on `$PATH`) are clean on every touched file.


---

napetrov:
Addressed the review from the rebased commit (`cd0fae1`):

**🟠 Major (`tests/test_pvxs_regression.py`) — fixed in `336739b`.** Correct catch: `int pvxs_op(int x)` mangles identically under `gnu++11` and a wrongly-forced `gnu++20`, so the original verdict/SONAME-advisory assertions couldn't actually distinguish "dialect stayed correct" from "dialect got force-bumped" — the test would have passed even with the detection bug reintroduced. Now dumps each side directly and asserts the schema v15 `ast_resolved_standard` provenance field is `None` (not forced). Verified this is decisive: manually forcing `--gcc-options -std=gnu++20` on the same fixture makes the new assertion fail as expected. The `compare` step and its verdict/SONAME assertions remain as a secondary full-pipeline smoke check.

**Nitpicks — not changing:**
- `docs/reference/snapshot-format.md` schema-version literals: no doc generator exists for this page today (only a sync *test*, `test_docs_snapshot_schema_version_matches_constant`, checked pre-rebase too). Building one is a separate, repo-wide documentation-architecture change out of scope here — same call made on the prior review round.
- Inline `_GUARDED_HEADER`/`_SOURCE` fixtures in `test_pvxs_regression.py`: this matches the established pattern for single-file integration tests here (`test_castxml_clang_parity_gate.py`, `test_header_scan_deadline_integration.py` do the same) — `tests/fixtures/` is for content shared across multiple test files, not test-local content.

`codecov/patch`'s earlier 50.81% failure was a premature evaluation before the canonical coverage-instrumented lane finished uploading — it self-corrected to 91.26% once that lane completed, no code change needed.

---

PR #635, opened 2026-07-26
feat: G32 Phase C — real compatible merge across translation units (ADR-050 D4)

napetrov:
## Summary

Implements G32 Phase C — the real compatible-merge lattice across translation units (ADR-050 D4), replacing Phase B's placeholder merge that hard-failed on any repeated `entity_key` across TUs regardless of whether the two declarations were actually compatible.

## Motivation

Phase A (comparability gate) and Phase B (manifest + real multi-TU dump) already merged. Phase B shipped its own merge step deliberately strict — "any repeat is an error" — as a documented placeholder, explicitly deferring real ODR-aware reconciliation to Phase C. Without it, a routine multi-TU manifest (e.g. a struct forward-declared in one TU and defined in another, or the same function redeclared identically in two TUs) fails extraction outright instead of merging cleanly.

## Changes

- **New `abicheck/tu_merge.py`**: `merge_fragments()` classifies each cross-TU redeclaration of the same `entity_key` as either:
  - a **trivial merge** (union provenance, keep the richer declaration) — a forward declaration + definition, a plain redeclaration, or a difference confined to an added default argument/initializer; or
  - a **genuine conflict** — anything else (differing return type, layout, member list, a typedef/constant's value) — raised as a new `TuMergeError` (`code="INCONSISTENT_DECLARATION"`).
  - Merge is deterministic regardless of TU-processing order (fragments are always folded in tu_name-sorted order, never the caller's own sequence order).
  - A single TU's own intra-fragment duplicate declarations are left untouched (not treated as a merge concern), preserving Phase B's documented tolerance for that case.
  - `HETEROGENEOUS_ABI_CONTEXT` is defined but unreachable today, since `dump_manifest.py`'s parse-time single-profile-per-manifest rule already prevents it (documented, not implemented — matches the ADR).
- **New `abicheck/tu_fragment.py`**: `TuFragment`/`MergedTuFragments`/`entity_key` move here from `dumper_manifest.py` into a dependency-free leaf module, so `tu_merge.py` and `dumper_manifest.py` can both depend on them without forming a `dumper_manifest -> tu_merge -> dumper_manifest` import cycle. Re-exported unchanged from `dumper_manifest.py` for backward compatibility.
- **`abicheck/errors.py`**: new `TuMergeError(SnapshotError)` — an extraction-time failure (not a `ChangeKind`), so existing `except SnapshotError` handling around `dump()`/`compare()` keeps catching it unchanged.
- **`dumper_manifest.merge_tu_fragments`** is now a direct alias of `tu_merge.merge_fragments` under its original Phase B name.
- Updated `tests/test_dumper_manifest.py`'s placeholder-era "any repeat raises" tests to reflect the new real-merge semantics (identical/compatible redeclarations now merge; only genuine conflicts raise), including through the real clang backend and a real compiled `.so`.
- New `tests/test_tu_merge.py`: order-independence property test, the same-TU-duplicate tolerance, a structural guard that `TuMergeError`'s codes are never registered as a `ChangeKind`, and end-to-end tests loading G32 Phase 0's own committed `odr_safe`/`odr_conflict` fixture headers through the real clang backend.

## Checklist

- [x] Docs updated if user-visible behaviour changed — N/A (internal extraction-pipeline behavior; ADR-050/G32 plan docs already describe this phase)
- [x] `pre-commit run --all-files` passes locally (ruff check/format, mypy, ai-readiness)

Full fast test suite: 19195 passed, 30 skipped, 4 xfailed, 0 failed. `mypy abicheck/`: clean. `scripts/check_ai_readiness.py`: 0 errors (same 64 pre-existing warnings as baseline).

---
_Generated by [Claude Code](https://claude.ai/code/session_016ta1SHKtixokM1UuP3CUk9)_

## Summary by CodeRabbit

* **New Features**
  * Manifest-driven ABI dumps now reconcile compatible repeated declarations across translation units, including forward/definition pairs and default-argument/initializer differences.
  * Public-header provenance is applied during merge to choose the most appropriate merged source details.
* **Bug Fixes**
  * Genuine cross-translation-unit inconsistencies now fail early with a structured `TuMergeError` (via an explicit error code), including mixed AST-producer protections.
* **Tests**
  * Updated and expanded unit and end-to-end coverage to validate merge reconciliation, determinism, and the new error behavior.


napetrov:
Pushed e179c34 addressing Codex's four P1 findings on `tu_merge.py` plus the Windows CI failure:

- **Provenance loss on merge**: a compatible cross-TU redeclaration now prefers whichever side's `source_location` classifies as a declared public header (`merge_fragments` now threads `public_header_paths`/`public_header_dirs` through), instead of an arbitrary tu_name-sorted pick.
- **Conflicting non-null defaults**: `f(int=1)` vs `f(int=2)` (and the `Variable` analogue — two differing non-`None` initializer values) now correctly raise `TuMergeError` instead of silently keeping one side.
- **RecordType kind mismatch**: a `union X;` forward declaration no longer merges into a `struct X { ... };` definition.
- **EnumType underlying-type mismatch**: `enum E : int;` no longer merges into `enum E : unsigned { X };`.
- **Windows CI failure** (`test_odr_conflict_fixture_raises_through_real_clang_backend`): the test forced `lang="c++"`, which apparently routed Windows' clang toward MSVC-mode mangling where the int/double return-type conflict stopped colliding under the same `entity_key`. Dropped the override (the fixture headers don't need it) to match the already cross-platform-proven pattern used elsewhere in the suite.

Also added regression tests for all four fixes and closed `tu_merge.py`/`tu_fragment.py` to 100% line+branch coverage (addresses the Codecov patch-coverage flag too).

---

napetrov:
Pushed b1e13c1 addressing two more Codex findings:

- **Qualified-name key collision**: `RecordType`/`EnumType` merge keys now use `qualified_name or name` instead of the always-bare `name`, so `one::X`/`two::X` no longer spuriously collide into a false `INCONSISTENT_DECLARATION`.
- **Provenance on forward-declaration merges**: the follow-up to the earlier provenance fix — a forward-decl/definition merge now keeps the definition's structural facts (fields/members) but prefers whichever side's `source_location` is public (new `_with_more_public_provenance` helper), so a public header that only forward-declares a type whose definition lives in a private header no longer reads as private.

Both have regression tests; `tu_merge.py`/`tu_fragment.py` stay at 100% line+branch coverage.

On the third finding ("Compare same-TU extras with declarations from other TUs") — I looked at this closely and don't think it's a new bug, so I'm leaving it as-is: the scenario requires one TU to itself contain two structurally-conflicting declarations under the same `entity_key` (e.g. `f(): void` and `f(): int` both from TU `b`). Since `candidates` is built by scanning fragments in order and appending every match, one TU's own contributions to a given key are always contiguous — so this isn't actually about `acc_tu` drifting to the *wrong* TU, it's TU `b`'s own two entries never being checked against each other. That's the pre-existing, intentional Phase B tolerance ("a single TU's own parser output can legitimately repeat a key... only a repeat introduced by a *later* TU is this merge's concern" — inherited from the original placeholder, and explicitly covered by `test_merge_fragments_leaves_intra_fragment_duplicates_untouched`). It's also not reachable from a real compiler: two genuinely conflicting declarations under one mangled name within a single successfully-compiled TU is a redefinition error the compiler itself would reject before producing an AST — so this only matters for hand-built/adversarial fragments, not real `dump()` output. Happy to revisit if there's a real-compiler path I'm missing.

---

napetrov:
Pushed a4341c7 addressing two more Codex findings:

- **Mixed AST producers across TUs**: `merge_fragments` now rejects a manifest whose TUs were extracted by different AST producers (`TuMergeError`, `code=HETEROGENEOUS_ABI_CONTEXT`) before merging any entities. Good catch — `dump_manifest.py`'s parse-time rule only guarantees a uniform *declared* compiler/target; `--ast-frontend auto`'s per-TU fallback can still independently land one TU on castxml and another on clang, and `dumper.py` trusts a single representative `ast_producer` for the whole merged snapshot (gating DWARF layout backfill). This is `HETEROGENEOUS_ABI_CONTEXT`'s first real trigger — it was previously documented as reserved-but-unreachable, which turned out to be wrong; updated those docstrings too.
- **Parameter names blocking routine redeclarations**: `void f(int value);` and `void f(int n);` are the same function type — parameter names aren't part of a C/C++ function's type — but the merge was comparing them as part of dataclass equality and raising `TuMergeError` on this very ordinary cross-TU redeclaration shape. Now blanked alongside `default` before comparison.

Regression tests added for both; `tu_merge.py`/`tu_fragment.py` stay at 100% line+branch coverage.

Also noticed a **"Full example matrix" CI failure** on an earlier commit (f974762) — that was a missing-artifact error (`validate_examples-release-headers.json` not found) from the Examples Validation workflow, caused by an upstream job getting cut short when a later push superseded that run mid-flight. Unrelated to this PR's code; should resolve cleanly on the latest push.

---

napetrov:
Pushed 39fd861 addressing this round's findings.

**Codex — 2 real bugs fixed:**
- **`class`/`struct` forward-decl compatibility**: `class X;` followed by `struct X { ... };` is valid, ordinary C++ (both compilers accept it — struct/class are the same underlying entity, differing only in default member access). Only `union` is a genuinely different, incompatible type category. New `_record_kinds_compatible()` replaces the strict `a.kind == b.kind` check.
- **Parameter names leaking from the losing side**: good catch — `_merge_functions` built the merged parameter list from `a` unconditionally, even when `b` won as the provenance base, so a private declaration's parameter names could leak onto a function attributed to the public header (which `diff_symbols.py` could read as a spurious `PARAM_RENAMED`). Parameters now come from whichever side is actually selected as `base`, with only `default` unioned from the other side.

**CodeRabbit — 3 quick wins applied**: changelog now documents `HETEROGENEOUS_ABI_CONTEXT`, `_T` is bound to a `_Provenanced` Protocol (drops the `attr-defined` suppressions), and `EnumMember` is hoisted to a module-level import in the test file instead of eight repeated local ones.

**Declined, documented instead of applied:**
- **Qualifying typedef/constant merge keys by namespace** (Codex, P1) — this is a real gap, but fixing it properly requires `dumper_clang.py`/`dumper_castxml.py` to actually record qualified typedef identity, and today's `TuFragment.typedefs` (a bare `dict[str, str]`) has nowhere to carry that — unlike `RecordType`/`EnumType`, which already have a separate `qualified_name` field I could key on. That's a producer-side data-model change, out of proportionate scope for this round; left as a known limitation rather than a quick fix that would just paper over it.
- **Marking the real clang-backend tests `integration`** (CodeRabbit) — this file's own docstring already explains why they're deliberately not: the `integration` marker's CI gate requires castxml, which these tests don't need (only clang). Matches the pre-existing precedent the file already established before this PR. I did apply the other half of that comment — removed the two `odr_safe`/`odr_conflict` tests' genuinely-unused `tmp_path` parameter.

`tu_merge.py`/`tu_fragment.py` stay at 100% line+branch coverage; mypy/ruff/ai-readiness clean.

---

napetrov:
Pushed a1e8b39 addressing Codex's finding: a `static` function's mangled spelling never needs cross-TU uniqueness (the symbol never leaves its own object file), so two unrelated TUs each declaring their own private `static void helper(int)` could mangle identically and either get silently folded into one entity or raise a false `INCONSISTENT_DECLARATION` when they happened to differ. `_merge_group`'s key function now receives `tu_name` alongside the entity, and the new `_function_key()` folds `tu_name` into the key specifically for `is_static` functions — non-static (externally-linked) functions are unaffected and still merge across TUs exactly as before.

Regression tests added for the collision case, the still-merges-within-one-TU case, and a non-static sanity check. `tu_merge.py`/`tu_fragment.py` stay at 100% line+branch coverage; mypy/ruff/ai-readiness clean.

I'll hold off on further pushes for a bit to let CI (in particular the Examples Validation workflow, which kept getting cancelled mid-flight by each rapid successive push) actually run to completion on a stable commit.

---

napetrov:
Round 6 pushed (a53bf03): fixes 3 more Codex findings on round 5's static-linkage TU-scoping (a1e8b39).

Round 5 used `fn.is_static` to decide whether a function key should be scoped per-TU (genuinely file-local) or shared across TUs. That signal turned out to be both incomplete and over-broad:

- **Missing anonymous-namespace entities** — clang does not set `storageClass="static"` for a function/variable inside an anonymous namespace, even though it has the same internal linkage as a file-scope `static`. `is_static` silently missed these, so two unrelated anonymous-namespace helpers with the same name in different TUs would have been wrongly merged into one.
- **Missing local-linkage variables entirely** — `Variable` has no `is_static` field, so round 5 didn't scope variables by TU at all.
- **Over-broad** — `is_static` is *also* `True` for a C++ `static` **member** function/variable (e.g. `struct Widget { static int make(int); }`), which has ordinary external linkage and mangles with no local-linkage marker. Round 5's fix would have wrongly kept these as separate per-TU entities across every TU instead of merging them — the opposite bug from the previous round in the same direction.

I verified the correct signal empirically against real `clang` (compiling snippets, then inspecting `nm` output and `clang++ -Xclang -ast-dump=json`) rather than trusting the finding at face value: a genuinely TU-local entity's **Itanium-mangled name** either starts with `_ZL` (file-scope `static`) or contains `_GLOBAL__N_` (anonymous namespace); a `static` class member has neither marker despite `storageClass=="static"` being identical in the AST for both cases.

Added `_has_local_linkage_mangling()` and use it for both `_function_key()` and a new `_variable_key()`, falling back to `fn.is_static` only for the plain-C case (`fn.mangled == fn.name`, where member-function ambiguity can't exist). Added 5 regression tests covering anonymous-namespace functions/variables (kept distinct across TUs) and static member functions/variables (still merged across TUs).

One gap remains documented, not fixed: plain-C file-scope `static` variables have no detectable signal (C has no mangling, `Variable` has no `is_static` field) — would need a model/schema change, out of scope for this round.

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (80 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors.

---

napetrov:
Round 7 pushed (0a342fe): fixes 3 more Codex findings on commit a53bf03 (round 6), plus documents one genuine gap that isn't fixable at this layer.

**Fixed:**

1. **Nested-namespace internal-linkage manglings weren't detected.** `namespace ns { static int state; }` mangles to `_ZN2nsL5stateE` — the Itanium `<local-name>` marker (`L`) sits *after* the namespace component, not as a `_ZL` prefix. Round 6's `_ZL`-prefix-only check missed this, so two unrelated TUs' own namespaced statics with the same spelling would have been wrongly folded together. Verified empirically against real clang (`nm` confirms the local-symbol convention for exactly this mangled name). Fixed by reusing `dumper_castxml.py`'s existing `_mangled_name_is_local_linkage` — a proper length-prefixed-identifier-chain parser — instead of writing a second, narrower `_ZL`-prefix-only check. No import-cycle risk: `dumper_castxml.py` only imports `.model`/`.provenance`, same as `tu_merge.py` already does.

2. **`deprecated` was silently dropped across a forward-declaration/definition merge.** A public `class [[deprecated("old")]] X;` forward declaration merged with an undecorated private definition lost the deprecation entirely, since `_with_more_public_provenance` only ever copied provenance fields from the losing side, never `deprecated` — and `diff_types.type_deprecated` depends on this field surviving. Now unions `deprecated` the same way `_merge_functions`/`_merge_variables` already union their own optional facts (default arguments / initializer values), rejecting two differing non-`None` messages as `INCONSISTENT_DECLARATION` rather than arbitrarily preferring one.

3. **Both-opaque struct/class forward redeclarations incorrectly conflicted.** `class X;` in one TU and `struct X;` in another (no definition anywhere in the manifest) raised `INCONSISTENT_DECLARATION`, because the existing class-key-compatibility check was only reached when *exactly one* side was opaque, not when both were. `union` remains incompatible with either, as before.

**Documented, not fixed** (same category as the existing typedef/constant namespace-qualification gap from an earlier round): a `RecordType`/`EnumType` declared inside an anonymous namespace has no signal here to detect its TU-locality from — unlike `Function`/`Variable`, it carries no mangled linker symbol, and `dumper_clang.py`'s AST walker drops an anonymous namespace's nameless scope segment entirely rather than encoding it distinctly, so `qualified_name` can't tell the difference. Closing this needs a new producer-side model field — out of proportionate scope here.

Added 7 regression tests covering all three fixes (namespaced-static function/variable TU-scoping, deprecated-attribute union + conflict for both types and enums, both-opaque compatible/incompatible forward-redeclaration pairs).

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (87 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors (no new import-cycle growth).

---

napetrov:
Round 8 pushed (47def96): fixes 2 more Codex findings on commit 0a342fe (round 7).

1. **`deprecated` wasn't unioned across ordinary redeclarations, only forward-decl/definition pairs.** Round 7 taught `_with_more_public_provenance` to union `deprecated` for the forward-declaration + definition case, but `_blank_provenance` (used by `_merge_functions`/`_merge_variables`'s bare-equality check and `_merge_types`/`_merge_enums`'s two-full-definitions branch) never blanked `deprecated`, so a routine "one TU's redeclaration happens to carry `[[deprecated]]`, the other doesn't" case still failed exact-equality and raised `INCONSISTENT_DECLARATION` — a case that's exactly as compatible as differing provenance. Fixed by extending `_blank_provenance` to blank `deprecated` too, and unioning it back explicitly in every affected merge path (with a genuine conflict — two differing non-`None` messages — still rejected). Added a shared `_union_optional_fact` helper to avoid re-deriving this same conflict-or-union check five times, and a new `_merge_identical_modulo_provenance` shared by the two-full-definitions branch of `_merge_types`/`_merge_enums`.

2. **The both-opaque struct/class merge (added last round) picked whichever side was `a` unconditionally when the two class-keys differed but were compatible** — and `a` is just whichever TU sorts first by `tu_name`. That means the merged `kind` could flip between two otherwise-identical dumps that merely gain a redundant opaque redeclaration in a new, alphabetically-earlier TU, producing a spurious `SOURCE_LEVEL_KIND_CHANGED` between them even though nothing about the real declared entity changed. Now picks `min(a.kind, b.kind)` — a function of the two kind spellings alone, never of which TU contributed which side.

Added 8 regression tests: deprecated union + conflict for functions, variables, and full-type-definitions; and an order-independence test for the opaque `kind` fix that specifically covers the discriminating case (the `struct` declaration on the alphabetically-first TU, which the old code would have wrongly kept).

Also confirmed the full unrestricted test suite (19342 tests) passes clean against round 7's code, ruling out the aggregation-noise Codecov numbers as a real regression.

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (94 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors.

---

napetrov:
Round 9 (7f335f4): the "Normalize MSVC function identities before grouping" finding on commit 47def96 is real, but I'm documenting it rather than fixing it this round.

`entity_key`'s "`fn.mangled` is a return-type-independent identity" assumption holds for Itanium mangling and this module's plain-C fallback (both verified against real clang/GCC output in rounds 5-6), but not for the Microsoft C++ ABI, whose decorated names for free functions encode the return type. Two cross-TU declarations conflicting only in return type would decorate to different names under MSVC mangling and both silently survive the merge instead of raising `INCONSISTENT_DECLARATION`.

This is real — and already evidenced in this PR's own test suite: `test_odr_conflict_fixture_raises_through_real_clang_backend`'s comment (added during round 6's Windows CI fix) documents observing exactly this symptom on real Windows CI and works around it in the test rather than at the source. A correct fix needs a return-type-independent normalization of MSVC decorated names — a scheme-specific parser I can't write and verify without a real MSVC/`clang-cl` toolchain to check output against, which isn't available in this sandbox (Linux-only). Rather than guess at MSVC's decoration grammar without a way to verify it, I've documented this as a known, accepted limitation in `_function_key`'s docstring — the same call already made in this PR for the anonymous-namespace-type and typedef-qualification gaps, both of which needed a producer-side signal this fix can't fabricate either.

No behavior change in this commit, docstring only. Verified: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (94 passed, unchanged), `mypy`/`ruff check`/`ruff format` clean.

Also confirmed the "Full example matrix" CI failure reported on commit `0a342fe` is stale — that commit was superseded by `47def96` before the run finished, matching the artifact-race pattern already diagnosed earlier in this PR's history; the equivalent jobs on the current HEAD are green.

---

napetrov:
CodeRabbit's full review on 7f335f4 (1 actionable + 3 nitpicks). Fixed one, declining the other three:

**Fixed (519ebeb)**: `_blank_provenance`'s docstring said "these same three provenance fields" but round 8 added `deprecated` as a fourth blanked field — stale count, corrected.

**Declined — "Mark the real-clang fixture tests as `@pytest.mark.integration`"**: same suggestion CodeRabbit made in round 4, already explained then and unchanged since — it contradicts this test file's own pre-existing, documented rationale (see its module docstring): the `integration` marker's CI gate requires castxml, but these tests only need clang and already self-skip via `shutil.which("clang")`, the same convention `test_dumper_manifest.py`'s equivalent real-backend tests use. Marking them `integration` would just make them silently skip on the `abicc`/`libabigail` lanes' clang-only hosts for no reason.

**Declined — `_mangled_name_is_local_linkage` import from `dumper_castxml`**: real coupling observation, but "trivial/low value" per CodeRabbit's own tagging, and a proper fix means relocating the helper into a new shared leaf module and updating `dumper_castxml.py` too — broader than a documentation-only nitpick round should take on. Not a live import-cycle risk today (`dumper_castxml.py` doesn't import from `tu_merge.py`), just an architectural preference; leaving as-is for now.

**Declined — parametrize the TU-locality mangling tests**: pure style preference, explicitly framed as "Optional" by CodeRabbit itself; no functional issue.

Verified: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (94 passed, unchanged), `mypy`/`ruff check`/`ruff format` clean.

---

napetrov:
Round 11 pushed (e303920): fixes the "Merge compatible declaration attributes instead of rejecting them" finding on commit 519ebeb.

`Function.contract_attributes` wasn't blanked in `_merge_functions`' bare-equality comparison, so two TUs legally redeclaring a function where only one side carries an additive attribute (e.g. clang accepts `int f(int);` alongside a later `[[nodiscard]] int f(int);` redeclaration, normalized as `"warn_unused_result"`) failed the comparison and raised `INCONSISTENT_DECLARATION` — an ordinary redeclaration, not a conflict, the same shape as the `deprecated` gap fixed in round 8.

Added `_merge_contract_attributes()`: `None` ("not captured" per the field's own docstring) on either side contributes no information and the other side's value is kept as-is; otherwise tokens are grouped by attribute family (the text before any `(`, since argument-bearing tokens already carry their arguments — e.g. `nonnull(1)`, `format(printf,1,2)`, per `dumper_clang._clang_contract_attributes`). A token present on only one side is additive and unioned in; two different tokens sharing a family (e.g. `format(printf,1,2)` vs `format(scanf,1,2)`) are a genuine conflict and still raise `TuMergeError`, matching the finding's own request to "reconcile compatible additive attributes while still rejecting contradictory attribute arguments."

Added 5 regression tests covering the additive case, `None` treated as unknown from either side, token-level union with a shared token deduplicated, and the same-family-differing-arguments conflict.

Also confirmed: the "Full example matrix" CI failure reported on commit `7f335f4` is the same stale-artifact-race pattern already diagnosed earlier in this PR — that commit is superseded by later pushes.

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (99 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors.

---

napetrov:
Round 12 pushed (d314fbe): fixes 2 more Codex findings on round 11's `_merge_contract_attributes` (commit e303920), documents a third as a known limitation.

1. **Fixed — `nonnull(1)`/`nonnull(2)` wrongly treated as conflicting.** GCC and clang both accumulate `nonnull` constraints across separate attribute occurrences (`__attribute__((nonnull(1))) __attribute__((nonnull(2))))` means both parameters are nonnull, not a contradiction) — `dumper_clang.py`'s own `_clang_contract_attributes` already keeps each occurrence as its own list entry rather than folding them together. Added `_SET_VALUED_ATTRIBUTE_FAMILIES = {"nonnull"}` to exempt that one family from the same-family-conflict check; every other argument-bearing family (`format`, `regparm`, ...) keeps strict exact-match conflict handling.

2. **Fixed — `ms_abi`/`sysv_abi` calling-convention tokens silently unioned instead of conflicting.** These are different *bare* families (no shared `(` to collide on), so round 11's per-family check never caught the mismatch — both would survive onto one merged function, which `diff_symbols.py`'s own `_check_contract_attributes_change` treats as `CALLING_CONVENTION_CHANGED` when it later diffs two already-merged snapshots. Reused `diff_symbols._is_cc_attribute` (the module's own canonical `_CC_ATTRIBUTE_BASES` set, not a reimplementation) and reject a merge whose surviving attributes include more than one calling-convention token.

3. **Documented, not fixed — uninstantiated C++ template functions can share `entity_key`.** Verified empirically against real clang: `template <typename T> struct A { void run(T); };` and an unrelated `template <typename T> struct B { void run(T, T); };` both parse with `mangledName=None` for their `run` method, so `_function_key`'s `fn.mangled == fn.name` branch — which assumes that shape means "plain C" — also catches uninstantiated templates, falling back to `fn.is_static` (ordinarily `False` for a non-static method) and landing both in the same bucket. `Function` carries no signal today to distinguish "genuinely unmangled" from "C++ but clang produced no mangled name" at this call site (unlike `RecordType.is_template_pattern`, there's no per-function equivalent). Same call as the MSVC return-type-mangling gap from round 9: needs a producer-side model/schema addition, not a guess, so it's documented in `_function_key`'s docstring rather than fixed this round.

Added 3 regression tests (nonnull-argument union, calling-convention conflict, matching-convention merge).

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (102 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors (including no new import-cycle from the `diff_symbols._is_cc_attribute` reuse).

---

napetrov:
Round 13 pushed (e9574bc): fixes the "Accept differing deprecation messages on redeclarations" finding on commit d314fbe.

Verified empirically before changing anything (both `clang++` and `g++`, `-pedantic-errors`, compiling `[[deprecated("a")]] void f(); [[deprecated("b")]] void f();`): both compilers accept it cleanly. That contradicts my own round-8 design, which treated two differing non-`None` `deprecated` messages as a genuine `TuMergeError` conflict, the same way a real ABI-relevant disagreement (differing default arguments, differing `format()` attribute arguments) is. A deprecation message is additive diagnostic metadata, not part of a declaration's type or an ODR-significant fact — there's nothing to actually conflict on.

Replaced `_union_optional_fact` (conflict-or-union, used only for `deprecated`) with `_pick_deprecated`, which never fails: prefer whichever side the caller already selected as the merge's provenance representative (via `_more_public_of`) when it carries a message, otherwise fall back to the other side's. Rewired all four call sites (`_with_more_public_provenance`, `_merge_identical_modulo_provenance`, `_merge_functions`, `_merge_variables`) to compute their representative side first, then pick `deprecated` from it — instead of resolving-or-rejecting independently before knowing which side would even win.

Rewrote the 4 existing tests that asserted `TuMergeError` on differing deprecated messages to instead assert the new deterministic pick.

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (102 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors.

---

napetrov:
Round 14 pushed (3d887bf): fixes the "Pass only explicit provenance paths into the merge" finding on `dumper_manifest.py`, on commit e9574bc.

Confirmed against `dump_manifest.py`'s own docstring: `roots` is a *scope* declaration (D1's `scope_fingerprint` input — "what's being compared"), not an ADR-015 provenance input — "a manifest with public headers but no separate provenance fields behaves exactly like a legacy `dump -H foo.h` invocation with no `--public-header`" (every declaration stays `UNKNOWN`). The later, authoritative `apply_provenance()` call in `dumper.py` classifies origin against only the manifest's *explicit* `public_header_paths`/`public_header_dirs` — `roots` is never part of that set.

`run_tu_loop()` was passing `roots + public_header_paths` into `merge_tu_fragments()`'s public-surface preference (`_more_public_of` in `tu_merge.py`), so a declaration reached only through a root (but not explicitly public) header could win the merge's provenance-representative slot over one reached through a genuinely public header in another TU — then `apply_provenance()`, which never considered `roots` public, would reclassify the merged result private/unknown, hiding a real public API change.

Split the single roots-augmented path list into two: the roots-inclusive one still feeds each TU's own header-AST parser call, where it intentionally scopes constant extraction the same "roots are always public" way the legacy CLI's `-H`/`--header` already does (`dumper_castxml._CastxmlParser`'s own docstring) — that usage is correct and unrelated to this bug, so I left it untouched. `merge_tu_fragments()` now receives only the manifest's explicit `public_header_paths`/`public_header_dirs`, matching what `apply_provenance()` classifies against downstream.

Added a regression test in `test_dumper_manifest.py` — I verified it actually discriminates by reverting the fix, confirming the test fails (`root.h:1` wins instead of `public/api.h:1`), then re-applying the fix. While re-verifying coverage I also found a pre-existing, previously-uncovered branch in `_more_public_of` (the "first-sorted TU is already public, short-circuit" path) and added a unit test for it in `test_tu_merge.py`.

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (104 passed) plus a broader `-k "manifest or tu_merge or provenance"` sweep (697 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors.

---

napetrov:
Round 15 pushed (1ae5a3d): fixes the "Validate forward-declaration alignment before merging" finding on commit 3d887bf.

Confirmed against `dumper_castxml.py`'s `_build_record_type`: unlike `fields`/`bases`/`is_abstract` (all gated on `is_opaque`), `alignment_bits` is captured unconditionally from castxml's `align` XML attribute, including for an opaque/incomplete record. That's correct on castxml's part — an explicit `__attribute__((aligned(N)))` on a bare forward declaration is itself an ABI-relevant fact independent of the member layout — but `_merge_types`' opaque-vs-definition branches were accepting the pair unconditionally after the class-key check, silently discarding the opaque side's alignment. Two TUs with `struct __attribute__((aligned(16))) X;` forward-declared in one and a naturally 4-byte-aligned `struct X { ... };` definition in another aren't interchangeable redeclarations, but the merge kept only the definition's (weaker) alignment.

Added `_record_alignments_compatible()` — the alignment analogue of the existing `_record_kinds_compatible()`: at most one side committing to a non-`None` value, or both agreeing, merges; two different non-`None` values raise `TuMergeError`. Wired into all three of `_merge_types`' opaque branches (both-opaque, and both orderings of exactly-one-side-opaque) alongside the existing class-key check. The both-non-opaque branch needed no change — `alignment_bits` isn't blanked by `_blank_provenance`, so two full definitions disagreeing on it already raised via plain equality.

Added 5 regression tests covering both branch orderings, the both-opaque case, a matching-alignment success case, and the unset-alignment-is-not-a-conflict case.

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (109 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors.

---

napetrov:
Round 16 pushed (7609a1d): fixes CodeRabbit's "Compatible-but-unequal alignments keep the winner's `None` instead of the captured value" finding on commit 1ae5a3d, plus its accompanying test-strength nitpick.

Round 15's `_record_alignments_compatible` only checked compatibility and let `_merge_types`' opaque branches return the merge's chosen provenance representative unchanged — so when that representative's own `alignment_bits` was `None` and the *other* side had a real captured value (e.g. an `aligned(16))` forward declaration merged into a definition whose producer didn't capture alignment), the captured fact was silently dropped. Every other optional fact in this module (`Param.default`, `Variable.value`, `contract_attributes`, `deprecated`) unions toward whichever side actually captured something — alignment should be no different.

Renamed to `_merge_record_alignment()`, returning `(ok, merged_value)` like the module's other optional-fact unions rather than a bare bool. Wired into all three opaque branches: compute the alignment union before calling `_with_more_public_provenance` (still bailing on a genuine conflict), then `replace()` the alignment onto whatever it returns when the union differs from what it kept.

Also applied CodeRabbit's own nitpick on the same round: strengthened `test_merge_fragments_type_reconciles_unset_alignment_on_definition_side` (previously only asserted `len(merged.types) == 1`) to assert the merged `alignment_bits` actually equals the captured value — which is exactly the assertion that would have caught this gap in round 15.

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (109 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors.

---

napetrov:
Round 17 pushed (e87ad4e): fixes the "Keep private defaults out of public declarations" finding on commit 7609a1d.

This one required reversing my own earlier decision. `_merge_functions` unconditionally unioned a parameter's default argument from `other` onto `base` whenever `base`'s own default was `None` — including when `base` is the *provably public* side and `other` is *provably private*. A default argument grants callers a real capability (calling without that parameter); a private-only header redeclaring the public signature with an added default does not extend that capability to the library's actual public consumers, who never see the private header. Unioning it in anyway made the merged snapshot claim a capability the public API never granted — and a later change to the private-only default would then surface as a false `PARAM_DEFAULT_VALUE_REMOVED`/`CHANGED` finding against the public surface, directly undermining this repo's own FP-rate gate.

Added `_other_is_strictly_less_public()`: true only when `_more_public_of` picked `base` *specifically* because it classifies as `PUBLIC_HEADER` and `other` does not (not merely an arbitrary `tu_name` tie-break between two equally-classified or unclassifiable sides). Wired into the default union: when `other` is provably private, `base`'s own default (already `None` in that branch) is kept rather than falling back to `other`'s.

This meant reversing `test_merge_fragments_function_unions_default_from_the_losing_side_onto_winning_names`, which had encoded the buggy behavior as intentional — renamed and rewritten to assert the corrected outcome. Added three more tests covering the mirror case (public side's own default still unions in fine), the no-public-info case (pre-round-17 behavior unchanged when visibility can't be determined at all), and the neither-side-is-public case (a supplied-but-non-matching `public_header_paths` must not make an arbitrary tie-break masquerade as a proven-public winner).

Verification: `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (112 passed) plus a broader `-k "manifest or tu_merge or provenance"` sweep (705 passed), 100% line+branch coverage on `tu_merge.py`/`tu_fragment.py`, `mypy abicheck/` clean, `ruff check`/`format` clean, `check_ai_readiness.py` 0 errors.

---

napetrov:
## Round 18: closed remaining public/private leaks + a `_merge_group` order-dependency bug

Addresses 4 findings from this round's Codex review of `e87ad4e` (commit `525f947`):

1. **"Ignore conflicting defaults from private redeclarations" (`tu_merge.py:937`)** — confirmed: the conflicting-default rejection ran unconditionally, *before* `base`/`other`/`other_is_private` were even computed, so round 17's visibility guard never got a chance to apply to it. Moved the check to after those are known, gated on `not other_is_private`. Applied the same fix to `_merge_variables`'s conflicting-`value` check (the variable analogue, same latent bug, not yet flagged but mechanically identical).

2. **"Exclude private semantic attributes from the public merge" (`tu_merge.py:1022`)** — confirmed: `_merge_contract_attributes`'s union was unconditional. Added `_suppress_private_only_attributes`, applied only when `other_is_private`, which drops tokens contributed only by the private side — except calling-convention tokens, which are kept and were already validated unconditionally (they describe the actual compiled ABI regardless of which header spells them, so Codex's explicit carve-out).

3. **"Keep private deprecation metadata off public entities" (`tu_merge.py:588`)** — confirmed, and the same leak existed at all four `_pick_deprecated` call sites (`_merge_functions`, `_merge_variables`, `_with_more_public_provenance`, `_merge_identical_modulo_provenance`), not just one. Added a `secondary_is_private` flag to `_pick_deprecated`, threaded through every call site.

4. **`_merge_group`'s same-TU "extras" exemption (pre-existing open thread, `tu_merge.py:510`)** — verified by hand-tracing: whether a TU's own repeated declaration under a key got validated against another TU's conflicting declaration depended on which of that TU's own entries happened to be listed first in its fragment — an accident of order, not the actual conflict. Reworked to group candidates per-TU up front (order-independent), fold each TU's representative, then validate any leftover repeat against the final cross-TU result rather than exempting it outright. Documented the one residual, inherent limitation (two *different* TUs both hitting the same producer-side key-collision quirk can still produce a spurious conflict) in the same style as the two adjacent documented gaps in this file.

The "Full example matrix" CI failure reported against `7609a1d` is stale — that SHA is round 16's, already superseded by round 17's `e87ad4e` before the run completed.

**Verified**: 15 new regression tests (one per fixed scenario, plus the order-dependent-vs-not pair for #4); `pytest tests/test_tu_merge.py tests/test_dumper_manifest.py` (125 passed); `--cov=abicheck.tu_merge --cov=abicheck.tu_fragment --cov-report=term-missing` (100% line+branch, no gaps); `mypy abicheck/` (clean, 282 files); `ruff check`/`ruff format --check` (clean); `python scripts/check_ai_readiness.py` (0 errors); broader sweep `pytest tests/ -k "manifest or tu_merge or provenance" -m "not integration and not libabigail and not abicc and not slow and not golden"` (718 passed, 2 skipped).

---

PR #661, opened 2026-07-31
feat: ADR-049 Phases 3 and 4 — provider-evidence ledger, persisted contract context, replay

napetrov:
## Summary

Closes ADR-049's Phase 3 and Phase 4: the observed provider-evidence ledger and the phase's own measurement (Phase 3), and the real type graph, persisted three-block contract context, and the replay / new-policy re-evaluation procedures (Phase 4).

## Motivation

The plan's own Phase 3 section ended with "the provider-evidence ledger this phase's own gate names still does not exist for either domain", and its Phase 4 section with "not yet done … wiring so a real snapshot actually persists these blocks; the original-replay and new-policy-re-evaluation *procedures* themselves; and populating `TypeGraphSnapshot`'s nodes/edges with real type-graph content." Each of those is implemented here, and both phases' gates are now enforced by tests rather than described.

Everything stays advisory, as ADR-049 requires until Phase 7: no verdict, exit code, finding set, or default behaviour changes, and every addition is gated on the caller opting into `--contract-evaluation`.

## Changes

**Phase 3 — the provider-evidence ledger (plan §4.1)**

- New leaf module `abicheck/contract_evidence_collect.py` emits one `EvidenceSearchRecord` per (provider, side) — `public_header`, `export_table`, plus the `post_manifest`/`forced_public_symbols` overlays when a run configures them — each with its own status, completeness, identity coverage, requested-vs-searched scope, and content digest, so a provider failure stays scoped to its own domain. It reports *which* completeness guard failed rather than mirroring `ExportSurface.exclusion_is_provable`'s single boolean; records `configuration_coverage` as honestly `not_started` (no variant source exists in this build — which is precisely why `UNKNOWN_UNPROVEN` is never emitted); and encodes "not consulted" as an absent entry, never a failed one.
- Every stamped finding carries `Change.contract_evidence_refs`, serialized on each report ledger that already carries the schema-2.23 contract fields. A non-entity finding cites nothing (its relevance follows from its `ChangeKind` alone); an overlay-decided one cites that overlay, attributed in `contract_evaluation.evidence_refs_for_change` so the matching reuses that module's own overlay rules instead of a second copy; `validate_decision_evidence` rejects a reference that names no known record.
- `scripts/measure_contract_shadow.py` (mirrored into the unit lane by `tests/test_contract_shadow_measurement.py`) runs the labelled FP-rate corpus through the shadow evaluator in all three contract domains and reports the phase's four "Measure:" quantities, gating on three zero baselines: proven public-break losses, unevidenced deltas, and findings with no recorded decision.

**Phase 4 — persisted context, replay, re-evaluation (plan §5.1)**

- `build_type_graph()` replaces `TypeGraphSnapshot`'s placeholder content with the snapshot's real record/enum/typedef/declaration graph, walked policy-independently (a private declaration is recorded exactly like a public one) and resolved through the same indexes the live closure walk uses — which is why the block carries an `identity_algorithm_version`.
- `abicheck/contract_context.py` assembles the three blocks from a real comparison (`DiffResult.contract_context`), `abicheck/contract_context_io.py` round-trips them byte- and order-independently including the complete resolved `CompatibilityEvaluationConfig` and its field provenance, and `reporter.py` emits them as the JSON report's `contract_context` block from all three report paths (full / leaf / root-cause).
- `abicheck/contract_replay.py` implements both procedures ADR-049 D6 names. Replay reads the receipt alone — this build's provider defaults cannot alter a recorded decision (tested by emptying the entire evidence block). Re-evaluation walks the persisted observations under a *different* contract mode with no re-collection and no live re-probe. Both fail closed on a version counter newer than this build; a mixed-version context is the ordinary re-evaluation case and is accepted. The replay evaluator is deliberately narrower than the live one, so `compare_decisions()` checks directional soundness: a weakening is a known coverage limit, a strengthening (or an `IN_CONTRACT` ↔ `PROVEN_OUT_OF_CONTRACT` flip) is a defect.

**Deliberate deviation, stated rather than glossed:** the blocks are persisted with the *comparison* (the JSON report), not inside `AbiSnapshot` via `dumper.py`/`serialization.py`. The evidence is two-sided by construction (`side: old|new` per record), everything it records is derived from content the snapshot already carries, and a new snapshot field would mean a `serialization.SCHEMA_VERSION` bump inside the ADR-050 comparability contract. The replay guarantee is unaffected — the persisted comparison record carries the evidence, so no replayed decision depends on re-reading a binary. Full reasoning is recorded in the plan's Phase 4 section.

**Report schema 2.24 → 2.25**: `contract_evidence_refs` (per finding) and `contract_context` (top level), both *declared* in `compare_report.schema.json` rather than merely accepted by its `additionalProperties: true` — the gap that let `suppression_audit` ship undeclared at 2.24.

**Measurement, first run** (32 corpus cases, 41 findings per domain): 0 proven public-break losses, 0 unevidenced deltas, 0 unexplained fact losses. `public` resolves 71% of findings and proves 7 internal-noise findings out of contract; `all` produces 16 deltas (every legacy-demoted finding, by definition of the mode); `exports` is 100% unresolved on this corpus because its synthetic snapshots carry no export table — reported rather than hidden, since that is the honest per-domain signal the phase asks for. One gate test asserts the `public` domain proves *something* out of contract, so the loss baseline cannot pass vacuously.

## Checklist

- [x] Tests added / updated for new behaviour — 4 new test files (`test_contract_evidence_collect.py`, `test_contract_context_io.py`, `test_contract_replay.py`, `test_contract_shadow_measurement.py`) plus an end-to-end schema-declaration test in `test_cli_compare_contract_evaluation.py`
- [x] Changelog fragment added (`changelog.d/20260731_220000_claude_adr049_phases_3_4.md`)
- [x] Docs updated — plan Phase 3/4 sections, `AGENTS.md` module map, `scripts/CLAUDE.md` inventory, regenerated `cli-reference.md` / `mcp-tools-reference.md` / published schema mirror
- [x] `scripts/verify.py --profile pr`: lint, typecheck, ai-readiness, fp-rate, tier-accuracy, docs-contract, schema-sync, repo-facts, fair-metadata all pass; unit lane green at 96.25% coverage (floor 95%). `fmt-check` fails identically on `main` for files this PR does not touch (`abicheck/__init__.py`, `annotations.py`, …) — pre-existing, not introduced here
- [x] No new `mypy` or `ruff` errors introduced (`mypy abicheck/`: clean, 308 files)

https://claude.ai/code/session_01BED1UqfQYTbTdPsQiWH5pa

---
_Generated by [Claude Code](https://claude.ai/code/session_01BED1UqfQYTbTdPsQiWH5pa)_

## Summary by CodeRabbit

* **New Features**
  * Contract-evaluation reports now include evidence references for individual findings.
  * JSON reports can include replayable contract context with provider evidence, evaluation details, and decision receipts.
  * Added conservative replay and re-evaluation across supported contract modes, with version and evidence validation.
  * Added measurement tools for tracking contract-evaluation outcomes and evidence quality.

* **Documentation**
  * Updated CLI, MCP, architecture, schema, and planning documentation.

* **Tests**
  * Added comprehensive coverage for evidence collection, serialization, replay, re-evaluation, reporting, and measurement safeguards.


napetrov:
Pushed `6b867d8` addressing the review. Both P1s changed behaviour, not wording:

**Codex P1 — export evidence collected independently of the initial mode.** Fixed. `_apply_contract_evaluation_shadow` now computes both export surfaces for every opted-in comparison rather than only under `contract=exports`. The finding was exactly right: collecting conditionally on the *original* run's mode contradicted the policy-independence the block advertises — a `public` run's persisted context answered `UNKNOWN_UNRESOLVED` under an `exports` re-evaluation for a pair whose export tables were fully observable. New test: `test_public_run_can_be_reevaluated_under_exports`.

**Codex P1 — complete negative-proof evidence before exclusion.** Fixed, and the reasoning splits by domain. Under `public` the replay now never concludes `PROVEN_OUT_OF_CONTRACT` from graph non-membership: §4.3's negative proof wants positive private/system-header provenance *plus* every stronger-or-equal provider completed, and the block carries neither (no per-entity header origin; `configuration_coverage` is `NOT_STARTED` on every record this build writes). Under `exports` the exclusion branch stays, because there the provider's `COMPLETE` state *is* `ExportSurface.exclusion_is_provable` — an observed table, a resolved root, every root typed, no unaccounted export, no unresolved type edge — which is the terminal exclusion the ADR names. Tests pin both directions plus the incomplete-provider case.

**Codex P2 — receipt keys.** Fixed by keying on the report's own `finding_id`. That needed the id's implementation to move to the dependency-free `finding_identity` leaf module (`report_finding_id`), re-exported from `reporter_markdown` as `_finding_id` so no existing import path changes: importing the renderer from `checker` — even function-locally — closes a cycle the `import-cycle-growth` gate rejects, which it does count (confirmed by running it).

**Codex P2 — policy overrides.** Fixed; `PolicyFile.overrides` now reaches `resolved_config.policy.overrides` with provenance, and round-trips.

CodeRabbit's items are all in too: the `decision_receipt` gained its own `schema_version` (checked by the D6 gate, declared with `required` in the schema); `compare_decisions` gained the `disagreed` bucket the `_CLAIM_STRENGTH` comment promised (still fails `is_sound` — both evaluators share one `_NOT_APPLICABLE_KIND_SLUGS` set, so disagreeing there is a real inconsistency); `ALL` mode cites the header provider like the live path; `graph_by_side`/`record_by_side` are annotated; the reporter's `type: ignore` is replaced by a narrowing guard; the provider-walk is now one shared `persisted_domain_view()`; spellings resolve through a per-side `graph_node_index()`; required-key reads fail as `TypeError`; the dead `type_name` lookup is gone; the MCP docstring qualifies the block as `output_format="json"` only (reference regenerated); and the `checker` internal-namespace derivation is shared.

Skipped one: the nitpick asking `decision_receipt`'s siblings to gain a `required` list — done for `decision_receipt` itself along with its new version field, and the two siblings already had theirs.

`verify.py --profile pr`: 8/8 steps pass, 22286 tests, 96.60% coverage. `fmt-check` is excluded from that run because it fails identically on `main` for files this PR does not touch (`abicheck/__init__.py`, `annotations.py`, …).

---

napetrov:
Pushed `affb65c` for the second Codex round — all three were real:

**Audit-ledger `finding_id` (P2).** Correct, and the "cannot recompute it either" half is the sharper point: those serializers omit `old_value`/`new_value`, which are inputs to the id. Rather than adding the field to four ledger builders, `_add_contract_evaluation_fields` now emits it when absent — so wherever a contract decision goes, the receipt's key goes with it, and an ordinary `changes` entry keeps the id `_change_to_dict` already set. Test asserts a demoted finding's ledger entry joins to `decision_receipt.relevance_by_finding` *and* that the two agree on the relevance value.

**One-sided export citation (P2).** Also correct, and it was my error from the first round: I set the two-sided reason set before `exports` had its own decision path, then never revisited it. `_exports_mode_decision` picks one `ExportSurface` by D4's side rule and decides from it alone, so the second citation claimed evidence that was never read — worst exactly in your example, where the other side's provider is unavailable. The set is now per-domain (`_TWO_SIDED_REASONS_BY_MODE`): `public` keeps both sides because `classify_change_surface` genuinely runs against `surface_unions(surf_old, surf_new)`; `exports` and `all` cite the authoritative side only.

**Suppression config (P2).** Fixed the same way as the policy-override finding: `suppression_config_for()` reduces the live list through its own `rule_identities()` + `source_sha256`. One deliberate carve-out — a `SuppressionList` with no `source_sha256` (assembled in memory, never read from a file) still records `None`, because `SuppressionConfig` requires a digest precisely so a recorded source can be proved to be the one that ran, and a fabricated digest would be worse than recording nothing. Both cases are pinned by tests.

On codecov: the first `24.91%` reading was taken before the unit lanes reported; the current figure is 92.55% against a 90% target, and this push adds tests for every line it changes.

`verify.py --profile pr`: 7/7 steps pass, 22289 tests, 96.60% coverage.

---

napetrov:
Third review round addressed in `23f3d21`.

**1. The decision receipt was frozen before scoped stamping ran** (Codex P2, `checker.py:783`). Confirmed and fixed. `compare()` builds the receipt from its own `all_changes`, but `--used-by`/`--required-symbol` scoping runs *after* it returns: it overwrites already-recorded findings with the stronger explicit-scope `IN_CONTRACT` decision and synthesizes fresh `scoped_only_changes` the receipt never saw — so the persisted receipt disagreed with the report emitted beside it, and `replay_original_decisions()` reproduced decisions that were never the run's own. New `contract_evaluation.refresh_contract_receipt()` re-keys it, called from the same shared `stamp_scoped_result_findings()` traversal both the CLI and MCP front ends already route through, so neither can forget it. It **merges** rather than rebuilds: the receipt also covers the suppressed/redundant/out-of-surface findings that never reach `result.changes`, and recomputing from the emitted collections alone would silently drop them. Regression test asserts every emitted finding carrying a decision appears in the receipt with that same decision; it fails on the pre-fix code.

**2. The persisted graph lacked the owner-class edge** (Codex P2, `contract_evidence_collect.py:307`). Confirmed and fixed. Both live surfaces seed a method root's enclosing class — `surface._seed_public_roots` via `_type_identifiers(owner)`, `export_surface._seed_export_roots` via its exact-identity owner map — so a replay walking only return/parameter edges could find the owner unreachable and turn a live `IN_CONTRACT` into `PROVEN_OUT_OF_CONTRACT`, the one direction `compare_decisions()` treats as unsound. `build_type_graph` now links it, using `surface.py`'s permissive spelling resolution rather than `export_surface.py`'s exact-identity map: one graph serves both domains, and over-linking can only ever *weaken* a replayed decision, while under-linking is exactly the strengthening the edge exists to prevent.

**3. Overlays didn't reach the resolved configuration** (Codex P2, `contract_context.py:150`). Fixed. `--public-symbol`'s forced-public set and `--post-manifest`'s committed-export allowlist both decide membership (ADR-049 D2 counts overlay-selected roots as part of the `public` domain's roots), so omitting them described a run that never happened. New `overlay_selection()` populates `contract.overlays` and `surface.explicit_scope` — recovered from the run's own evidence ledger rather than re-read from the caller's arguments, so the two halves of one context cannot disagree. The digest combines each overlay's own `input_identity` instead of re-digesting the merged list: a combined list alone can't tell "one manifest naming A and B" from "two overlays naming one each".

**4. `graph_node_index` alias guard** (CodeRabbit). Fixed — it now requires the `alias:` source node to be present before following its edges, matching `resolve_graph_node`. Test covers both the agreement property on a real graph and the truncated-graph case where they diverged.

**5. `report_finding_id`'s `"\x1f"` delimiter** (CodeRabbit) — not taken, deliberately. The ambiguity it guards against requires a literal `\x1f` (ASCII unit separator) inside a kind slug, symbol, old/new value, source location, or description; with a fixed six-field join, no other input can produce a colliding key, and no producer in this codebase can emit that byte. Changing the encoding rehashes **every** finding id, which is a documented-stable, user-visible fingerprint (report schema 2.3) that consumers key waivers and cross-run correlation on. Certain cost, unreachable risk — the delimiter stays. Reasoning recorded in the plan doc alongside the fixes.

**6. Plan-doc wording** (CodeRabbit) — updated: `finding_identity.report_finding_id` is named as canonical, `reporter_markdown._finding_id` as the compatibility alias, with the cycle reason that moved it.

Gates: `scripts/verify.py --profile pr` — 12/14 passed, `docs-build`/`distribution-build` skipped for uninstalled modules locally, and `fmt-check` fails repo-wide on `main` too (no `[format]` section in `ruff.toml` vs. `E501` being ignored; CI's lint job runs `--only lint,typecheck,docs-build` and doesn't gate on it). `unit-pr` green including the 95% coverage floor; `ai-readiness` 0 errors, no import-cycle growth.

---

napetrov:
Fourth review round addressed in `15f0678`. Both new Codex P2s on `23f3d21` were real; the rest of the events on this PR are echoes of findings already fixed in `6b867d8`/`affb65c`/`23f3d21` and now marked resolved.

**1. Export roots matched by alias, not linker identity** (`contract_evidence_collect.py:450`). Confirmed. `_export_table_declarations` intersected each declaration's `_symbol_keys` against `ExportSurface.export_symbols`, which is alias-expanded so a *finding* naming any encoding can be looked up — the exact trap `export_surface._unresolved_type_edges` already documents avoiding for its own rootness test. A binary exporting the C symbol `foo` alongside an unexported `ns::foo` puts the bare tail `"foo"` in both declarations' key sets, so the unexported C++ declaration was persisted as an export root. `ExportSurface` now carries the `root_identities` set `_seed_export_roots` already computes, and the collector matches on linker identity against it. Regression test reproduces the exported-C-plus-unexported-C++ pair and fails on the pre-fix code.

**2. Overlays ignored on `public` re-evaluation** (`contract_context.py:314`). Confirmed. `persisted_domain_view` read only the `public_header` provider, so an entity kept because the user named it re-evaluated to `UNKNOWN_UNRESOLVED` even though the overlay's own evidence entry sits in the same block. Overlay manifests now contribute roots to the `public` domain, which is what ADR-049 D2 says ("roots selected by explicit overlays"). Two deliberate constraints: the manifest holds spellings, not node ids, so each is resolved through the same `graph_node_index` a finding's own spelling uses — one naming nothing the graph knows contributes no root rather than a guess; and it is `public`-only, since `exports`' roots are the observed table and a user's assertion is not an observation of the binary. Both are pinned by tests (`decl:hidden` reaching `record:Secret` in the receipt's closure under `public`; absent from `exports`' roots).

Because both fixes flow through `persisted_domain_view`, the receipt and the replay closure move together — they cannot disagree about the same domain.

Gates on `15f0678`: `scripts/verify.py --profile pr` — 11/14 passed, `docs-build`/`distribution-build` skipped for modules not installed locally, `fmt-check` fails repo-wide on `main` too (`ruff.toml` has no `[format]` section while `E501` is ignored; CI's lint job runs `--only lint,typecheck,docs-build` and does not gate on it). `unit-pr` green including the 95% coverage floor; `ai-readiness` 0 errors, no import-cycle growth; full fast suite 22286 passed.

---

napetrov:
Fifth review round addressed in `c6d7dca` — all three Codex P2s on `15f0678` were real, and two of them were regressions the previous round's overlay fix introduced.

**1. POST-manifest roots must stay exact** (`contract_context.py:351`). Confirmed. Folding overlays into the `public` domain's roots resolved *every* overlay spelling through `graph_node_index`'s alias tier, but the live evaluator matches `--post-manifest`'s allowlist against `Change.symbol` verbatim — so a manifest naming bare `foo` also rooted an unexported `ns::foo` carrying the same bare tail, turning a live `PROVEN_OUT_OF_CONTRACT` into a replayed `IN_CONTRACT`. `_CLAIM_STRENGTH` ranks those two equally, so `compare_decisions` classifies the flip as strengthened, i.e. unsound. `graph_node_index` gained a `follow_aliases` parameter and the tier is now selected per provider (`_OVERLAY_FOLLOWS_ALIASES`): exact for the manifest, alias-following for `--public-symbol`, each matching its own live semantics. Test pins both directions on one snapshot.

**2. Overlay-rooted decisions cited the wrong provider** (`contract_replay.py:318`). Confirmed. The root nodes were folded in but their provenance was dropped, so a declaration retained solely by `--public-symbol` came back citing `public_header:old` — evidence the decision never used, which is exactly the confusion `validate_decision_evidence` documents. `PersistedDomainView` now carries `overlay_roots_by_side` (record id → the nodes it contributed) and `_PersistedDomain.refs(side, nodes)` cites the matching overlay *alongside* the root provider — the closure walk that reached the entity is still the root provider's own work, so both belong. `refs` is taken after node resolution rather than before.

**3. An explicitly-empty POST manifest is a selected source** (`checker.py:715`). Confirmed and fixed in both places. `parse_manifest()` supports a manifest committing to zero exports, and `post_processing` branches on `is not None` for exactly that reason — the empty allowlist actively scopes every concrete export out. The truthiness checks in `checker._apply_contract_evaluation_shadow` and `collect_contract_evidence` collapsed it to "no manifest selected", so the ledger recorded no overlay and the resolved config denied it ran. Both are `is not None` now. No live decision moves: `contract_evaluation`'s own matching is `if public_surface_allowlist and …`, so an empty frozenset behaves identically there — only the ledger and the resolved configuration gain the fact.

**Not taken** (CodeRabbit, `tests/test_cli_compare_contract_evaluation.py:599`): annotating `tmp_path`/`monkeypatch` on the new test method. Every sibling method in that class takes them unannotated; annotating one would break the file's local convention for no behavioral gain.

Gates on `c6d7dca`: `verify.py --profile pr` — 11/14 passed, `docs-build`/`distribution-build` skipped for modules not installed locally, `fmt-check` fails repo-wide on `main` too (unchanged, and not gated by CI's lint job). `unit-pr` green including the 95% coverage floor, `ai-readiness` 0 errors, no import-cycle growth.

---

napetrov:
Sixth round, `a1a9ace` — one of the two Codex P2s fixed, the other needs a call I'd rather not make unilaterally.

**Fixed: a POST manifest narrows, it doesn't add** (`contract_context.py:387`). Confirmed. `--post-manifest` states the *whole* committed-export surface — `post_processing` scopes out a public declaration missing from it — but the overlay folding treated it as purely additive, so with public `api`/`kept` and a manifest naming only `kept`, `decl:api` stayed rooted and a live `PROVEN_OUT_OF_CONTRACT` replayed as `IN_CONTRACT`. A manifest now *replaces* the header roots for its side; `--public-symbol` still only ever widens whatever remains. Test pins both halves.

**Not yet fixed: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`). The finding is real — `ns1::Foo` and `ns2::Foo` share `RecordType.name == "Foo"` on the CastXML/clang path, so both collapse onto `record:Foo` and their field edges merge, which can turn a live ambiguous-identity `UNKNOWN_UNRESOLVED` into a replayed `IN_CONTRACT`. I'm holding off because the fix isn't local: the `decl:`/`record:`/`enum:`/`typedef:`/`alias:` encoding is the module's documented contract, `_TypeIndex` resolves every edge through it, and the persisted graph ships in the report under schema 2.25 — so re-keying records changes the graph in every persisted context, not just the ambiguous case. Two candidate approaches (key on `qualified_name or name` with the bare name demoted to an `alias:`, vs. persisting the ambiguous-name set and having replay degrade to `UNKNOWN_UNRESOLVED` on a hit) differ in blast radius and in which side of the conservative direction they err on. I've asked the author which they want before touching it.

Worth noting on the pattern: every round in this area has been a strengthening path opened by making replay *more* conclusive. The alternative worth considering is dropping overlay-root folding from re-evaluation entirely and keeping only the (inert) overlay persistence in `evaluation_context` — that trades a completeness gap (`UNKNOWN_UNRESOLVED` where the overlay would have said `IN_CONTRACT`) for no soundness risk at all, which is the direction this module's own contract prefers.

Gates on `a1a9ace`: `verify.py --profile pr` — 11/14 passed, `docs-build`/`distribution-build` skipped for modules not installed locally, `fmt-check` fails repo-wide on `main` too and isn't gated by CI's lint job. `unit-pr` green including the 95% coverage floor; `ai-readiness` 0 errors.

---

napetrov:
Seventh round, all three fixed in `695acee`.

**1. Provider availability now gates replayed membership** (`contract_replay.py`). Reproduced exactly as reported: an `elf_only_mode` snapshot leaves `PublicSurface.resolvable` false, so live answers `UNKNOWN_UNRESOLVED`/`UNAVAILABLE` at its `not auth.resolvable` gate — but the ledger entry still carries the declarations and type graph the collector built *before* bailing out, and the closure walk reached the entity through them and returned `IN_CONTRACT`/`COMPLETE`. `_PersistedDomain.domain_is_available` is the persisted counterpart, and it's one-to-one rather than an approximation: `contract_evidence_collect` writes `UNAVAILABLE` on exactly the `resolvable` check the live path branches on, for both providers. Regression test asserts `compare_decisions(...).is_sound` against the real live decisions, not just the replayed value.

**2. Overlay roots are direct matches, not closure seeds** (`contract_context.py`). Also confirmed: seeding the walk with a forced-public declaration pulled its whole signature closure in, so forcing `hidden_api(Secret *)` made a `Secret` finding `IN_CONTRACT` on replay against a live `PROVEN_OUT_OF_CONTRACT`. Both live overlay checks are per-finding overrides matched against the finding's own symbol, so overlay nodes stay *roots* (D2's "roots selected by explicit overlays", and still what the receipt reports as `evaluated_contract_roots`) but no longer seed the closure — that's the new `PersistedDomainView.closure_seeds_by_side`, the root provider's own declarations alone.

These two turned out to fix each other's edge. Live checks `force_public_symbols`/`public_surface_allowlist` *ahead* of its resolvability gate, so the replay does the same: a user-named entity still resolves `IN_CONTRACT` when the provider observed nothing, which is why fix 1 didn't cost the overlay coverage fix 2 preserves.

**3. `surface.internal_namespaces` provenance** — added, keyed to the policy file like `policy.overrides`. Only when the list is populated: `build_evaluation_context` receives a plain tuple, which can't distinguish "no policy file" from "a policy file that stated an empty list" (`PolicyFile.internal_namespaces_stated` is the field that can, and it doesn't reach here), and claiming a source for a possibly-unstated value is worse than claiming none. Tested both directions.

**Still not fixed, still waiting on your call: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`, from round six). Unchanged from what I wrote last round — the fix isn't local (the `decl:`/`record:`/`enum:`/`typedef:`/`alias:` encoding is the module's documented contract, `_TypeIndex` resolves every edge through it, and the persisted graph ships in the report under schema 2.25), so it wants a decision rather than a patch: key on `qualified_name or name` with the bare name demoted to `alias:`, or persist the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED`.

On the pattern I flagged last round — five consecutive rounds of strengthening paths in the overlay/replay area — this one does read differently. Rounds four through six were each "the previous fix opened a new way to over-claim"; these two are the general rule the case-by-case fixes were converging on (overlays override findings, they don't widen surfaces; a provider that observed nothing roots nothing), and both are now enforced structurally rather than per-case. I'd still take the offer from last round if you'd rather not carry overlay folding at all — but I no longer think it's the safer option by default.

`scripts/verify.py --profile pr`: 11 passed, `docs-build`/`distribution-build` skipped for uninstalled modules, and the pre-existing repo-wide `fmt-check` (not gated by CI's lint job). Full unit suite 22296 passed.

---

napetrov:
Eighth round, fixed in `0a863de`. Two substantive, three test/doc.

**A missing `decision_receipt` now fails closed** (`contract_context_io.py`). `persisted_context_from_dict` defaulted an absent block to an empty version-1 receipt, so `replay_original_decisions()` on a truncated payload returned `{}` — indistinguishable from a comparison that genuinely recorded no decisions. The key is required now, like its two siblings; same rule the version ceiling follows.

**`surface.internal_namespaces` provenance survives an explicitly empty list.** You were right that `internal_namespaces_stated` is a real distinction I dropped rather than one that couldn't reach the assembler — `build_evaluation_context` now takes it alongside the tuple and `checker.compare` forwards it from the `PolicyFile`. A policy file that writes `internal_namespaces: []` keeps its provenance; a value no policy file contributed still claims no source. Both cases tested.

Three smaller ones: the overlay test now asserts the exact `UNKNOWN_UNRESOLVED` for the collateral type (`PROVEN_OUT_OF_CONTRACT` would have passed a `not IN_CONTRACT` check while over-claiming exactly what that domain can't support) and no longer shadows the `secret` record with a `Change`; the POST-manifest evidence assertion checks the two sides instead of a count; and the plan's wording now says the provider reported the domain unavailable rather than that it observed nothing.

**Already fixed on HEAD, re-verified against the current code** — these were reported against `66d34b5`/`6b867d8` and the events replayed after the fixes landed:
- `checker.py:679` (P1, export evidence collected only under `contract=exports`) — export providers are collected for every opted-in comparison since `695acee`'s predecessor; the reasoning is in the comment at that call site.
- `contract_replay.py:181` (P1, graph-only `public` exclusions) — `can_prove_exclusion` is `exports`-only; `public` non-membership returns `UNKNOWN_UNRESOLVED`, with the `configuration_coverage`/provenance argument written out in its docstring.
- `contract_evidence_collect.py:846` (cite only the authoritative side) — two-sided citation is scoped by `_TWO_SIDED_REASONS_BY_MODE`; `exports` cites one side.
- `contract_evidence_collect.py:307` (owner-class edges) — `_link_owner_class` is in `build_type_graph`.
- Overlays in the resolved config, policy overrides, suppression config, scoped-receipt refresh, `graph_node_index` alias guard, receipt versioning, MCP docstring, `reporter.py` type suppression — all landed in earlier rounds.

**Still open, still waiting on you: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`). Unchanged — the fix isn't local (persisted graph encoding under schema 2.25, every edge resolved through `_TypeIndex`), so it wants a decision: key on `qualified_name or name` with the bare name demoted to `alias:`, or persist the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED`.

`scripts/verify.py --profile pr`: 11 passed, `docs-build`/`distribution-build` skipped for uninstalled modules, plus the pre-existing repo-wide `fmt-check` that CI's lint job doesn't gate. Full unit suite 22297 passed. Codecov patch coverage is back above the 90% target as of the last few rounds.

---

napetrov:
Ninth round, all three fixed in `5fce76e`. They're one finding wearing three hats: **the replay resolver was looser than the live matcher it stands in for**, and every looseness was a strengthening path. Each fix is "mirror what live already does" rather than a new rule.

**1. The bare `::` tail is `public`-only now** (`contract_replay.py`). The live exports matcher passes `allow_tail_fallback=False` for exactly the `ns::foo` / exported-C-`foo` collision — offering the tail as a replay spelling reintroduced it one layer down, at node resolution, after the export-root fix had correctly removed it at the provider layer. `_entity_spellings` takes the mode now and mirrors `_symbol_matches`'s own per-mode argument.

**2. `caused_by_type` is consulted for type- and member-level findings only.** Mirrors `_exports_mode_decision`'s `type_scoped` gate and its stated reason: closure membership answers a *type*-level question, so a symbol-level finding on an unexported helper must not land in contract because its `caused_by_type` is one some other exported signature reaches. Applied in both modes — being stricter than the live matcher can only weaken.

**3. Owner edges match by exact record identity** (`contract_evidence_collect.py`). You're right that my justification for the permissive lookup was wrong, and specifically about *why*: I argued over-linking only ever weakens, which ignores that one graph serves both domains. `api::run()` yields the owner string `"api"` — namespace noise, not a class — and permissive resolution matched it against an unrelated `other::api` by bare tail, into the export closure. `_TypeIndex.resolve_exact_record` mirrors `export_surface._seed_export_roots`'s `owner_seed_by_identity` map. Exact matching costs nothing real: `owner_class_of` always reconstructs a complete scope chain from a qualified name or an Itanium/MSVC mangling, so a real class matches exactly and a non-method's namespace noise correctly matches nothing — the same conclusion `type_reachability.py`'s owner seeding reached for the same reason.

All three regression tests verified failing against the previous code before the fix went in.

Worth saying plainly after nine rounds: the overlay/replay work converged two rounds ago on "overlays override findings, they don't widen surfaces," and this round converges the *resolver* on the matching rule to go with it — replay may be stricter than live, never looser. Both are now enforced structurally rather than per-case, which is why I think this class is closed rather than merely quiet.

**Still open, still yours to call: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`) — unchanged, and now the only outstanding item. Key on `qualified_name or name` with the bare name demoted to `alias:`, or persist the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED`. Happy to do either; it needs a decision because the encoding ships in the report under schema 2.25.

`scripts/verify.py --profile pr`: 11 passed, `docs-build`/`distribution-build` skipped for uninstalled modules, plus the pre-existing repo-wide `fmt-check` CI's lint job doesn't gate.

---

napetrov:
Tenth round, both fixed in `97e2514`. Same rule as the ninth round — **replay may be stricter than the live matcher, never looser** — and one of the two is the literal mirror image of a fix from that round.

**1. `contract_replay.py:380` — export alias pruning.** Confirmed and fixed. `compute_export_surface` already does this pruning on its own root set (`export_symbols -= nonroot_keys - matched_exports - root_identities`), so live answers `PROVEN_OUT_OF_CONTRACT` for a finding on the unexported C `foo`; the replay resolver went through the persisted graph's `alias:` edges, reached `decl:_ZN2ns3fooEv`, found it in the export closure, and said `IN_CONTRACT`. `_PersistedDomain.resolve` now drops an export root reached only through a spelling that a **declaration** outside the root set also answers to, keeping live's two exemptions: a root's own canonical node identity, and record/enum nodes (live's `nonroot_keys` is built from declarations alone, so a same-named type must not prune a genuine root). Worth naming explicitly: this and the round-nine tail-fallback fix are the two halves of one collision — there the finding was qualified and the root bare, here the finding is bare and the root qualified.

**2. `contract_replay.py:364` — POST manifest in `all` mode.** Confirmed and fixed, including the empty-manifest case. `persisted_domain_view` only collected overlay providers under `PUBLIC`, so `all` never saw the manifest at all; live checks `_REASON_POST_MANIFEST_NOT_COMMITTED` *ahead* of its own `all`-mode shortcut, exactly because a committed-export manifest is an exact closed-domain observation rather than the header-origin classification `all` drops. The manifest's spellings are now read in every mode (`PersistedDomainView.post_manifest_by_side`), and an `all`-mode finding it could not have admitted answers `UNKNOWN_UNRESOLVED`.

Two things about that second fix are deliberate and worth flagging rather than leaving to be rediscovered:

- **It weakens rather than reproducing the live `PROVEN_OUT_OF_CONTRACT`.** The replay reproduces `_run_allowlist`'s keep conditions only approximately, so it knows the finding is not demonstrably in contract without being able to prove the exclusion.
- **Two of those conditions are left out on purpose, both in the weakening direction.** The concrete-export test (`sym in export_ids`) is not reproduced because the persisted export provider's declarations are *not* `_snapshot_export_ids` — it reads the raw platform tables — so consulting the provider could answer "not a concrete export, therefore kept" for a symbol live did demote, which is a strengthening. And `--public-symbol`'s rescue is not honored, because live honors it only when header scoping is *also* on, and the persisted context records the resolved contract mode rather than that flag.

Both reproduce as strengthenings before the fix (verified by `git checkout --`-ing the two source files and re-running: `test_exported_alias_cannot_root_an_unexported_namesake` and `test_all_mode_honors_a_persisted_post_manifest` both fail, the latter with `strengthened=('27581ef21a441ec9',)` against real live decisions). `scripts/verify.py --profile pr` is clean apart from the repo-wide pre-existing `fmt-check` (no `[format]` section in `ruff.toml`; CI's lint job runs `--only lint,typecheck,docs-build`, so it isn't gated).

**Still open, still yours to call: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`) — unchanged from the ninth round, and still the only outstanding item. `ns1::Foo` and `ns2::Foo` merge into one node. Either key on `qualified_name or name` with the bare name demoted to `alias:`, or persist the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED`. Happy to do either; it needs a decision because the encoding ships in the report under schema 2.25.

---

napetrov:
Eleventh round, both fixed in `0680b6a`. Unlike the last few, these are **joinability** gaps, not soundness ones — nothing replayed a stronger claim; the receipt and the report just couldn't be correlated.

**1. `contract_evaluation.py:1729` — missing-contract labels absent from the receipt.** Confirmed and fixed, and it had a second half Codex's note implies but doesn't state: the synthesized entry carried no `finding_id` at all, so even with a receipt key there would have been nothing on the report side to join it to (it never routes through `_change_to_dict`, which is what sets the id for an ordinary entry). Both halves come from the same root cause — a missing contract member has no backing `Change`, so each format hand-builds its own dict — and are fixed by giving that event one identity: `finding_identity.missing_contract_finding()` returns the `Change`-shaped identity, **including the description**, which is one of `report_finding_id`'s own hash inputs. The CLI JSON fold, the MCP tool, and `refresh_contract_receipt` all key from it now, so they agree by construction rather than by three hand-copied literals happening to match. `missing_contract_kind()` moved next to it for the same reason, folding the four independent `gate_scope` → slug derivations (reporter_markdown, sarif ×2, mcp_server) into one.

**2. `contract_context_io.py:621` — `relevance_by_finding` defaulted to `{}`.** Confirmed and fixed; same fail-closed rule the enclosing `decision_receipt` block got in the eighth round, one level down. `schema_version` stays defaulted on purpose and I'd rather not change that: an absent counter has a defined meaning (version 1, per its own comment), where an absent decision map does not.

Fail-without-fix verified for both by `git checkout --`-ing the three source files and re-running: `test_scoped_stamping_refreshes_the_persisted_receipt` and `test_missing_relevance_map_raises_type_error` both fail. The first needed a real fixture change to reproduce at all — the existing one's only missing symbol was already covered by a `CONSUMER_REQUIRED_SYMBOL_REMOVED` finding, so `uncovered_missing_symbols` correctly deduped it away and no synthetic entry was ever produced; it now carries a second, genuinely uncovered symbol, and the assertion no longer filters on `finding_id` (filtering on it was exactly what hid the unjoinable entry).

Also in this commit, from CodeRabbit: the three exports-domain replay tests now assert the exact `PROVEN_OUT_OF_CONTRACT` rather than "not `IN_CONTRACT`" (agreed — those four values make different claims, as an earlier round's own comment says), and `test_scoped_stamping_refreshes_the_persisted_receipt` gained its parameter annotations.

`scripts/verify.py --profile pr` is clean; `mypy abicheck/` is back to zero after the first pass of this fix tripped it (a repo-wide `ruff format` on `mcp_server.py` relocated a pre-existing `# type: ignore[attr-defined]` onto the wrong line — reverted, only the two intended hunks are in the diff). The one remaining failure is the repo-wide pre-existing `fmt-check`, which CI's lint job doesn't run.

**Still open, still yours to call: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`) — unchanged, and still the only outstanding item. Key on `qualified_name or name` with the bare name demoted to `alias:`, or persist the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED`. Happy to do either; it needs a decision because the encoding ships in the report under schema 2.25.

---

napetrov:
Fixed in `d021681` — a third absent-vs-empty slip, one layer over from the last two.

`evidence_refs_for_change` gated its `--post-manifest` attribution on the allowlist's *truthiness*, so a manifest that validly commits to zero exports scoped everything out and then had the resulting exclusions cite `public_header` — the provider that decided nothing there. It tests `is not None` now, matching what `collect_contract_evidence` already records for the same input. The `IN_CONTRACT` half of that branch is unaffected either way (nothing is a member of an empty set), so only the exclusion path changes.

Fail-without-fix verified: `test_an_empty_manifest_still_owns_its_exclusions` fails on the previous commit with `'public_header:old' != 'post_manifest:old'`. `--profile pr` clean apart from the repo-wide pre-existing `fmt-check`.

On the **codecov/patch** check that keeps posting: it's measuring patch coverage against a 90% target and has been red for the whole PR; the repo's own gate is `verify.py`'s 95% **line+branch floor** on the canonical unit lane, which passes. Each round has been moving patch coverage up (25% → 82% → 94%) as the fixes brought tests with them; I'm not going to chase the remaining few points with tests that don't correspond to a real behaviour, but say the word if you want that check green before merge.

**Still open, still yours to call: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`) — unchanged, and still the only outstanding item. Key on `qualified_name or name` with the bare name demoted to `alias:`, or persist the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED`. Happy to do either; it needs a decision because the encoding ships in the report under schema 2.25.

---

napetrov:
Both round-13 findings were real, and both are the same absent-vs-fabricated shape as the last few rounds — just moved from the evidence ledger into the resolved configuration itself. Fixed in `314b185`.

**1. The persisted gate was the built-in default, not the one the run was scored with** (`contract_context.py:260`)

Confirmed with a repro before touching anything: a plain `compare old.json new.json --contract-evaluation` exits **4** on the legacy floor and persists `"exit_code_scheme": "severity"`; `--severity-abi-breaking warning` runs and persists `abi_breaking: "error"`. Both are wrong claims rather than omissions — `GateConfig()`'s defaults happen to be the severity scheme and the built-in levels, so the block asserts them for every run.

Root cause is exactly as you described: `checker.compare` never sees the gate. `cli_compare_helpers` resolves `resolved_cfg.exit_code_scheme`/`sev_config` at line 1258 and applies them after `compare_snapshots()` returns at 1673.

I took the *refresh-before-serialization* route rather than threading gate parameters through the Tier-2/Tier-1 boundary. Adding `exit_code_scheme`/`severity` kwargs to `service.compare_snapshots` → `checker.compare` would put gating vocabulary into a core verb that does no gating, and that plumbing is Phase 5's job (route every front end through one already-resolved config) — building a private version of it here would be the speculative half-step the plan explicitly defers. So: new `contract_context.with_resolved_gate()`, called once from `cli_compare_helpers` right after `compare_snapshots()` and before any report is rendered.

Provenance is threaded too, and this is the one place it can be *more* precise than the core verb's `API_REQUEST` under-claim — the front end genuinely observes which layer selected each field:

| run | `gate.exit_code_scheme` | `gate.severity` |
|---|---|---|
| no flags, no config | `legacy` / `built_in_default` | defaults / `built_in_default` |
| `--severity-abi-breaking warning` | `severity` / `built_in_default` | `warning` / `explicit_cli` |
| `--exit-code-scheme legacy` | `legacy` / `explicit_cli` | — |

The second row's `built_in_default` on the scheme is deliberate and correct: `auto` is the built-in default and *it* resolved to `severity` because severity became active — no one selected the scheme. Verified the JSON round-trips byte-identically through `persisted_context_from_dict`/`to_dict`.

Nothing about a relevance decision, closure, or per-finding receipt entry changes — the gate stays `NOT_APPLICABLE` to contract membership. This only fixes what the audit record says about how the run was gated.

**2. A digest-less suppression list is still a selected source** (`contract_context.py:289`)

Also real, and worse than the docstring implied. It wasn't only hand-built `SuppressionList([...])`: `merge()` drops *both* halves' digests even when each was loaded from a file, and `compat/_helpers.py` builds every ABICC `-skip-symbols`/`-skip-internal-*` list through the digest-less constructor. So a run actively suppressing findings persisted "no suppression source was selected at all" — the exact conflation `SuppressionConfig` exists to prevent, one layer up.

Took your first suggestion: the digest falls back to `content_digest(rule_identities())`. I don't think that counts as the fabricated digest the old docstring was refusing — it authenticates the rules that actually ran, those identities are persisted verbatim in the same block, it uses the ledger's own `content_digest` convention, and I checked that nothing in the codebase re-reads a suppression file to verify this field against its bytes (the only readers are `contract_context_io`'s serializer and `compatibility_evaluation_frontend`, which pass it through). The file-digest and rules-digest cases differ in *what content* is authenticated, and both authenticate content the receipt carries.

**Verification.** Four regression tests, each confirmed to fail without the fix (copied the sources aside, `git checkout --`, re-ran, restored):

- `test_persisted_gate_is_the_one_the_run_was_scored_with`
- `test_explicit_exit_code_scheme_records_its_own_provenance`
- `test_a_digestless_suppression_list_is_still_a_selected_source`
- `test_a_merged_suppression_list_keeps_its_rules`

`scripts/verify.py --profile pr`: `lint`, `typecheck`, `ai-readiness`, `unit-pr` (incl. the 95% line+branch floor), `fp-rate`, `tier-accuracy`, `usecase-docs-sync`, `docs-contract`, `repo-facts`, `schema-sync`, `fair-metadata` all pass. `mypy abicheck/` is clean. (`fmt-check` fails repo-wide in this environment on `main` too — no `[format]` section in `ruff.toml` — and CI's lint job runs `--only lint,typecheck,docs-build`, so it isn't gated. I checked `ruff format --diff` on every file I touched: none of my lines appear in it.)

**Still open, still yours to call: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`) — unchanged, and still the only outstanding item. `ns1::Foo` and `ns2::Foo` merge into one persisted node. The two fixes I'd pick between are keying on `qualified_name or name` with the bare name demoted to an `alias:` tier, or persisting the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED` rather than guessing. I've held off because the encoding ships in the report under schema 2.25, so changing it is a consumer-visible decision rather than a review fix.

---

napetrov:
Both round-14 findings were real, and both are the *previous* round's fixes missing a sibling one module over. Fixed in `47d82f5`.

**1. The MCP front end has the identical gate defect** (`mcp_server.py:769`)

Correct, and I should have caught it when I fixed the CLI — `abi_compare` resolves its own scheme and severity the same way (any `severity_*` argument opts into the severity-aware scheme) and rendered the context without refreshing it. Reproduced before fixing:

| call | `exit_code` / scheme | persisted gate (before) |
|---|---|---|
| defaults | 4 / `legacy` | `severity`, `abi_breaking: error` |
| `severity_abi_breaking="warning"` | 0 / `severity` | `severity`, `abi_breaking: **error**` |

It calls the same `with_resolved_gate()` now, with two front-end-specific differences that are deliberate:

- **`API_REQUEST`, not `EXPLICIT_CLI`.** This is a typed API caller; it cannot claim a CLI option was typed. That's D7's own peer layer for exactly this, and it's the same under-claim `checker.compare` already makes for its own arguments.
- **A `"scoped"` scheme resolves back to its underlying one.** `--used-by`/`--required-symbol` reports `exit_code_scheme: "scoped"`, which isn't one of the two values `GateConfig` accepts (`__post_init__` rejects anything but `legacy`/`severity`), so `result.scoped_exit_code_scheme` — the scheme the scoped code was actually computed under — is what gets recorded.

One thing worth noting since it looks odd in the output: for the severity call, `gate.exit_code_scheme` reads `api_request`, not `built_in_default` the way the CLI's equivalent does. That's correct rather than inconsistent — MCP has no `auto` scheme at all, so the caller's severity argument is directly what selected `severity`; the CLI's `auto` is a built-in default that resolved itself.

**2. The shadow gate measured two of five collections** (`measure_contract_shadow.py:219`)

Also correct. `checker._apply_contract_evaluation_shadow` stamps and records `suppressed_changes`, `redundant_changes`, and `reconciled_changes` too — its own docstring documents each, one per Codex round — so a regression dropping a decision or a receipt entry for an audit-bucket finding left `fact_losses` at zero. All five are walked now.

The three audit buckets get their own `delta_matrix` rows rather than being folded into `out_of_surface`, and `_is_delta` answers `False` for them on purpose: a suppression, a display dedup, and a build-context reconciliation are policy/presentation decisions about an *already-decided* finding, not claims about contract membership, so there is no legacy contract claim for a shadow decision to differ from. They're measured (so a lost decision still fails `fact_losses`), never counted as a delta.

**What I did not do, and why:** add corpus cases that populate those buckets. A suppression case means threading a `SuppressionList` through `check_fp_rate.CASES` — the *shared* corpus fixture whose own FP-rate gate would need re-validating against the changed shape — and redundancy and reconciliation are produced by the pipeline rather than selected by corpus input, so there's no corpus-level knob for them at all. The three rows stay empty on today's corpus and the gate goes live for them the moment such a case exists, rather than the moment someone remembers to add the collection. Recorded in the plan alongside the fix.

**Verification.** Three regression tests, each confirmed to fail without the fix. Since the corpus can't populate the new buckets, `test_audit_bucket_findings_are_measured_too` drives them directly: it wraps `compare` to relocate one already-decided finding into each bucket in turn and asserts the finding count is unchanged, `fact_losses` stays empty, and the decision lands in that bucket's own row.

`scripts/verify.py --profile pr`: `lint`, `typecheck`, `ai-readiness`, `unit-pr` (incl. the 95% floor), `fp-rate`, `tier-accuracy`, `usecase-docs-sync`, `docs-contract`, `repo-facts`, `schema-sync`, `fair-metadata` all pass; `mypy abicheck/` clean; `measure_contract_shadow.py` still reports 0/0/0 on baseline. (`fmt-check` fails repo-wide in this environment on `main` too — CI's lint job doesn't run it. `mcp_server.py` and `test_mcp_server_coverage.py` were already unformatted before this change; I checked `ruff format --diff` on both and none of my lines appear in it.)

**Still open, still yours to call: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`) — unchanged, and still the only outstanding item. `ns1::Foo` and `ns2::Foo` merge into one persisted node. Either keying on `qualified_name or name` with the bare name demoted to an `alias:` tier, or persisting the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED`. Held because the encoding ships in the report under schema 2.25, so it's a consumer-visible decision rather than a review fix.

---

napetrov:
Both round-15 findings were real, both reproduced end-to-end against a live run before fixing, and the second one turned up a third defect neither review named. Fixed in `5029f9a`.

**1. `contract_evidence_collect.py:216` — unmangled overloads shared one `decl:` node.** Confirmed with a minimal snapshot: a public `over()` and a private `over(Secret *)`, neither carrying a mangled symbol, both landed on `decl:over`, so the public root inherited the private overload's edge to `Secret` and re-evaluation returned `IN_CONTRACT` for a private `Secret` layout change the live run returned `PROVEN_OUT_OF_CONTRACT` for — `strengthened`, not `weakened`.

Took the discriminator route. The node key falls back to `name + "(params)->return"` for the one case where the plain name names more than one *unmangled* declaration — the same "most specific available identity, ambiguity-safe" tiering `finding_identity.py` and `diff_helpers.TypeMap` already use. Three things worth stating about the shape:

- It is a **tie-break, not a new encoding**. `contract_replay._without_shared_export_aliases` exempts a spelling that is a root's *own* node identity from its alias pruning, so refining an unambiguous `over` to `over()->int` unconditionally would lose that exemption and prune a genuine export root — the same strengthening, one corner over.
- Two unmangled declarations agreeing on name, parameters *and* return type still share a node. That is lossless (their signature edges and their `owner_class_of` result are derived from exactly those fields), so the residual case needs no ledger-level ambiguity flag — the alternative you offered isn't required once the discriminator covers the separable cases.
- Rootness still keys on **linker** identity, which is what the export table matched. Every member of an unmangled overload group shares one, so they are rooted or not as a group — exactly as live roots them.

**2. `contract_evidence_collect.py:275` — the owner identity set missed `qualified_name`.** Confirmed: `name="Widget"`/`qualified_name="ns::Widget"` with an exported `ns::Widget::draw` gave live `IN_CONTRACT` and replay `PROVEN_OUT_OF_CONTRACT`.

Rather than adding `qualified_name` to the existing set, the set is now a field-for-field mirror of `export_surface.compute_export_surface`'s own `owner_seed_by_identity` map — the map the live exports closure actually seeds through — including its typedef-collision ambiguity union. Re-deriving it was the root cause; mirroring it also closed a **third defect neither review flagged**: the old set registered a *leaf-only* bare `name`, so an `api::run()` namespace fragment matched an unrelated `other::api` "exactly" after all — precisely the collision the exact-match rule exists to close, reopened from the record side. That one has its own regression test (`test_a_leaf_only_bare_name_still_seeds_no_owner`), which fails on `47d82f5`.

**Observed while mirroring that map, deliberately not fixed here.** The graph's general *type-reference* resolution still goes through `surface._index_surface_types`'s un-augmented index (keyed on `name` and its `::` tail), while `compute_export_surface` augments its own copy with every `qualified_name`. A field spelled `ns::Widget` on the castxml/clang path therefore reaches the record live-in-`exports` but produces no edge here. Fixing it changes the `public` closure too, whose live index is *not* augmented — and over-linking there flips a live `PROVEN_OUT_OF_CONTRACT` to a replayed `IN_CONTRACT`, which `compare_decisions` counts as strengthening just as under-linking does. Getting both domains right at once needs its own scoped design, not a drive-by extension of an owner-edge fix. Recorded in the plan doc's eleventh-round section.

**Verification.** Four new regression tests (three graph-level in `test_contract_evidence_collect.py`, two end-to-end `compare_decisions(...).strengthened == ()` in a new `TestNodeIdentitySoundness`); all fail on `47d82f5` and pass on `5029f9a`. `scripts/verify.py --profile pr`: `lint`, `typecheck`, `ai-readiness`, `unit-pr`, `fp-rate`, `tier-accuracy`, `usecase-docs-sync`, `docs-contract`, `repo-facts`, `schema-sync`, `fair-metadata` all pass; only the known repo-wide `fmt-check` failure (no `[format]` section in `ruff.toml`, and CI's lint job runs `--only lint,typecheck,docs-build`, so it is ungated) — my own files are clean under `ruff format --check` both before and after. `mypy abicheck/` clean.

**Still open, still yours to call: bare `record:<name>` node identity** (`contract_evidence_collect.py:284`) — `ns1::Foo` and `ns2::Foo` still merge into one node. Note that finding 2 above is adjacent but does *not* resolve it: the owner map now recognizes a qualified identity, but the node it seeds is still keyed by the bare `name`. The two options from my earlier comment stand — key on `qualified_name or name` (bare name demoted to `alias:`), or persist the ambiguous-name set so replay degrades to `UNKNOWN_UNRESOLVED`.

---

napetrov:
All three round-16 findings were real. Fixed in `4e6fa70`.

**1. `contract_replay.py:503` — a spelling resolving to an unrelated node *kind*.** Reproduced end-to-end exactly as described (PE snapshot: exported `api(foo *)`, reachable `struct foo`, unexported function `foo`): live `PROVEN_OUT_OF_CONTRACT`, replay `IN_CONTRACT`.

Root cause is structural rather than a missing case: this module resolves every spelling through one flat index, while the live evaluator asks two separate questions of two different node kinds — `_symbol_matches` against `export_symbols` (declarations) and `_confirmed_type_matches` against `export_types` (types). `_entity_spellings` is now `_entity_lookups`, returning `(spelling, admissible node kinds)`:

- a **symbol** reaches a `decl:` node only under `_symbol_matches`'s own first-branch rule — for a type-level kind that means the symbol must be *mangled*, which also closes the mirror-image leak your example doesn't name (a `type_size_changed` on `Widget` borrowing an exported *function* called `Widget`);
- a symbol reaches `record:`/`enum:`/`typedef:` only when the finding is type-scoped at all;
- `caused_by_type` — a type name by construction — reaches type nodes alone.

One deliberate asymmetry: membership (closure hit, and the "unplaceable" gate) uses the kind-filtered set, but the **overlay override and evidence attribution stay on the full resolution**. `force_public_symbols` is matched live by `_change_matches_symbols`, which asks nothing about node kind, so filtering there would have made the replay stricter than live on a path that is supposed to mirror it exactly.

**2. `contract_context.py:237` — `policy.base` naming the ignored argument.** Correct: `checker.compare` computes `effective_policy = policy_file.base_policy if policy_file is not None else policy`, so with a file the `policy` argument is not merged, it is *replaced*. Threaded a `policy_from_file` flag and the receipt now records `policy_file`, the same convention `policy.overrides` and `surface.internal_namespaces` already use for values from that same file. Regression test asserts both directions (file present → `policy_file`; no file → `policy`).

**3. `contract_context.py:302` — collapsed severity provenance.** Also correct, and worse than just a wrong label: the aggregate key `gate.severity` is not one the canonical resolver has at all — `compatibility_evaluation_frontend.SEVERITY_CATEGORY_FIELDS` tracks `gate.severity.<category>`. `with_resolved_gate` now takes a per-category mapping and writes those keys. The CLI derives each category's layer independently (`--severity-preset` counts as a typed flag for all four, since it sets all four); the MCP tool records all four uniformly, which is honest there — one `SeverityConfig` object argument genuinely is one layer for every category. The mixed case you named is pinned by a test: `--severity-abi-breaking warning` gives `abi_breaking → explicit_cli` and `addition → project_config` in the same run.

**Verification.** Three new regression tests, all failing on `5029f9a` and passing on `4e6fa70`. `scripts/verify.py --profile pr`: `lint`, `typecheck`, `ai-readiness`, `unit-pr`, `fp-rate`, `tier-accuracy`, `usecase-docs-sync`, `docs-contract`, `repo-facts`, `schema-sync`, `fair-metadata` all pass; only the known repo-wide `fmt-check` failure (no `[format]` section in `ruff.toml`; CI's lint job runs `--only lint,typecheck,docs-build`, so it is ungated) — every file I touched is clean under `ruff format --check`, and the three that aren't were already unformatted before this branch, so I hand-formatted only my own lines rather than reflowing them. `mypy abicheck/` clean.

**Still open, still yours to call: bare `record:<name>` node identity** (`contract_evidence_collect.py`) — `ns1::Foo` and `ns2::Foo` merge into one node. Unchanged, and still the only outstanding item.

---

napetrov:
`2637eec` closes the **last standing item on this branch** — bare `record:<name>` node identity — plus the two round-17 findings.

**Bare `record:<name>` identity (the item that was waiting on a decision).** Taking it now rather than continuing to hold it. Reproduced first: `ns1::Foo` and `ns2::Foo`, both spelled bare `Foo` by a castxml/clang producer, collapsed onto one node with their field edges unioned — live `UNKNOWN_UNRESOLVED`, replay `IN_CONTRACT`.

Went with the first of the two options I'd posted: identity is now `qualified_name or name` (`_type_identity`), bare leaf demoted to an `alias:` spelling. The reason the second option (persist the ambiguous-name set) turned out unnecessary is worth recording — **the graph already carries the ambiguity once the nodes are separate**: two records answering to one spelling *are* two nodes, so the collision is visible in the resolution itself, with no new persisted field and no schema change.

Worth stating explicitly, since narrowing a closure is normally the dangerous direction: **this one cannot strengthen.** A snapshot with an ambiguous bare tail already reports `identity_coverage=PARTIAL`, and `can_prove_exclusion` requires `COMPLETE` — so the only conclusive negative is already off the table for exactly the snapshots this changes.

**`contract_replay.py:602` — gate membership on identity coverage.** Real, and separating the nodes was necessary but *not* sufficient — checking your exact shape (a global `Foo` beside a namespaced `ns::Foo`, DWARF-style) confirmed it still strengthened after the node split. Looking at `_confirmed_type_matches` closely, live rejects on **two** clauses, and the second is the one at work here: *"a qualified name whose own trailing tail is ambiguous"*. A finding on `ns::Foo` resolves to exactly one node, but an exported signature spelling the bare `Foo` linked *both* records, so the closure hit proves nothing. Both clauses are now mirrored.

That second clause needed one collection-side change to be checkable at all: the bare `::` tail of a qualified identity is now an `alias:` spelling in the persisted graph. With a DWARF producer — which bakes the whole path into `name` — nothing else recorded it, so the collision was simply invisible on replay.

I did **not** take the blunter option you offered first (require `identity_coverage COMPLETE` before accepting any membership): that would kill positive membership for every finding on a side that has a single ambiguous name anywhere. Your second suggestion — "retain enough per-spelling ambiguity data to reject only ambiguous matches" — is what landed, and it needs no extra persisted data. Rootness is still decided *before* ambiguity, matching live's own order: a declaration that is a root answers its own membership by its own linker identity, which `_exports_mode_decision` returns on before it computes any type candidate.

**`contract_context.py:237` — CLI source for contract mode.** Confirmed: a typed `--contract exports` recorded `api_request` / `checker.compare`. Added `with_field_provenance`, the value-free counterpart of `with_resolved_gate` (the core verb resolved the *value* correctly; only the front end knows the *source*). The CLI refreshes `contract.mode` only when the flag was really typed, so the `LEGACY_ALIAS` provenance `resolve_legacy_contract_mode` records for `--scope-public-headers` survives untouched — pinned by a test in both directions. Also renamed `_gate_provenance` → `_cli_field_provenance`, since it is no longer gate-only.

**Verification.** Four scratch repros (the two round-15 cases, the round-16 kind collision, and both shapes of this round) all now agree with live and none strengthen. Three new regression tests, each confirmed failing with only the source stashed. `scripts/verify.py --profile pr`: everything green except the known repo-wide `fmt-check` (ungated in CI; my files are clean under `ruff format --check`). `mypy abicheck/` clean.

**Nothing is outstanding on my side now** — no held items, no known unfixed findings.

---

napetrov:
All three round-18 findings were real. Fixed in `475e0ba`.

**1. `contract_replay.py:681` — duplicate records sharing one canonical identity.** Reproduced. This one is worth stating plainly because it contradicts something I asserted in a docstring last round: `_index_surface_types` tallies record **objects** per name (`combined_counts[name] += len(entries)`), not distinct identities — so two `Foo` declarations make the spelling ambiguous live, even though they share an identity and therefore one persisted node. My claim that "two entries with the same identity share one node and are therefore not ambiguous, the same call `export_surface` makes" was simply wrong.

Rather than patch the inference a third time, I removed it. `EvidenceSearchRecord` now persists `ambiguous_identities` — the same `ambiguous_type_names` set live consults — and `_type_identity_is_ambiguous` answers both of `_confirmed_type_matches`'s clauses from it. The graph-derived helper is deleted.

That's the pattern worth recording: this is the third round on one sub-problem, and each time the reconstruction was *nearly* faithful, which reads exactly like a correct one until someone finds the case it drops. An observation the replay needs is cheaper to persist than to reconstruct. I took your second option rather than the blunt `identity_coverage COMPLETE` gate, which would have killed positive membership for every finding on any side with a single ambiguous name.

**2 & 3 — the two severity-provenance findings.** Both correct, and both were **pinned by tests I wrote in the immediately preceding round**. The CLI test asserted `gate.severity.addition → project_config` for a run with no `.abicheck.yml` at all; the MCP test asserted all four categories `api_request` when only `severity_abi_breaking` was passed. I'd written both from the implementation rather than from the contract, so they locked in the bug. Corrected, and the fixes use data that was already available: `BuildConfig` carries the four levels separately, and `abi_compare` takes them as separate arguments.

**One thing your findings surfaced indirectly.** Adding `_severity_from_config` pushed `cli_compare_helpers.py` to 2017 lines — past the 2000 hard cap. The receipt plumbing now lives in `cli_compare_receipt.py`. First attempt tripped `import-cycle-growth` (it imported `_param_from_cli` back); rather than extend `IMPORT_CYCLE_ALLOWLIST`, which AGENTS.md rules out for a routine change, the new module is a leaf by construction — "which parameters did the user type" is the caller's question (it holds the Click context) and arrives as data. Allowlist untouched.

**Verification.** Five scratch repros across the last three rounds all agree with live, none strengthen. Three new/corrected tests, each confirmed failing with only the source stashed. `verify.py --profile pr` back to green — `ai-readiness` and `unit-pr` both failed mid-work on the line cap and are passing again; only the known repo-wide `fmt-check` remains. `mypy abicheck/` clean across 309 files.

Separately, I ran a self-review of the whole PR and found two gaps neither bot has raised, which I'll fix next unless you'd rather I stop: **no enum coverage at all** in the new tests (the `_snap` helper doesn't even accept `enums`, yet `enum:` is keyed by the same `_type_identity` and participates in ambiguity), and **zero corpus snapshots with a colliding bare tail** — so the entire ambiguity path that produced these last three rounds of defects has no corpus-level signal, only unit tests.

---

napetrov:
Прошёлся по всему: открытые треды + находки из моего саморевью. `b34d0a6`.

## Новый P2 (`contract_evidence_collect.py:281`) — исправлен

Коллизия ключей **между тирами**, а не внутри одного. Приватная header-only `foo(Secret *)` рядом с объявлением, у которого записанный `mangled` буквально `foo`, ложились на один `decl:foo` — публичный корень наследовал ребро к `Secret`. Мой предыдущий фикс проверял коллизии только среди немангленных. Теперь любой записанный linker identity зарезервирован, и display-name fallback уступает ему независимо от тира. Воспроизведено до фикса, тест падает без него.

## Главное: закрыта причина, по которой все эти раунды находили ревьюеры, а не CI

Все дефекты этой фичи были одного вида — replay заявляет больше, чем live. Корпусной проверки на этом пути **не было вообще**, только руками написанные unit-кейсы. Теперь `measure_contract_shadow.py` переигрывает каждый корпусный кейс через реальный wire-формат и отдаёт `replay_strengthenings` с baseline 0.

Сразу же выяснилось, что **гейт вакуумный**: две намеренные регрессии (снял kind-фильтр, потом отказ по неоднозначности) обе прошли — потому что 0 из 32 корпусных снапшотов имели коллизию по bare tail, так что ни одно replay-решение не могло отличаться ни при какой реализации. Добавил кейс `ambiguous_namespaced_leaf`, после чего те же регрессии гейт валят. «Гейт зелёный» и «гейт способен упасть» зафиксированы как отдельные утверждения.

## Enum-путь: первое покрытие, сразу две находки

В этом наборе тестов ни разу не создавался `EnumType`, хотя enum'ы键 keyed тем же `_type_identity`, алиасятся так же и делят ту же множество неоднозначностей. Дало две настоящие дыры:

1. **Все member-level находки деградировали в `UNKNOWN_UNRESOLVED`** — `_type_candidates` срезает `Mode::B` до владельца `Mode`, а replay нет, поэтому `enum_member_added` на экспортированном enum не резолвился ни во что. Это ослабление (то есть sound), но делало replay бесполезным для целого семейства.
2. Кросс-тирная коллизия выше.

## Старые треды: проверены по коду, не «на слово»

GitHub держит 18 тредов открытыми, но фиксы вошли в более поздние коммиты — Codex-треды не резолвятся автоматически. Прогнал механическую проверку каждого утверждения против текущего HEAD, все 16 проверяемых подтвердились:

`negative-proof before exclusion (P1)` · `distinct finding identities` · `suppression config` · `resolved gate` · `MCP gate refresh` · `CLI contract-mode source` · `identity-coverage gate` · `ambiguity persisted` · `unmangled overload identity` · `owner-class edges` · `overlay roots in replay` · `overlay records cited` · `receipt relevance map required` · `scoped missing labels` · `audit buckets measured` · `explicitly-empty policy value`

Один **сознательно не взят**, как и раньше: замена `"\x1f"`-разделителя в `report_finding_id`. Коллизия требует литерального U+001F внутри слага/символа/значения, чего ни один продюсер здесь не выдаёт, а смена кодировки перехеширует каждый finding id — документированно стабильный, видимый пользователю отпечаток (schema 2.3), на который завязаны waiver'ы и кросс-ран корреляция. Цена определённая, риск недостижим.

## Проверка

Шесть repro-скриптов по всем раундам — все совпадают с live, ни одного strengthening. `verify.py --profile pr` зелёный, кроме известного репо-широкого `fmt-check`. `mypy abicheck/` чист (309 файлов). Корпус 33 кейса, FP=0/FN=0.

Из моего саморевью остаётся один незакрытый пункт, намеренно: домен `exports` даёт 100% unresolved на корпусе, потому что синтетические снапшоты не несут export-таблицы. Это честно отражено в отчёте, но означает, что у `exports` корпусного сигнала по-прежнему нет — для него нужен отдельный кейс с ELF/PE-таблицей, и это своя работа, а не довесок сюда.

---

PR #664, opened 2026-08-03
feat: ADR-049 — --pack manifests configure the run (Phase 5), and the contract-coverage exit is applied (Phase 7, first slice)

napetrov:
## Summary

Two ADR-049 steps, stacked at the maintainer's direction:

1. **Phase 5's `packs` axis** — the one item the plan recorded as still open before Phase 7. `compare --pack` / `scan --against ... --pack` select a manifest that really changes the verdict and the exit code, rather than only the receipt.
2. **Phase 7's first slice** — the contract-coverage ledger's long-computed `0`/`1` contribution now moves the process exit status instead of only being printed.

Plan reference: `docs/contribute/plans/public-contract-default.md` §§ Phase 5 item 2, Phase 7.

## Part 1 — `--pack` configures the run (Phase 5)

Phase 1 built pack manifests, D8's conflict rule, and the wiring that folds a selected pack into one resolved `CompatibilityEvaluationConfig`. Phase 5 wired every front end to that object. What none of it did was **apply** a pack. A `--pack` flag was written and reverted before merge for exactly that reason: a `kind: policy` pack overriding `func_removed` was recorded as active configuration while the verdict and exit code ignored it — and the parity tests written alongside it passed, because they only asserted that the two commands *resolve* packs identically, which a flag that does nothing satisfies.

- **`abicheck/pack_application.py` (new)** — the application layer, deliberately *not* a second resolver. It reads back off the already-resolved configuration only the fields whose `ValueProvenance.source_kind` is `pack_manifest`, so D7 precedence and D8 conflict detection stay in exactly one place and a value the resolver ruled out is unreachable from here. What the packs did supply is folded into the two objects the run is scored from: a `PolicyFile` (`policy.overrides`, `surface.internal_namespaces`) and the resolved compare config (`gate.exit_code_scheme`, `gate.severity.*`).
- **Ordering, which is not incidental:** the configuration is resolved from the *explicitly given* `--policy-file` and only then are the packs folded in. Folding first would present a pack's own override back to the resolver as an explicitly stated policy-file value — outranking the packs it came from, and misreported in the receipt.
- **`UNAPPLIED_PACK_FIELDS`** — a pack may only assign a field this build actually applies. It is the *complement* of what is applied rather than a second hand-kept list, so a newly-routable field is applied or listed, never neither.
- **Emptiness is judged on the revision that configures the run.** The rule lives in `load_selected_packs`, which every path that resolves or validates packs goes through — validating early and resolving later is two reads, and a generated or concurrently edited pack that shrinks between them would otherwise configure the run as exactly the decorative pack the rule rejects.
- **Structural rejections, same "configure or reject" rule:** a `kind: gate` pack on `scan` (whose exit code follows its verdict directly), `--pack` without `--against` on `scan`, and `--pack` on a directory/package `compare`, whose fan-out dispatches before the effective configuration is resolved.

## Part 2 — the contract-coverage exit is applied (Phase 7)

`contract_coverage_exit.py` turns the ledger's `0`/`1` into a real exit code, and only that.

- **Folded with `max`, in exactly one place per command** — `compare` inside `cli._exit_with_severity_or_verdict` rather than at each call site, so a command cannot pick up a compatibility exit and forget the orthogonal one; `cli_scan_baseline` through the same function, since a ledger gating one command and not the other is precisely §6.4's cross-command divergence.
- **`max` is what makes §7's orthogonality executable:** a coverage failure raises a clean `0` to `1` and provably cannot lower a gate's `2`/`4`, which would demote a real ABI break to "warnings only". It never rewrites a finding's compatibility decision or gate contribution either.
- **A run without `--contract-evaluation` is unaffected** — no selected domain means nothing to be short of evidence for. That is every pre-existing invocation.
- **`contract.unresolved` gained its first engine consumer** and therefore left `UNAPPLIED_PACK_FIELDS`. It is the one applied pack field nothing folds into a legacy object: its consumer reads it off the resolved configuration the receipt already persists. `warn` zeroes the floor and changes nothing else (§6.2) — the failures stay listed and unsuppressible, because accepting incomplete assurance is not hiding it.
- **The reporter now emits the applied number**, derived by the same function the exit path uses. A field named "exit contribution" that disagreed with the actual exit status was harmless while nothing applied it and a trap once something did.

### Every consumer that publishes an exit status folds it, and explains it

Applying the axis in the CLI is not enough on its own: a consumer that reports its own exit contract and does not fold this one lets a caller accept a run the tool considered gated. All of these now agree, and each says *which* axis moved the number:

- **`compare` and `scan --against`** — the `max` fold, plus a stderr notice for every renderer that omits the ledger (only `--format json` carries it, and `--stat` carries it in no format).
- **MCP `abi_compare`** — the same fold on `exit_code`, plus a top-level `contract_coverage` block emitted for *every* `output_format`, since a client should not have to branch on the renderer to learn why its gate failed.
- **SARIF** — `invocations[0].exitCode` folds the floor and `exitCodeDescription` names the axis. `executionSuccessful` deliberately stays `true`: per the SARIF spec it reports whether the tool ran to completion, and the spec's own example pairs `exitCode: 1` with `executionSuccessful: true`.
- **The composite Action** — `verdict: COVERAGE_INCOMPLETE`, rather than misreporting the axis as an operational failure (`ERROR`) or a severity-policy one (`SEVERITY_ERROR`). Where `compare` genuinely shares exit 1 between two axes, they are told apart by the report's pre-fold `severity.exit_code` rather than guessed.
- **JUnit and HTML** publish the *gate's* exit code, which this axis never changes, so both are correct unchanged — checked, not assumed.

Both CLIs' `--help` and `--dry-run` exit-code summaries list the axis too, so a CI integration reading a command's own contract does not meet exit 1 as an undocumented failure.

### Phase 7 is not finished, and the plan says so

Still open, recorded in the plan's Phase 7 section rather than implied: the evaluator is still advisory (it runs *after* `verdict` is computed, over the final `kept`, so a `NOT_EVALUATED` finding still scores — reordering that is the large, FP-rate/tier-accuracy-sensitive piece), `--contract-evaluation` is still opt-in, and `aggregate` folding the ledger into its own coverage axis still holds its reserved wiring.

**The migration-window precondition was not met.** Phase 7 opens with "after release notes and a migration window"; the maintainer directed the flip and stated no window was run. That is a deliberate decision, not an oversight — release notes ship with the change instead of ahead of it, and the plan records it explicitly so a later reader does not mistake it for one.

## Checklist

- [x] Docs updated (`exit-codes.md` gains the coverage axis; plan and `AGENTS.md` updated)

### On the tests specifically

Both parts lead **every** behavioural test with an exit code that differs with and without the feature — deliberately, given how the reverted `--pack` passed its own parity tests. A policy pack takes `compare` from 4 to 0 and `scan --against` the same way; two packs disagreeing on one field is 64. For the coverage axis: an unresolvable domain takes an otherwise-clean run from 0 to 1, the rollback domain (`--contract all`) stays 0, a breaking pair stays 4 (the `max` claim), and `compare`/`scan` are asserted to reach the same code. Each new test was checked against a mutant that removes the code it covers.

Where a whole *class* of divergence kept recurring, the test pins the class rather than an instance — `--pack`'s help text is asserted against `applied_pack_fields()` in both directions, and the Action's report parsing is driven with a `PATH` mirroring the real one minus `jq` alone.

### Gates

mypy clean (317 files) · ruff clean · ai-readiness 0 errors · docs-contract 0 errors · full unit suite 22813 passed / 0 failures · golden lane 10 passed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hr271xFmweQLvnCbRYuEF7)_


napetrov:
Pushed 7b780c6 addressing the review. Summary of what changed and what didn't.

**Fixed — three real findings:**

1. **A gate pack could override an explicitly stated `--exit-code-scheme legacy`** (Codex P1). Reproduced: a severity-only gate pack turned a BREAKING comparison's exit 4 into 0. The root cause is worth naming precisely — `pack_application.py`'s own first stated rule is "read the resolver's answer, never re-derive it", and `apply_to_compare_config` was the one place that broke it, re-deriving the scheme as `"severity" if severity_active`. The resolver already had it right: its `auto` default is computed by `_severity_active`, which folds in the gate pack's own levels, while an explicit `--exit-code-scheme` contributes an outranking `EXPLICIT_CLI` candidate. Now pinned by a parametrization over all three directions (explicit legacy / explicit severity / unstated), not just the case that regressed.

2. **`--pack` validated after the `--dry-run` emit** (Codex P2, CodeRabbit). Manifest validity now runs ahead of it, matching the convention this file already states twice. **Pack-vs-pack conflict detection deliberately did not move**: D8 exempts a field another layer already states, and those layers aren't resolved that early — checking there would make a dry run *stricter* than the real run, which is the same divergence inverted.

3. **`pack_paths` had two sources of truth** (CodeRabbit, Major). A caller filling `params` alone would have its packs recorded in the receipt and applied to nothing — precisely the decorative-`--pack` regression this module exists to prevent. One source now; `contract_evaluation` is an explicit keyword.

Also: the unapplied-field rejection test is derived from `UNAPPLIED_PACK_FIELDS` itself with a sync assertion, so a field added later arrives with a test rather than an untested branch. (Note the suggested triples had `assurance.require_evidence` as `kind: policy` — that's a contract-kind route, so the loader would have rejected it for the wrong reason and the test would have passed without exercising the rule.) Plus the AGENTS.md code-span line break, the two test-callback annotations, and the three `--pack` restrictions stated in `docs/use/policies.md`.

**Not changed, with reasons:**

- **`contract.overlays` routable but unapplied** — this is the registry's whole point, not an inconsistency. "Assignable by the resolver" and "applied by the engine" are deliberately different sets; `UNAPPLIED_PACK_FIELDS` is what makes the gap explicit and enforced, and the reference table states both in one row ("Assignable fields" / "Applied today").
- **Fixtures moved to `conftest.py`** — contrary to actual practice here; `test_scan_compare_parity.py` and its neighbours all define module-local fixtures. Moving two single-file fixtures into the shared conftest would pollute it for one test module.
- **The second manifest read** — inherent to validating *before* resolution, now that validation moved ahead of the dry-run emit. The resolver's own "one read per resolution" rule is about not splitting a single resolution's identities across two reads, which this doesn't do; documented on the helper.
- **`codecov/patch` at 21%** — a partial-upload artifact, not a real gap. It fired while the canonical coverage lane was still running, off the integration-only upload (integration tests don't touch `pack_application.py`). It has since self-corrected to 99.15%.

---

napetrov:
Pushed c2ca1d0 for the new Codex P2 (`cli_compare_receipt.py#L311`, "Revalidate the manifest version used for application"). **Confirmed real, and fixed.**

The gap was exactly as described: `validate_pack_manifests` runs before the `--dry-run` emit, then a full snapshot resolution happens before `resolve_cli_config` performs its own read. A manifest replaced in between reached the resolver unvalidated — an unapplied field it gained (`contract.unresolved`) was routed by the resolver, ignored by `pack_application`, and never rejected. The run silently dropped an assignment the "a pack may only assign what this build applies" rule promises to fail on.

The fix runs the same check again inside `resolve_and_apply`, on the read that actually configures the run. Both calls stay, and the call site now says why so neither is later removed as redundant: **the early one is what makes a dry run agree with the real run; the late one is what makes the applied-field rule true of the version applied.**

One note on the verification, since it matters for how much the new test is worth. My first attempt to confirm the test fails against unfixed code silently did nothing — the patch pattern missed a blank line, so the "disabled" run was actually running the fixed code and passed vacuously. I caught that and redid it properly: without the second check the test exits 4 (assignment silently ignored), with it 64. So the claim here comes from a run that really did fail.

Full unit suite after the change: 22744 passed, 0 failures. mypy clean, ruff clean, ai-readiness 0 errors, docs-contract 0 errors.

**Two older Codex threads are still shown unresolved but were fixed in 7b780c6** (Codex doesn't auto-resolve threads the way CodeRabbit does):
- P1 "Preserve an explicitly selected legacy gate" — fixed; `apply_to_compare_config` now reads the resolver's `gate.exit_code_scheme` instead of re-deriving it, pinned by a parametrization over explicit-legacy / explicit-severity / unstated.
- P2 "Resolve packs before the compare dry-run return" — manifest validity moved ahead of the emit. Pack-vs-pack *conflict* detection deliberately stayed behind: D8 exempts a field another layer already states, and those layers aren't resolved that early, so checking there would make a dry run stricter than the real run.

Also for the record: the earlier "Full example matrix" red was not a code failure. GitHub cancelled the superseded run for 7b780c6 when I pushed d313819 — its four artifact-producing jobs ended `cancelled` at 10:35:21 while the aggregation job had started at 10:35:22 and then hit `FileNotFoundError` on artifacts that were never produced. Both prior runs carry `conclusion: cancelled`; the run on the current head is the authoritative one.

---

PR #671, opened 2026-08-04
fix: address CodeFactor Security, Duplication, and the complexity findings above 35

napetrov:
## Summary

Closes the 36 CodeFactor findings in the Security and Duplication classes, plus the 6 `Complex Method` findings scored above 35 — 42 of the 107 reported issues.

## Motivation

CodeFactor reports 107 issues on this repository:

| Class | Count | Severity | This PR |
|---|---|---|---|
| Complexity — *Complex Method* | 70 | Minor | the 6 scored **> 35** |
| Security — hardcoded `/tmp` (bandit B108) | 22 | Major | ✅ |
| Security — `xml.etree` `fromstring` (B314) | 5 | Major | ✅ |
| Security — `yaml.load` (B506) | 2 | Major | ✅ |
| Security — swallowed exception (B110/B112) | 2 | Minor | ✅ |
| Duplication — *Duplicate Code* | 6 | 1 Major, 5 Minor | ✅ |

### Security (30)

Every fix confirmed by running `bandit` over the whole tree before and after: none of B108/B314/B506/B110/B112 is reported anywhere now.

- **B108 hardcoded `/tmp` (22)** — every one is an opaque *fixture* path in a test (fake checkout roots in `test_graph_reconcile`, a fake suppression path, an absolute-path-shape parametrization), never a real temp file. Re-rooted at `/work/`, chosen because it is outside the `include`/`src` root-marker vocabulary `graph_reconcile` anchors moved-file detection on, so the fixtures keep meaning exactly what they meant.
- **B314 `xml.etree.ElementTree.fromstring` (5)** — `tests/test_contract_coverage_ledger.py` now uses `defusedxml`, matching what `tests/test_junit_report.py` already does. `defusedxml` is a runtime dependency, so nothing new is pulled in.
- **B506 `yaml.load` (2)** — both call sites pass a `yaml.SafeLoader` *subclass* (it only adds a duplicate-key check), which bandit cannot name-match. Marked `# nosec B506` with the reason, the idiom this repo already uses for B202/B310.
- **B110/B112 swallowed exceptions (2)** — a failed cleanup in `buildsource.inline._run_cleanups` and an unclassifiable `compare` operand in `cli_options` now log at `debug` instead of a bare `pass`/`continue`. Both stay best-effort; both paths gained a test.

### Duplication (6)

Each was two copies of one procedure; each is now stated once, with the reason it must stay that way recorded where a future reader will hit it:

- `service.run_dump`'s PE and Mach-O branches shared a byte-identical post-dump tail → `_finish_native_snapshot`. The ELF branch deliberately stays separate: it also attaches SYCL metadata and honors `dwarf_only`.
- `service_dump_cache` handed `run_dump` the same 22 arguments in two places → stated once, so a newly threaded parameter cannot reach one path and silently skip the other.
- `reporter_markdown`'s leaf-change and root-cause views shared their opening block → `_view_preamble`. Full mode stays separate (four extra count rows, and it applies `--show-only` with no note line).
- The L5 call-graph and type-graph extractors ran a byte-identical bounded `clang -ast-dump=json`, kept in step *by hand* through comments pointing at each other across several review rounds of deadline fixes → new leaf module `buildsource/clang_ast_run.py`.
- `check_ai_readiness.py` and `check_docs_contract.py` each carried an identical `Findings` class (the second even said so in its docstring) → `scripts/findings_report.py`, subclassed by each gate to pin its summary label.
- `gen_g20_fixtures.py` and `gen_reachability_examples.py` each carried the same write-or-`--check` driver → `scripts/fixture_sync.py`, with new unit tests for the drift half the generators' own tests structurally cannot reach.

### Complexity > 35 (6)

Decomposed into named steps. No behaviour change: same inputs, same outputs, same evaluation order.

| Function | Score | What it became |
|---|---|---|
| `mcp_server.abi_compare` | 55 | the two ADR-043 scoping branches (~200 of 868 lines) → `_apply_used_by_scope` / `_apply_required_symbols_scope` |
| `cli_compare_helpers.run_compare` | 50 | `_reject_incoherent_compare_flags` + `_preflight_manifests_and_audit` |
| `graph_reconcile.reconcile_added_removed` | 44 | three tiers → methods on a `_Reconciler` over a `_KindPass`; the repeated pair construction → one `_record_pair` |
| `dumper_ast_config_cpp20._strip_inactive_if_zero_blocks` | 42 | moved to a new `dumper_ast_config_cpp20_chains` module; state machine → `_frame_for_chain_open` + `_line_for_open_arm` |
| `cli_compare_fold._fold_scoped_compat_into_text` | 39 | a `_ScopedFold` with `into_json` / `into_text` / the `--stat` variant as methods |
| `type_reachability.directly_referenced_stdlib_types` | 38 | partition / seed / walk over a `_StdlibReferenceScan`; the five-check public-root guard stated once |

Two of these paid off beyond the metric. `abi_compare`'s tool function keeps its signature and docstring untouched — the MCP schema and `docs/reference/mcp-tools-reference.md` are generated from them. And the `dumper_ast_config_cpp20` move takes ~590 lines out of a file that sat **23 lines** under the AI-readiness 2000-line hard cap, with the dependency running one way (parent imports the chains module, never the reverse) so no new import cycle forms; the name is re-exported from the parent, which is where its own test imports it from.

## Not in this PR

The 64 remaining `Complex Method` findings, all scored ≤ 35. Many are Click command definitions and MCP tool signatures whose branchiness is inherent to the dispatch they perform; the rest are worth splitting as scoped work with its own verification per module.

## Verification

- `pytest tests/ -m "not integration and not libabigail and not abicc and not slow and not golden"` — 23156 passed, 32 skipped, 4 xfailed
- `mypy abicheck/` — clean, 330 source files
- `ruff check abicheck/ tests/ scripts/` — clean
- `python scripts/check_ai_readiness.py` — 0 errors
- `python scripts/check_fp_rate.py` — 0 FP / 0 FN, within baseline
- `python scripts/check_tier_accuracy.py` — OK (top-tier correct, under-call monotonic)
- `python scripts/check_docs_contract.py` — 0 errors
- both fixture generators `--check` — no drift
- `bandit -r` over the tree — no B108/B314/B506/B110/B112 remaining

## Checklist

- [x] Tests added / updated for new behaviour — no behaviour change; new tests cover `fixture_sync`'s drift detection and the two logging paths
- [x] Changelog fragment added (`scriv create`) — three fragments (`Fixed`, `Security`, `Changed`)
- [x] Docs updated if user-visible behaviour changed — none is; `scripts/CLAUDE.md`'s inventory gained the two new script modules


napetrov:
Addressed in 476f11a.

**Applied**

- **`fixture_sync.py` — missing file with empty expected content.** Real, though pre-existing: the `current = ... if path.is_file() else ""` line is carried over verbatim from both generators. Absence is now checked as drift in its own right, before the content comparison. Added `tests/test_fixture_sync.py` covering the driver's drift half — which the generators' own tests structurally cannot reach, since the committed fixtures are by construction in sync — including the empty-fixture/absent-file case.
- **`_slow_loads` annotations** in `tests/test_call_graph.py` and `tests/test_type_graph.py`.
- Also covered the two logging paths codecov flagged as uncovered: a failing cleanup must not skip the cleanups after it (and must log), and an unclassifiable `compare` operand contributes no kind (and must log).

**Not applied**

- **Renaming `_finish_native_snapshot`'s `compile` parameter (A002).** `A` is not in this repo's ruff select list — `ruff.toml` selects `E`/`F`/`W`/`I`/`UP`, and `ruff check` is clean on the branch. `compile: CompileContext | None` is the established spelling in `api_types.py`, `cli_resolve.py`, and five other functions in `service.py` itself; renaming only the new helper would make it the odd one out among the twelve `compile=compile` call sites in the same file. Worth doing as a repo-wide rename if the parameter name is to change, not on one extracted helper.
- **Removing the import-time `sys.path` insert** in `check_docs_contract.py` / `gen_g20_fixtures.py` / `gen_reachability_examples.py`. All three already inserted `ROOT` at import for their `abicheck` import, and `check_ai_readiness.py` has done exactly this for its own sibling gate module (`adr_status_sync`) since that split — the added line follows the existing pattern rather than introducing one. It is also load-bearing: `tests/test_docs_contract.py` loads the gate via `spec_from_file_location`, which does not put the script's directory on `sys.path`, so the sibling import fails without it.

---

napetrov:
@coderabbitai the Title check verdict ("incorrectly implies that complexity findings above 35 were addressed") is based on a stale review — the commit status CodeRabbit posted for `d0c9b40` is *Review rate limited*, so that commit was never read. Its walkthrough still lists only the security/duplication files and does not mention `type_reachability.py`, `graph_reconcile.py`, `mcp_server.py`, `cli_compare_fold.py`, `cli_compare_helpers.py`, or the new `dumper_ast_config_cpp20_chains.py` at all.

`d0c9b40` splits all six functions CodeFactor scores above 35 — `abi_compare` (55), `run_compare` (50), `reconcile_added_removed` (44), `_strip_inactive_if_zero_blocks` (42), `_fold_scoped_compat_into_text` (39), `directly_referenced_stdlib_types` (38) — across 8 files, +1926/−1449. No re-review needed on my account; flagging it only so the pre-merge table isn't read as a factual claim about the diff.

---

PR #673, opened 2026-08-04
feat: G33 Phase 5 — a typed DumpRequest, and the MCP parity it unlocks

napetrov:
## Summary

Closes G33's last open phase: `dump` gets the typed request `compare` has had since ADR-037 D2, and the MCP `abi_dump`/`abi_scan` tools reach the evidence and config surface their CLI counterparts already accept.

## Motivation

The phase's own framing was that `abi_dump`/`abi_scan` should copy the depth/sources/build-info parity `abi_compare` gained in Phase 4. **There was none to copy** — Phase 4 moved that tool onto `CompareRequest`, which *carries* those fields (Phase 2), without exposing them as tool arguments. So `abi_dump` had to gain the concepts outright.

That is why this is not only MCP surface. `resolve_input` has always been the one way to turn a path into a snapshot, but everything a real `dump` does *around* it — inferring a collect mode, embedding inline L3–L5 evidence, walking dependencies, enforcing that an explicit `--depth` was reached — lived only in `cli.py`'s `dump_cmd`. Adding the arguments without a typed request would have meant a second copy of those four steps inside `mcp_server.py`: precisely the invariant the plan's §2 exists to protect. `abi_dump` sat at a five-argument subset of `abicheck dump` for several releases with nothing noticing.

## Changes

**Typed `dump` request and one resolution**

- `api_types.DumpRequest` — one `InputSpec` plus the how-it-runs fields `CompareRequest` also keeps at request level. Deliberately carries nothing about classification (policy, suppression, scope, severity, contract): a dump produces evidence and renders no verdict. Both requests now validate through one set of module-level helpers, so `dump` and `compare` reject an identical mistake with identical text — ADR-037 D9's front-end parity, extended across the two commands.
- `service_dump_pipeline.run_dump_request` — the four steps above, over the same per-input primitives `compare` resolves through. Excludes the CLI's provenance/presentation layer (`--dry-run` rendering, git/build-id stamping, `fold_dump_provenance_into_json`), which describes how one front end *reports* a dump.
- `service_input_resolution` — those primitives. All of it was `service_compare_pipeline`'s private helpers (`_resolve_side`, `_embed_side_build_source`, `_enforce_requested_depth`), lifted out of the *pair* and re-expressed for one input, so a change to how an input resolves lands on both commands at once. The pair-shaped decisions stayed behind on purpose: the pair-wide C++20 dialect override exists because two sides must agree on a standard, and the concurrency rule is about two extractions at once — neither means anything for a lone dump.

**MCP surface**

- `abi_dump` gains `depth`, `sources`, `build_info`, `dump_manifest`, `public_header_dirs`, `include_dependencies`, `dwarf_only`, `debug_format` and the `ast_frontend`/`gcc_path`/`gcc_prefix`/`gcc_options`/`sysroot`/`nostdinc`/`frontend_context` compile-context family, and routes through `run_dump_request`. Its MCP-local guards are unchanged, carried as request fields (path containment, file size, and the linker-script pin that keeps the size check authoritative).
- `abi_scan` gains `build_info`, the same compile-context family, and the `--against` config surface ADR-049 Phase 5 §6.4 already required it to have (`policy`/`policy_file`/`suppression_file`/`contract_evaluation`/`contract_mode`).
- `abi_compare` gains `contract_mode` — ADR-049 Phase 6's `--contract`, a field `CompareRequest` already had but no MCP caller could set. Adding it to `abi_scan` alone would have left the two tools disagreeing about the same ADR.
- `--contract`'s two usage rules have one implementation shared by both tools, since `ScanRequest` has no `validate()` of its own to state them in.

**Housekeeping the change forced**

- `IMPORT_CYCLE_ALLOWLIST` gains both new modules, on the terms `service_compare_pipeline` was signed off under in Phase 2 (CLAUDE.md "M1-3"): each is a *split* of an existing member, and every edge they have is one the code already had one module over, moved rather than added. Both import only leaf modules at module load — the package still imports cleanly.
- `mcp_server.py` crossed the 2000-line hard cap on the new parameters, so its argument-translation layer moved to `mcp_server_inputs.py` verbatim, re-exported for existing callers and tests.
- Generated docs regenerated (`gen_mcp_reference.py`, `gen_python_api_reference.py`); `AGENTS.md`'s module map, the G33 plan, ADR-055's "still open" list, and the plans index updated. AGENTS.md's "Depth contract, CLI vs. API/MCP" known-gap entry — which had been *closed as stale* on the grounds that no service/MCP surface promised a depth-qualified snapshot — is corrected rather than deleted: `DumpRequest.depth` is now exactly such a promise, and the gap closed by giving the typed surface the capability and the gate together.

**Deliberately not done**, and named rather than implied: the native `dump` CLI still resolves its own inputs. That is `dump`'s analogue of Phase 2's structural half; it needs a real seam found in `dump_cmd` first, the way `resolve`/`classify` was found for `compare`. Guessing at one inside a phase scoped to MCP parity is how a second resolution path gets created rather than removed.

## Checklist

- [x] Tests added / updated for new behaviour — `tests/test_typed_dump_request.py` (53 tests: `DumpRequest` validation parity with `CompareRequest`, `run_dump_request`'s depth floor / collect-mode inference / compile-context merge / dependency opt-in, and the phase's gate as a signature check over the three MCP tools). Three existing test files updated where they patched names `abi_dump` no longer calls — the same relocation ADR-055 D4 made for `abi_compare`.
- [x] `pre-commit run --all-files` passes locally — verified via `scripts/verify.py --profile pr`: `lint`, `typecheck`, `ai-readiness`, `unit-pr` (23233 passed, coverage 96.71% against the 95% floor), `fp-rate`, `tier-accuracy`, `docs-contract`, `docs-build`, `repo-facts`, `schema-sync`, `fair-metadata` all pass. Two steps did **not** pass cleanly and neither is caused by this change: `distribution-build` skipped (no `build`/`twine` in this environment), and `fmt-check` fails repo-wide — 523 files already fail `ruff format --check` on a clean `main`. The five files this PR adds are format-clean.
- [x] No new `mypy` or `ruff` errors introduced — `mypy abicheck/` clean at 331 files, `ruff check` clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01T9VEP9bknij5BFpCWzMw28)_

## Summary by CodeRabbit

* **New Features**
  * Added typed dump requests with validation, evidence collection, dependency handling, compiler-context options, and depth enforcement.
  * Expanded `abi_dump`, `abi_compare`, and `abi_scan` with contract, build, frontend, debug, policy, suppression, and public-header options.
  * Added support for source inputs, manifests, compiler settings, and dependency filtering.
  * Contract mode is now captured in comparison results.
* **Bug Fixes**
  * Improved validation for invalid paths, unsupported combinations, Android inputs, size limits, and debug formats.
* **Documentation**
  * Updated MCP and Python API references with the new options and dump workflow.


napetrov:
Pushed `84c54c7` — both Codex findings were real, verified against the code before fixing, and each has a regression test.

**P2 — `android` in the header compile context.** Confirmed: `abi_dump(ast_frontend="android", build_info=…)` on a real compiled ELF died with `Unknown AST frontend 'android'` before any build/source evidence was embedded. The mechanism is exactly as described — `_run_dump_uncached` gives an explicit `compile.frontend` precedence over the `header_backend` argument, and `_resolve_header_backend` rejects anything outside castxml/clang/hybrid/auto.

One correction to the scoping, which changed where the fix went: this is **not** confined to the new MCP argument. A typed `CompareRequest` whose `InputSpec.compile.frontend` is `"android"` failed identically — verified with the same fixture — so the new `ast_frontend` parameter made a pre-existing defect *reachable* rather than introducing it. Fixed in the shared resolution (`service_compare_evidence._header_ast_frontend_only`), since the invariant is "what may reach `resolve_input` must be a header-AST frontend" and it holds for both commands. Nothing could depend on the old behaviour: it raised. The request-level `frontend` still carries `android` for L4 source-ABI replay, the only layer that can act on it.

**P1 — nonexistent inferred evidence paths.** Confirmed: `abi_dump(sources=<typo>)` returned `status: "ok"` with no evidence. Fixed with a shared `_existing_path` helper on every argument that infers collection. One deliberate widening: `abi_scan`'s `sources`/`compile_db` predate this PR and carried the identical hole, so they're fixed alongside the new `build_info` rather than left as a knowingly-broken twin — the `scan` CLI already declares all of them `click.Path(exists=True)`, so this only aligns the MCP surface with it.

Both verified end to end (`ok` for the android case, `<argument> not found: <path>` for each missing path), plus `verify.py --profile pr` → lint / typecheck / ai-readiness / unit-pr / docs-contract all pass.

---

On the two CI signals that fired against the previous head, `9697403`:

- **`codecov/patch` at 33.01%** — this was a partial-report artifact, and Codecov has since corrected itself to **97.46%** (target 90%) on the same commit. The status posted at 18:22:10 while the only lane uploading the `unit` flag finished at 18:19:18; the per-file numbers it quoted (`mcp_server.py` 6%, `api_types.py` 50%) match an integration-flag-only report, and `api_types` cannot be at 50% in any real run of the full suite. Local measurement over the same test selection puts every changed file at 95–100%.
- **`integration-tests (windows-latest)`** — 5 `cmake build failed for caseNNN` failures with **empty stderr**, in `test_example_autodiscovery.py`'s MSVC builds of example C++ fixtures. Nothing in this diff is on that path (the integration lane imports none of the modules it touches), and a genuine MSVC compile failure prints `error C####`. The test already documents this failure mode in the code — it serializes `cmake --build` on the shared build tree under xdist precisely because concurrent writes to shared CMake state "can be flaky on Windows/NTFS". Not claiming it's pre-existing on the base — that job was green on `7f53727f` — so this push re-runs the matrix and I'll treat a repeat as a real signal rather than flake.

---

napetrov:
Pushed `c36051c` — both second-round findings confirmed and fixed, each with a regression test.

**Size-check file `build_info`.** Confirmed by inspection: `bi_path` was the only file-shaped input on either tool with no `_check_file_size` call. Rather than add a fourth call site I folded the guard into `_existing_path`, the one helper every evidence path already goes through (file-only — a size limit means nothing for a directory), which also let the now-redundant explicit checks for `compile_db`/`against`/`dump_manifest` go. Verified: an 8 KB `compile_commands.json` under a 4 KB limit is now rejected on both tools, and a directory `build_info` is unaffected.

**Validate `abi_scan`'s compile-context knobs.** Confirmed. Fixed in `_compile_context_from_args` — the one place both tools assemble that context — reusing `api_types`' own message helpers rather than restating them, so the text cannot drift from `DumpRequest.validate()`'s. Verified both tools now answer `unsupported AST frontend 'castxmll': choose from android, auto, castxml, clang, hybrid` and `unsupported frontend context 'gpu': choose from device, host`, and that `"DEVICE"` normalizes to `device`.

One thing your comment didn't name that fell out of it: `android` is a legal `--ast-frontend` value, and a `ScanRequest` has **no** request-level frontend field to carry it for L4 — so leaving it in the compile context would have hit `_resolve_header_backend`'s raise inside the scan worker, the same defect the previous round fixed for `dump`/`compare`. The shared helper now downgrades a non-header-AST frontend to `auto` for the *compile context* specifically; `DumpRequest.frontend` still receives the caller's value where the source-ABI layer can act on it.

**A test-infrastructure note worth recording**, since it cost real time and will bite the next person. The new size-guard test passed alone and failed in the full suite: under some selection orders the suite ends up with **two distinct `abicheck.mcp_shared` module objects**, so patching the one the test file imports leaves the one `mcp_server_inputs` holds untouched and the guard reads the original 500 MB (`_check_file_size.__globals__["MCP_MAX_FILE_SIZE"]` was still `524288000` while the test's `mcp_shared.MCP_MAX_FILE_SIZE` was `4096`). The fixture now patches through the module under test, which is identity-proof. The source-side call was switched to module-qualified access (`mcp_shared._check_file_size`) as well — that is `mcp_shared`'s own documented rule for its runtime-mutable state, which `main()` overrides from `--max-file-size`. The duplicate-module condition itself is pre-existing test-infra behaviour and I've left it alone.

`verify.py --profile pr` → lint / typecheck / ai-readiness / unit-pr / docs-contract all pass (23,25x tests, coverage 96.71% against the 95% floor).

For the record on the earlier signals: `codecov/patch` settled at **97.60%** (target 90%) and passed on `84c54c7`, confirming the 33% first report was the partial-upload artifact.

---

napetrov:
Pushed `cef50a0` — both Codex P2s fixed, three of four CodeRabbit items taken, one skipped as incorrect. Each fix has a regression test; `verify.py --profile pr` (lint / typecheck / ai-readiness / unit-pr / docs-contract / docs-build) passes.

**Codex: seed the build's L2 include dirs.** Confirmed — `buildsource.l2_seed.seed_l2_includes` exists for exactly this and `cli_dump_helpers` calls it twice, while the typed path passed only `side.includes`. Fixed in `resolve_side_snapshot`, which means **`compare`'s typed path is fixed with it**: `InputSpec.sources`/`build_info` landed in ADR-055 D1 and the same gap has existed there since, so this is not confined to the surface this PR adds.

One deliberate divergence from the CLI, called out in the code: `allow_inferred_build_query=False`. Passive discovery of an existing compile database still applies, but a Tier-2 API call must never *execute* a build system (cmake/make/bazel) as a side effect of resolving an input — the CLI may only because the user typed a command that says so. Seeded temp dirs drain in a `finally`, after the parse and any embed consumed them, since an inferred build dir can hold the generated headers they point at.

**Codex: propagate `contract_mode` into the MCP receipt.** Confirmed, and it was my own doc-vs-code drift: adding the argument made `mcp_compare_receipt`'s explicit claim — "takes no contract-mode parameter … those resolve as built-in defaults" — false. Since ADR-049 Phase 7 the selected domain decides which findings policy scores, so a receipt naming the default misreports the domain behind the verdict and the coverage gate. Threaded through `resolve_tool_config`/`record_resolved_config` into the `ExplicitCompatibilityInputs.contract_mode` field that already existed, and corrected the claim in both the module docstring and AGENTS.md.

**CodeRabbit.** Two were docs my previous commit made stale — the `ast_frontend` description called `android` an L2 header-AST frontend when it is source-ABI-replay only (fixed at the generator source, regenerated), and the `--max-file-size` row still excluded every `build_info` after the file form gained the guard. The nitpick is taken as well: `FRONTEND_CONTEXTS` plus one message helper, so the vocabulary and its wording live in one place.

**One CodeRabbit item skipped as incorrect** — the `SnapshotError` docstring entry in `service_compare_pipeline.py`. The comment says "this `raise` will raise `NameError`", but there is no `raise SnapshotError` in that file: line 275 is a `Raises:` docstring entry, and the exception genuinely propagates from `service_input_resolution.embed_side_build_source` (which raises it) and from `service.resolve_input`. Documenting an exception a callee raises is correct and needs no import.

On the docstring-coverage pre-merge warning (43.59%): the three modules this PR adds are at **100%** (`service_dump_pipeline` 2/2, `service_input_resolution` 5/5, `mcp_server_inputs` 8/8). The figure is dominated by test methods (25/85 in the new test file), which this repo does not docstring individually — `tests/CLAUDE.md` sets the convention, and the repo's own `test-assertion-density` gate passes. Not adding ~60 docstrings to move a bot metric.

---

napetrov:
Pushed `428b003` — both Codex P2s on `cef50a0` fixed, each with a regression test. `verify.py --profile pr` (lint / typecheck / ai-readiness / unit-pr / docs-contract / docs-build) passes.

**Forward the sysroot to dependency resolution.** Confirmed: `cli_dump_helpers` forwards `--sysroot` to `populate_dependency_info` and `cli_resolve` does the same, while the typed helper passed `None`. So a `follow_dependencies=True` run under a cross/sysrooted extraction searched the host defaults and reported the target's dependencies unresolved. It now takes the input's own `compile.sysroot` — the only sysroot a typed request can carry — which fixes `compare`'s typed path with it, since both go through this helper.

**Reject `android` on `abi_scan`.** Confirmed, and it was a consequence of my previous commit: downgrading a non-header-AST frontend to `auto` is correct for `abi_dump`, whose `DumpRequest.frontend` still carries the value into source-ABI replay, but `ScanRequest` has no request-level frontend field, so the value vanished and the scan would run an ordinary auto/castxml pass for a documented argument.

Of the two remedies you offered I took **reject**, not "add a request-level frontend": threading a new `ScanRequest` field into the engine's L4 replay is a feature, and faking it with a downgrade is what created this bug in the first place. The error names `abi_dump` as the tool that does carry the value. A regression test pins that `abi_dump` still accepts `android`, so the rejection cannot spread to the tool where it is meaningful.

Codecov settled at **98.04%** on `cef50a0` (target 90%), and CodeRabbit has marked both of its docs findings resolved.

---

napetrov:
Pushed `e8190b2` — the audit-only finding is fixed. `verify.py --profile pr` (lint / typecheck / ai-readiness / unit-pr / docs-contract) passes.

Confirmed by reading the path: `abi_scan` without `against` sets `mode="audit"`/`baseline=None`, `run_scan` calls `_reject_comparison_only_fields`, and that `ValidationError` is raised **inside the spawned worker** — `run_scan_subprocess` rethrows it as `RuntimeError`, which this tool then reports as a sanitized unexpected error, after paying for a process spawn.

Checked before spawning now, reusing the engine's own `_reject_comparison_only_fields` rather than restating its field list — a field that becomes comparison-only later must not need a second edit in the MCP layer. The message keeps the engine's wording (so the two surfaces agree) and appends the argument an MCP caller can act on, since the shared text is `ScanRequest`-shaped:

> ScanRequest field(s) policy only take effect with a baseline comparison … **On this tool, set `against` to compare against a baseline, or drop these arguments.**

Worth noting: three tests added earlier in this PR were themselves passing these arguments without `against`, and the new guard rejected them. They now pass `against`, which is what those parameters' own docs ("With `against`: …") describe. The guard catching my own tests is the evidence it works.

That's nine review findings across five rounds, all resolved. For anyone reading back, the two I did **not** act on, each with a reason on the thread: CodeRabbit's `SnapshotError` report (it read a `Raises:` docstring entry as a `raise` statement — the exception genuinely propagates from a callee), and the docstring-coverage warning (the three modules this PR adds are at 100%; the 43.59% is dominated by test methods this repo doesn't docstring, and its own `test-assertion-density` gate passes).

---

napetrov:
Pushed `e40f318`. Good catch, and it was against my own fix — reproduced before changing anything: `run_dump_request` with `CompileContext(frontend="clnag")` returned a snapshot, and `validation_errors()` came back empty. The `android` → `auto` downgrade was rewriting *every* non-header value, so a typo went from raising `Unknown AST frontend` to silently producing a successful default-backend run. That is worse than the bug it fixed: the caller gets a plausible answer from a frontend they never asked for.

Took **both** remedies you offered, because they defend different callers:

- **Validate the side value.** `validate()` now checks the per-input `compile.frontend` on both request types. The request-level `frontend` was already validated; this field never was, which is why a typo reached the extraction layer at all. Same message shape as every other bad value.
- **Narrow the strip.** The downgrade now applies only to a frontend that is *known* but header-less (`SUPPORTED_FRONTENDS - HEADER_AST_FRONTENDS`, i.e. `android`). An unknown value falls through untouched so `_resolve_header_backend` still raises for it — the second line of defence if some future caller skips `validate()`.

A regression test pins all three cases directly on the helper (`android` → `auto`, `clnag` → unchanged, `clang` → unchanged), so narrowing it back cannot silently resurrect this.

`verify.py --profile pr` (lint / typecheck / ai-readiness / unit-pr) passes; Codecov settled at 98.13% on `e8190b2`.

---

napetrov:
Took all four review items in `c00a67d`. One turned out to be worth more than its "trivial" rating; the other three are as described.

**Extend the audit-only parametrization to `policy_file`/`suppression_file`** — done, and this one is not just closing a gap in the field list's coverage. The guard reuses the engine's `_reject_comparison_only_fields`, so the list itself is shared rather than restated — but `policy_file` and `suppression_file` reach it by a materially different path than `policy`/`contract_evaluation` do: `abi_scan` resolves, size-checks and **parses** both files (`load_suppression_and_policy`, `mcp_server.py:1777`) before the guard runs at line 1806. Verified both still reject before any worker spawns:

```
suppression_file -> ScanRequest field(s) suppression only take effect with a baseline comparison ...
policy_file      -> ScanRequest field(s) policy_file only take effect with a baseline comparison ...
```

Worth noting for anyone reading the table: the engine names the `ScanRequest` field, so the `suppression_file` argument is rejected under the name `suppression`. The test asserts that spelling, with a comment saying why, rather than the argument name.

**Consolidate the two sysroot tests** — done. I kept the substantive half of the concern rather than only the deduplication: the replacement now reads the sysroot out of `**kwargs` as well as the positional slot, so if `populate_side_dependency_info` ever switches to keyword arguments the test keeps asserting on the real value instead of raising `TypeError`. Left it as a class-scoped fixture rather than moving it to `conftest.py`, since it's used by exactly these two tests — `tests/CLAUDE.md`'s guidance is about fixtures shared beyond one module.

**Rename the local `compile`** — done. For the record on the linter attribution: `A001` is not in this repo's `ruff` select list, so `ruff check abicheck/` was and is clean; the finding only reproduces under the explicit `--select A001` the analysis script passed. Shadowing a builtin is a fair readability nit on its own merits, so I took it regardless — as `ctx`, which reads better than `compile_context` at these two use sites.

**Changelog `public-header` split across lines** — real, and rendered exactly as described. Fixed and the paragraph reflowed.

On the pre-merge **docstring coverage 39.72%** warning: unchanged from my earlier reply, and I re-measured it against this head. The modules this PR adds are fully documented — `service_dump_pipeline.py` 2/2, `service_input_resolution.py` 6/6, `mcp_server_inputs.py` 9/9, all 100%. The repo-wide figure is dominated by `tests/test_typed_dump_request.py`'s 115 defs, which are individual test methods under documented test classes; adding a docstring to each would satisfy the metric without telling a reader anything the test name doesn't.

`scripts/verify.py --profile pr --only lint,typecheck,ai-readiness,unit-pr`: 4 passed, 0 failed — 23279 passed, coverage 96.70% against the 95% floor.

---

napetrov:
Fixed the L4 replay finding in `05afec0`, and diagnosed the `Full example matrix` failure — it isn't this diff.

### Pass the selected clang into source replay — real, fixed

Confirmed and precisely as described. `embed_build_source` defaults `clang_bin` to a bare `"clang"`, and both other callers already override it — `cli.py:341` passes `clang_bin=clang_bin`, and `scan_engine.py:324` passes `resolve_source_frontend_clang_bin(...)` under a comment describing this exact failure ("silently replayed L4 through a plain `clang` that may not even understand the real build's flags"). The typed path omitted the argument entirely, so it was the one caller still carrying the bug the other two had already fixed.

Reproduced before and after:

```
icpx via gcc_path      -> clang_bin='/opt/oneapi/icpx'    (was: absent -> "clang")
no compile context     -> clang_bin='clang'               (unchanged)
```

Mirrors `scan_engine`'s call including `exclude_cl_style=False`, which `resolve_source_frontend_clang_bin`'s own docstring calls for: L4 re-drives a CL compile unit with `--driver-mode=cl` itself, so unlike the S2 pre-scan it must *not* fall back off a CL-mode driver — otherwise an Intel DPC++ build replays through a plain clang that can't parse it. Regression tests cover all three cases, the CL-style one included.

### `Full example matrix` — caused by my own push timing, not the diff

Worth stating precisely rather than waving at flake. This workflow sets `concurrency: cancel-in-progress: true`, keyed on the PR number. Pushing `c00a67d` cancelled the in-flight `e40f318` run, and three jobs came back `cancelled`. In `Artifact modes`, the failure sequence is visible in the step list: `Upload artifact-mode results` is **skipped** despite carrying `if: always()` — cancellation skips even `always()` steps. So `examples-validation-artifact-modes` was never uploaded.

`full-example-matrix` then ran anyway, because it is `if: always()` with `needs: [validate-examples, artifact-modes, clang-header-fallback]` — and `always()` is true for cancelled dependencies too. It downloaded an artifact that doesn't exist and died on the first read:

```
FileNotFoundError: 'downloaded/examples-validation-artifact-modes/validate_examples-release-headers.json'
```

Nothing in this PR touches `examples/`, `tests/validate_examples.py`, `scripts/check_examples_validation_status_sync.py`, or that workflow. Note also that a genuine *content* failure could not produce this error: the step redirects with `>`, so the file would exist and fail as a `JSONDecodeError`, not a missing path.

There is a real (pre-existing) robustness gap underneath — a downstream `if: always()` job that can't tolerate cancelled upstream jobs will fail this way on any superseding push. `if: ${{ !cancelled() }}` on `full-example-matrix` would fix it. I've left that alone deliberately: it's a CI-workflow change unrelated to G33 Phase 5, and it belongs in its own PR rather than riding along here. The fresh run on the current head has this job running normally.

### Codecov

The report flip-flopped between 98.16%/7 lines and 97.89%/8 lines across the two heads — `service_input_resolution.py` appeared only because my rename made those lines part of the diff. It's partial-report timing, same as earlier in this PR; the remaining items are branch partials, no gate failure.

`scripts/verify.py --profile pr --only lint,typecheck,ai-readiness,unit-pr`: 4 passed, 0 failed — 23282 passed, coverage 96.71% against the 95% floor.

---

napetrov:
Real, fixed in `72dc603` — with one field added to the list and one correction to the framing.

**The gap is exactly as described.** `abi_dump` size-checks every `headers=` entry (`mcp_server.py:283`), while a `dump_manifest`'s named headers went unchecked. And the asymmetry is exact rather than incidental: `dump_manifest.py`'s own documentation calls `roots` *"the manifest-mode equivalent of `--header`/`-H`"* and a TU's `forced_includes` *"the manifest equivalent of a single named `--header`"*. So the same file was accepted or rejected purely on which spelling the caller used. Reproduced against a 1 MB limit and a 2.9 MB header, before and after:

```
via headers=      : header is 2.9 MB, exceeds limit of 1 MB          (before and after)
roots             : dump_manifest root is 2.9 MB, exceeds limit of 1 MB
forced_includes   : dump_manifest forced include (tu) is 2.9 MB, exceeds limit of 1 MB
public_header     : dump_manifest public header is 2.9 MB, exceeds limit of 1 MB
```

**A third field, not named in the review:** `public_header_paths` is file-valued too (the ADR-015 provenance input), and had the identical gap. Checking only `roots`/`forced_includes` would have left it open. The two *directory*-valued fields — `public_header_dirs` and a TU's `includes` — are deliberately excluded, being `-I` search roots with no content of their own to bound; there's a test pinning that they aren't rejected.

**One correction to the framing, since it matters for how alarming this is.** I checked whether manifest paths also escape a containment boundary, which would have made this more than a size issue. They don't — there is no such boundary to escape: `_safe_read_path` states outright that it *"does NOT restrict to a specific directory (read paths are user-specified)"*. So this is a resource-limit consistency gap, not a sandbox escape.

**A finding in my own test, worth recording.** The test asserting the two paths agree passed in isolation and failed in the full suite. `mcp_server` imports `_check_file_size` as a *function*, and that function reads `MCP_MAX_FILE_SIZE` from its own module globals — so under a selection order leaving two distinct `abicheck.mcp_shared` objects (the hazard `TestEvidenceFileSizeGuard`'s fixture already documents), patching the module object left the `headers=` side at the real 500 MB limit and the comparison passed *vacuously*. The fixture now patches the globals dict each call path actually reads. Had I only run the new test file, I'd have shipped a test that asserted nothing on one of its two sides.

`scripts/verify.py --profile pr --only lint,typecheck,ai-readiness,unit-pr`: 4 passed, 0 failed — 23290 passed, coverage 96.71% against the 95% floor.

---

napetrov:
Closed the last standing Codecov item in `d36cd69`, and hit a flaky test worth flagging for the repo rather than for this PR.

**`service_dump_pipeline.py`'s 1 missing line** was the only Codecov entry constant across every single report on this PR, through all the flip-flopping. It turned out to be the `android` + raw-source-tree rejection — another behavior the module's own docstring asserts with nothing exercising it. `DumpRequest.frontend` accepts `android` (unlike the per-input `compile.frontend`, which is header-AST only), but a raw source tree under it has no extractor, so it's a usage error rather than a silently weaker snapshot. Now tested, along with the escape the error message itself names (`has_sources=True` with no inline sources) — if that stopped working the message would be actively misleading. The file is at **100%** statement and branch coverage.

That leaves all four files this PR adds at 100%: `service_dump_pipeline.py`, `service_input_resolution.py`, `mcp_server_inputs.py`, `mcp_server.py`'s new paths.

**A pre-existing flaky test, not from this PR.** The gate run before this one failed on `test_typedef_matching_stays_fast_with_many_candidates_and_typedefs` (`tests/test_dumper_scoping_dependency_retention.py`) — a file this PR never touches. It's a wall-clock guard, `assert elapsed < 5.0`. Checked three ways:

- passes in isolation in **1.48s**, a 3.4× margin under the threshold;
- the whole file passes on the **stashed pre-change tree** (51 passed);
- the full `unit-pr` gate re-run passes clean — **23292 passed, 96.71%** coverage.

So it's load sensitivity under full-suite coverage instrumentation, not a regression. Worth someone giving that assertion more headroom or a `slow` marker at some point, since a 5s wall-clock bound will keep tripping on a loaded runner — but that's a change to an unrelated test file and doesn't belong in this PR.

For the record on Codecov: it flip-flopped again this round, 96.20% → 98.99%, with `mcp_server.py` dropping off the list entirely between the two. Same partial-report timing as before.

---

napetrov:
Closing note on coverage — Codecov has settled at **99.49% patch, 0 missing lines**, and the two remaining items don't warrant a change.

They're branch partials (`175->200`, `195->200`) inside `_safe_write_path`'s **Windows-only** block. That function is byte-identical to `origin/main` — verified with a direct diff — and appears in this PR only because the `mcp_server.py` → `mcp_server_inputs.py` split relocated the lines. Platform-specific arcs in moved code, unreachable on the Linux coverage lane, and not something this PR introduced.

**Correcting myself:** I wrote "all four files at 100%" in my previous comment, reading Codecov's "0 Missing" for `mcp_server_inputs.py`. That file is at 100% *statements* but has these two branch partials, so "100%" was imprecise for it. The three files this PR genuinely adds — `service_dump_pipeline.py`, `service_input_resolution.py`, and `mcp_server.py`'s new paths — are at 100% statement and branch.

That closes every coverage item I could act on. The path from the first report to here was: three genuinely untested behavior claims (the `hybrid` and `android` source-tree rejections, and the `ClickException` → `SnapshotError` translation), each asserted in a docstring with nothing exercising it, plus the manifest size guard. All now tested.

---

napetrov:
Pushed `f4c2211`, addressing the three follow-up items from an independent `/review` pass I ran over this PR's own diff earlier. All behavior-neutral — confirmed by the existing suite passing unchanged (466 targeted tests, full `unit-pr` gate: 23292 passed, 96.72% coverage).

**Collapsed the repeated try/except boilerplate.** `abi_dump` had four separate `try: ... except ValueError as exc: return json.dumps(...)` blocks (`phd_paths`, `depth`, the `sources`/`build_info`/`dump_manifest`/compile-context group, the manifest size guard) doing the identical conversion — merged into one, since every validator in that phase raises `ValueError` (`ManifestValidationError` included, it subclasses `ValueError`). `abi_scan`'s two blocks added by this PR merge the same way; its third, *pre-existing* depth check stays separate on purpose — it sits before the unrelated `hdr_paths`/`inc_paths` construction, and folding it in would reorder pre-existing statements for no benefit. Validation order is unchanged in both, so which error a multi-invalid-argument call reports first is unchanged too.

**Promoted the two genuinely cross-module validators to public names.** Of `api_types.py`'s private per-field validators, only `_frontend_value_errors`/`_frontend_context_errors` are actually imported across a module boundary (by `mcp_server_inputs._compile_context_from_args`, for `abi_scan`'s compile-context validation). Renamed to `frontend_value_errors`/`frontend_context_errors` and added to `__all__` — matching how `service_input_resolution.py`'s genuinely cross-module helpers are already named without a leading underscore elsewhere in this same PR. The underscore was never accurate for what these two became once a second module started depending on them.

**Documented rather than changed** why `CompareRequest.replace()`/`DumpRequest.replace()`'s `**changes: Any` is deliberate: a per-field `TypedDict` would need to track every field these dataclasses gain, which is exactly the maintenance burden a request dataclass exists to avoid in the first place.

**One unrelated cleanup:** a stray `SPDX-License-Identifier` line got inserted into `api_types.py`'s header by an editor-level hook partway through this session — removed it, since it was inconsistent with every sibling file's header, including the ones this PR itself adds.

No changelog fragment for this commit: none of it is user-visible behavior, and the PR already carries a fragment covering everything that is.

---

napetrov:
Fixed in `2cd45f5` — real, confirmed, and broader than reported.

**`public_header_dirs` — confirmed exactly as described.** Reproduced before the fix:

```
validation_errors(): []
validate() did NOT raise -- request considered VALID
run_dump_request raised: SnapshotError -- Failed to dump '...': dump_manifest and
public_header_dirs are mutually exclusive -- declare the equivalent in the manifest itself.
```

`dumper.dump()`'s own runtime check names five conflicting parameters when `dump_manifest` is set: `headers`, `extra_includes`, `public_headers`, `public_header_dirs`, `scope_header_dirs`. The Tier-2 pre-flight (`_side_errors`, shared by `CompareRequest` and `DumpRequest`) only checked `headers`.

**`includes` has the identical gap, not named in the review.** `InputSpec.includes` is `dumper.dump()`'s `extra_includes` parameter — traced through `service.resolve_input` — and reproduces the identical failure mode:

```
run_dump_request raised: SnapshotError -- ... dump_manifest and extra_includes are
mutually exclusive -- declare the equivalent in the manifest itself.
```

Both are now caught in one combined check (mirroring `dumper.dump()`'s own grouped-message style), verified after the fix:

```
public_header_dirs   -> caught: dump_manifest and the input side's public_header_dirs are mutually exclusive ...
includes             -> caught: dump_manifest and the input side's includes are mutually exclusive ...
both                 -> caught: dump_manifest and the input side's includes, public_header_dirs are mutually exclusive ...
headers-only: caught: dump_manifest and the input side's headers are mutually exclusive ...   (unchanged)
clean manifest-only errors: []   (no false positive)
```

`scope_header_dirs`, the fifth field `dumper.dump()` checks, is deliberately not added — no typed-request field ever populates it, so there's nothing live to conflict.

**Verified end to end**, not just at `DumpRequest.validate()` in isolation: added a test class that calls `abi_dump` itself and confirms the tool now returns the fast usage error instead of the old late `SnapshotError`.

`scripts/verify.py --profile pr --only lint,typecheck,ai-readiness,unit-pr`: 4 passed, 0 failed — 23297 passed (+5 new), coverage 96.72%.

---

PR #676, opened 2026-08-06
feat: ADR-049 CLI-unification cleanup — output/release parity, aggregate matrix, bundle/CI fixes (CLI-audit P1)

napetrov:
## Summary

Six related ADR-049/CLI-unification fixes from the &#34;CLI-audit P1&#34; status review, landed as separate commits on this branch:

1. **Output-format parity** — SARIF, JUnit, and the native HTML report now carry the same per-finding ADR-049 contract-decision fields JSON/Markdown already have.
2. **Release/package contract-evaluation** — a directory/package `compare` (the per-library release fan-out) now supports `--contract-evaluation`/`--contract`, instead of hard-rejecting them.
3. **Aggregate semantic matrix** — `aggregate`'s `finding_matrix` now carries each affected profile's own ADR-049 contract decision, not just which profiles have the finding.
4. **Bundle compile-context forwarding fix** — `check-project.yml` no longer forwards a per-cell `compile.binding` override onto `bundle` cells, closing an explicitly-acknowledged pre-existing bug.
5. **`--dry-run` doc-accuracy fix** — corrected the "always exits 0" claim in `action.yml`/docs, which was verifiably false (a malformed invocation or an unsatisfiable requested depth still exits nonzero, by design).
6. **Real-package CI lane hardening** — `realworld-validation.yml`'s paths filter now covers contract/comparability/compiler-profile/schema changes, and gained a real end-to-end `--contract-evaluation` step against a real Debian package.

## 1. Output-format parity

Under `--contract-evaluation`, every finding is stamped with a full contract-decision shape: `contract_relevance`, `contract_reason_code`, `contract_assurance`, `compatibility_evaluation_status`, `compatibility_decision`, `gate_contribution`, `contract_evidence_refs`. JSON and Markdown already rendered the whole shape, but:

- **SARIF** only emitted a partial subset, and only for a `NOT_EVALUATED` finding.
- **JUnit** exposed none of this per finding.
- **HTML** showed contract detail only in the &#34;Not Evaluated&#34; appendix section, never for an `IN_CONTRACT`/`NOT_APPLICABLE` finding in the main tables.

`sarif.py`/`junit_report.py`/`html_report.py` now all attach the full canonical field set per finding whenever `contract_relevance` is stamped. Strict no-op for any report that never opts into `--contract-evaluation`.

## 2. Release/package contract-evaluation

Directory/package `compare` (the per-library release fan-out) used to hard-reject `--contract-evaluation`/`--contract` with a &#34;not supported for directory/package comparisons yet&#34; `UsageError`. It now threads both straight into each library pair's own `service.run_compare()` call — the exact same Tier-2 chokepoint a single-pair `compare` uses — so a library compared through the fan-out gets the identical ADR-049 contract decision it would from comparing it individually.

- Each library's own `--output-dir` report already carries the full per-finding contract shape for free (same `to_json()` a single-pair compare produces).
- ADR-049 Phase 7's orthogonal contract-coverage floor is `max()`-aggregated across every library into the release's own exit code, surfaced in the release JSON summary as `contract_coverage_exit_contribution` (same field name/fold rule as single-pair `compare`). `--fail-on-removed-library`'s exit `8` is checked ahead of this coverage-only fallback, so a removed library's own signal is never masked by an unrelated coverage gap.
- `--pack` remains rejected for directory/package comparisons — applying a pack's policy/contract/gate overrides per library still needs its own resolve-once-apply-per-pair design (ADR-049 D8 conflict detection resolves against a single library pair today).

## 3. Aggregate semantic matrix

`scope`/`affected_profiles` answer *whether* a profile has a finding, not *why* compatibility policy did or didn't act on it. Two profiles compared under different `--contract` domains can both report the same finding while reaching opposite conclusions — one's evidence proves `IN_CONTRACT` (gating), the other's leaves it `UNKNOWN_UNRESOLVED` (not gating) — which `scope: all_profiles` alone would flatten into &#34;uniformly understood&#34;.

- When at least one affected profile ran `--contract-evaluation`, its `finding_matrix` entry now gains a `profile_contract` array — one entry per affected profile (in `affected_profiles` order), each with that profile's own `contract_relevance`, `compatibility_evaluation_status`, `compatibility_decision`, and `gate_contribution`, read back verbatim from its own report.
- Purely additive and inert for a comparison that never evaluated a contract — the field is omitted entirely, so an ADR-049-unaware CI matrix's `finding_matrix` renders byte-identical to before.
- Aggregate schema bumped to `1.4`; `docs/use/aggregate-reports.md` documents the field with a worked GCC-`IN_CONTRACT`/Clang-`UNKNOWN_UNRESOLVED` example matching the review's own example.

## 4. Bundle compile-context forwarding fix

A bundle cell's operand is the bundle-staging *directory*, and `cli_resolve._reject_compile_context_for_set_inputs` already hard-rejects `--gcc-path`/`--gcc-options` for a directory/package compare (the per-library release fan-out never threads a single-pair L2 compile context to each pair). `check-project.yml`'s reusable workflow was forwarding a profile's per-cell `compile.binding` override onto bundle cells too, turning a previously working bundle check into a hard operational error — flagged in the G34 producer/consumer-compiler-profile-separation plan doc as a known, unfixed pre-existing bug. Fixed the same way the identical `ast-frontend` hazard was already fixed: gated on `matrix.kind != 'bundle'`, falling back to the workflow-global input for a bundle cell rather than silently dropping it.

## 5. `--dry-run` doc-accuracy fix

`action.yml`/docs claimed `--dry-run` &#34;always exits 0&#34;. Verified against the code that this is false: `abicheck/dry_run.py`'s `DryRunResult.blockers` exits 1 when a requested depth/evidence contract can't be satisfied, and `dump_cmd` raises `UsageError`/`BadParameter` (exit 64) for a malformed invocation unconditionally before the `--dry-run` branch, by design. Corrected the docs to state the real contract, carving out the one genuine exception (the composite Action's own `abi-baseline` auto-fetch, deliberately tolerated under `--dry-run`).

## 6. Real-package CI lane hardening

`realworld-validation.yml`'s `pull_request` `paths` filter was scoped only to package/bundle plumbing, so a change to contract evaluation, comparability, compiler-profile resolution, or a report schema could silently skip the one lane that runs `abicheck compare` against a real package binary rather than a synthetic snapshot — a gap the review named explicitly. Extended the filter to cover those subsystems, and added a second real-package compare step exercising `--contract-evaluation --contract public` end to end — a paths-filter fix alone would only trigger the lane, not actually cover the feature. Verified locally against a real Ubuntu `zlib1g`/`zlib1g-dev` `.deb` pair before committing.

## Testing

- Focused unit + CLI-level + workflow-structure tests added for every change above (positive coverage replacing old rejection tests, exit-code-fold precedence tests, JSON-field-presence tests, schema validation, workflow-forwarding assertions, paths-filter/step-content assertions).
- Full fast unit suite (`pytest tests/ -m &#34;not integration and not libabigail and not abicc and not slow and not golden&#34;`) passes: 23400+ passed, 0 failed.
- `mypy abicheck/`, `ruff check`, `ruff format --check` (on touched lines — several touched files already carried pre-existing ruff-format drift unrelated to these changes; diff kept scoped to what was actually touched) all clean.
- `scripts/check_ai_readiness.py`: 0 errors.
- Full CI green on HEAD (across multiple commits, including a transient Windows CMake-build timeout in `integration-tests (windows-latest)` that cleared on retry — unrelated to this diff).
- CodeRabbit review findings addressed across all rounds (see individual commits).

## Checklist

- [x] Changelog fragments added (`scriv create`)
- [x] Docs updated (`docs/reference/cli-reference.md`, `docs/reference/github-action-inputs.md` regenerated; `docs/reference/exit-codes.md`, `AGENTS.md`'s exit-codes section, `docs/use/aggregate-reports.md`, `docs/use/github-action.md`, `docs/use/scan-levels.md` updated)
- [x] `mypy`/`ruff` clean, no new errors

## Scope note

This closes the review's &#34;output formats&#34; item in full, the &#34;release/package contract-evaluation parity&#34; item for the `--contract-evaluation`/`--contract` half, the &#34;aggregate: next semantic gap&#34; item, the bundle+compile-overlay pre-existing bug, the dry-run doc-accuracy item, and part of the CI-hardening item (real-world eval path coverage). Investigated and deliberately declined this session, each requiring its own separately-scoped follow-up (found real, non-trivial risk on empirical investigation, not just size): `--pack` support in the release fan-out and MCP `abi_compare` (both risk reintroducing the "decorative `--pack`" bug class this codebase's history shows is easy to get wrong), the three public-contract-default gate losses (root-caused to a template-argument-handling gap in `surface.py`'s shared public-type closure — the foundational module every detector depends on, not a small additive fix as initially assessed), consumer-graph Phase 2, producer/consumer two-pass extraction, and native `dump` CLI → `DumpRequest` convergence (confirmed via investigation: the seam Phase 2 needed for `compare` doesn't yet exist for `dump`, and the ELF dump path bypasses the Tier-2 chokepoint entirely today). Also out of reach for this session: the Action A/B stable-API-vs-wrapper decision (a product decision, not a technical one) and branch-protection required-checks/Codecov SHA association (GitHub admin-panel settings, no available tool to change them).

---

Co-Authored-By: Claude Sonnet 5


napetrov:
Status check-in: this run's two failures so far (`unit-tests (windows-latest, 3.13, false)` and `integration-tests (windows-latest)`) are both the identical GitHub Actions infra error — `Failed to resolve action download info. Error: Service Unavailable` — failing during action setup, before any of this PR's code runs. 16 of 18 jobs in this run are still stuck `queued` well past normal start time, which points to a broader runner-provisioning outage rather than anything in this diff.

No code action needed. Once the run finishes (or the outage clears), I'll re-run the failed jobs. No unresolved review comments — CodeRabbit's only comment so far was a rate-limit notice with no findings.

---

napetrov:
Pushed a fix for `codecov/patch` (was 82%, target 90%): the 9 flagged partial branches were all independent optional `Change` fields (`contract_reason_code`/`contract_assurance`/`contract_evidence_refs`) whose "unset" side had no test, plus one genuinely unreachable branch (`evaluation_status_of()` always resolves to a real status once `contract_relevance` is known non-`None`) that I removed rather than test around.

- Added a test per format for a stamped-but-otherwise-empty finding to cover the optional fields' false branches.
- Removed the dead `if status is not None` guard in both files (`cast` keeps mypy happy without reintroducing it).

Also worth flagging: both touched files already had pre-existing `ruff format` drift unrelated to this PR (my local ruff run wanted to reflow ~150 unrelated lines in each — looks like a ruff-version mismatch against whatever last formatted `main`, since `pyproject.toml` pins `ruff>=0.3` with no upper bound). I kept this PR's diff scoped to only the lines it actually touches rather than pulling in that unrelated reflow.

Full fast unit suite (23386 passed), mypy, and ruff check are all clean on the new commit.

---

napetrov:
Addressed CodeRabbit's nitpick — extended `test_html_evaluated_finding_carries_contract_decision` to cover `compatibility_decision`/`contract_evidence_refs` too, and added a companion "stamped but no optional fields" test (same pattern as the SARIF/JUnit tests). This also closes the `codecov/patch` gap the HTML commit introduced: each of the badge's four optional-field `if` statements only had its "set" branch exercised locally.

On the `codecov/patch` failure itself (58.33%, target 90%): the canonical coverage lane (`unit-tests ubuntu-latest, 3.13, false`) was still `queued` when that check ran — so it was comparing against a stale upload from before the HTML/coverage-gap commits existed, not against current code. Verified locally with `--cov-branch` that the actual diff has no gaps in the touched lines. Will keep an eye on it once the real lane catches up.

---

napetrov:
Addressed CodeRabbit's latest review:

- **Merge conflict** — resolved (merged `origin/main`, which had picked up a stale-in-a-different-way fix to the same `--contract-evaluation` help text via #675; kept this branch's more up-to-date version and confirmed it matches `scripts/gen_cli_reference.py`'s output byte-for-byte).
- **Precedence wording bug (real, confirmed)** — `AGENTS.md` and the changelog fragment both said the coverage-only exit `1` was "checked ahead of" `--fail-on-removed-library`'s exit `8`, backwards from what `_exit_compare_release` actually does (removed-library `8` is checked first; coverage-only `1` is the fallback). `docs/reference/exit-codes.md` already had it right — fixed the other two to match, and fixed the same sentence in the PR description above.
- **Test exit-code specificity** — the two invalid-invocation tests now assert exit `64` (`cli._EXIT_USAGE_ERROR`) specifically, not just "nonzero".
- **Parametrize nitpick** — `test_zero_contribution_is_a_true_no_op` now uses `pytest.mark.parametrize` instead of a manual loop.
- **`docs/reference/cli-reference.md` hand-edit concern** — it was already regenerated via `scripts/gen_cli_reference.py` (confirmed again after the merge — no diff), not hand-edited; the row just looked identical to a hand-edit in the diff view.
- **Docstring-coverage warning (12.5%)** — not acted on: this repo's own `test-assertion-density`/AI-readiness gates don't require per-function docstrings on test functions, and the touched test files already follow the file's existing inline-comment convention rather than per-function docstrings.

Full fast unit suite re-verified after the merge (23404 passed), ai-readiness gate clean.

---

napetrov:
Status check-in on the `codecov/patch` failure (9.85%): same race condition seen twice earlier on this PR — the canonical coverage lane (`unit-tests ubuntu-latest, 3.13`) was still `in_progress` when codecov posted, so it's scoring against a stale upload from before this merge commit. This isn't a real regression — I verified locally with `--cov-branch` on every touched file after each commit and confirmed no gaps. Will keep watching; expect codecov to re-post with accurate data once that lane finishes (as it did the last two times).

---

PR #683, opened 2026-08-09
docs: ADR-058 and G36 — Native Compatibility Agent Skills architecture

napetrov:
## Summary

Adds ADR-058 and its companion plan G36, defining the architecture and product model for native software compatibility Agent Skills published by abicheck.

## Motivation

abicheck has no agent-skill presence today, and every existing integration surface (CLI, MCP, Actions) is only reachable by a caller who already knows abicheck exists. The 2026 Agent Skills ecosystem (Claude Code, Copilot, Codex, Cursor, Gemini CLI) has converged on a portable `SKILL.md`/`.agents/skills/` format, and a grounded ecosystem search found no existing published skill that performs deterministic native ABI/binary-compatibility analysis — a real, validated, currently-unoccupied opportunity. This work was grounded in a full inspection of the current repo (CLI command tree, MCP tools, typed Request/Result objects, ADR-047's scenario-first precedent, ADR-043/054's CLI admission bar, docs authoring rules, examples/validation corpus) and in external research into the current Claude/Copilot/Codex/Cursor/Gemini/skills.sh/MCP-Registry landscape.

## Changes

- **`docs/contribute/adr/058-native-compatibility-agent-skills.md`** — the ADR: a user-task-first portfolio of four P0 skills (`native-binary-compatibility-review`, `native-api-evolution`, `native-release-compatibility`, `native-consumer-compatibility`), explicit skill-admission criteria, a three-layer model (public skills / shared domain knowledge / execution backends), a CLI-normative / MCP-optional execution decision, a `skills-src/` → generated `.agents/skills/` publication model with no symlinks, eleven non-negotiable safety invariants (never manufacture a false green result), and the minimal real product gaps identified (`abicheck info --format json`, a stable comparability reason-code field) — deliberately not resurrecting `doctor`/`init`/the baseline registry.
- **`docs/contribute/plans/g36-native-compatibility-agent-skills.md`** — the phased implementation plan (P0 architecture/first release, P1 reliability/distribution, P2 contingent portfolio expansion), mirroring G30's item format (Problem/Change/Files/Tests/Docs/Dependencies/PR boundary/Status).
- Wired both new documents into `docs/contribute/adr/index.md` and `docs/contribute/plans/index.md` per this repo's existing conventions.

## Checklist

- [x] Docs updated (this PR *is* the docs change)
- [x] No changelog fragment — this PR touches no `abicheck/**/*.py`
- [x] `python scripts/check_ai_readiness.py` passes with 0 errors (ADR-index-nav-sync, claude-md-coverage, doc-count-sync all clean for the new files)
- [ ] Tests / `pre-commit run --all-files` — not applicable, no code changed

---

---
_Generated by [Claude Code](https://claude.ai/code/session_018LrbhdzGMh1s9AV6dv9pVx)_

## Summary by CodeRabbit

- **Documentation**
  - Added an accepted architecture decision for native compatibility workflows covering binary changes, API evolution, release compatibility, and consumer compatibility.
  - Added an implementation plan outlining phased delivery, validation, safety requirements, reporting improvements, and cross-tool consistency.
  - Updated documentation indexes with links, status, scope, and implementation tracking for the new decision and initiative plan.


napetrov:
Pushed a fix-up commit (51bcb27) addressing all five specific Codex review comments on this PR, plus several more issues an independent review pass found:

- P0.9's `doc_type: use` → `doc_type: how-to` (the correct value from `docs/AGENTS.md`'s enum).
- Added the missing front matter to the G36 plan page.
- The comparability reason-code proposal (ADR §"Required abicheck product capabilities", G36 P0.5) now targets the JSON field that actually exists today (`reason.kind`/`reason.message`) instead of a nonexistent top-level `comparability` block, and its enum now covers every `PROFILE_FIELD_KEYS`/`SCOPE_FIELD_KEYS` mismatch cause plus an explicit fallback code.
- P1.5's validation-log page now requires an `mkdocs.yml` nav entry, matching how its own cited precedent (`abicc-parity-status.md`) is actually navigated.
- Also fixed, found independently: a dangling "Product positioning" section reference (added the section), a misquote of ADR-047's text, a fabricated "~31 commands" figure, an 8-vs-11-file mismatch between the ADR's `skills-src/shared/` tree and the plan's P0.1 item, and a duplicated admission-bar paraphrase (now both point at `AGENTS.md`/ADR-054 D6 instead of independently restating it).

`python scripts/check_ai_readiness.py` still reports 0 errors after these changes.


---

napetrov:
Pushed dedb472 addressing this second round of Codex comments:

- `reason.code` → `reason.codes` (array) — a comparability mismatch can hit multiple fields simultaneously (e.g. `compiler_family` and `abi_dialect` at once); a singular code would force discarding one cause or inventing an arbitrary precedence, and a skill fixing the reported cause would just fail again on the silently omitted one.
- Added the DPC++-only `frontend_context_kind` field to the enum coverage with its own dedicated code, rather than letting it fall into the generic fallback.
- Fixed the MCP `abi_compare` proposal: its `not_comparable` envelope emits `reason` as a bare string today, not an object like the CLI's `{kind, message}` — now specified as an additive `reason_codes` sibling field, not a breaking type change to the existing `reason` field.
- Softened P0.4's `abicheck info` admission-bar claim: criterion 2 ("operand is a domain object") is the one genuinely debatable call for a zero-operand command, now flagged as needing explicit maintainer sign-off when that PR actually lands, with `--version` cited as the existing precedent, rather than asserted as mechanically cleared.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 748616b (amended from an earlier commit whose message got mangled by shell backtick expansion — force-pushed the corrected version) addressing this third round:

- **`scan --against` reason-code gap**: confirmed `abicheck/scan_engine.py` hits the same `ProfileMismatchError`/`ScopeMismatchError` at its own comparability gate and today emits only free text (`diff_summary = {"reason": str(exc)}`). Added it to P0.5's file list and a scan/compare parity test, so `scan`'s CLI/MCP surfaces (sharing the engine) aren't left out of the typed-reason-code work.
- **P0.9 dangling-link risk**: fixed — `.agents/skills/` lives outside the `docs/` tree `mkdocs` builds from, so a relative link from `docs/use/agent-skills.md` to a generated `SKILL.md` would be unresolved under `mkdocs build --strict`. The plan now specifies this repo's existing full-GitHub-URL convention for linking a non-`docs/` path (matching `docs/use/security-hardening.md`'s existing pattern), not a relative link.
- **ADR-058 front matter**: not changed, and I want to flag why rather than silently skip it. Of the 57 ADRs in `docs/contribute/adr/`, only one (ADR-051) carries YAML front matter — the corpus has consistently not applied `docs/AGENTS.md`'s general front-matter rule to ADRs, using the `adr/index.md` Status-line convention as its own metadata mechanism instead. Adding front matter to this one new ADR alone would make it the outlier, not the compliant one. Happy to revisit if a maintainer wants ADRs brought into the front-matter schema as a deliberate, repo-wide decision — but that's a broader call than this PR should make unilaterally.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 0bc10d6 addressing this fourth round:

- Removed the hand-copied `ChangeKind` (396) and examples-corpus (195) counts from both documents — both already have a fact owner elsewhere (`checker_policy.py`, `ground_truth.json`) per `docs/AGENTS.md`'s own convention against hand-copying a volatile count; referencing the catalogs by name is sufficient.
- Closed the P0.2/P0.4/P0.5 sequencing gap: P0.2's `SKILL.md` files correctly can't reference `abicheck info` or `reason_codes` before those product changes land (P0.2 only depends on P0.1), but nothing integrated them into the already-authored skills afterward — added an explicit small follow-up-PR note to P0.2, P0.4, and P0.5.
- P0.8: split trigger tests into a deterministic half (description-vs-corpus matching — stays in the required `pr` profile) and a real-agent-invocation half (needs a vendor binary, credentials, network, tolerates nondeterministic routing — now explicitly its own opt-in/scheduled lane, mirroring how this repo already keeps `integration`/`libabigail`/`abicc` external-tool markers out of the default fast test command).

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed d8e57ba addressing this fifth round:

- P0.5: added `abicheck/schemas/compare_report.schema.json` to the file list — the `reason` object's fields live in the JSON Schema itself, not in `schemas/__init__.py` (version constants/registry only); without the schema update, P0.7's schema-based drift check wouldn't recognize `codes` at all.
- P0.5: added `abicheck/cli_compare_release_helpers.py` (`_compare_one_library`/`_format_release_json`) plus a regression test — a not-comparable library in a directory/package release today returns only `reason: str(exc)` into the actual `summary.json`/release-JSON entry, which is what `native-release-compatibility` reads, and is distinct from the separate not-comparable document the original file list already covered.
- ADR: scoped the trigger-test "must select the intended skill" requirement to the five of seven Product-positioning phrasings that actually map to a P0 skill — it previously required all seven, which directly contradicted P0.8's own carve-out for the two P1-candidate phrasings.
- G36 P0.1: clarified that `skills-src/shared/safety-invariants.md` is a one-time extraction from ADR-058's frozen decision-record text (ADRs here aren't edited after acceptance, only their Status line), not a second independently-maintained living copy — once it exists, it's where a future correction actually happens.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 1185d44 addressing this sixth round:

- P0.3: `scripts/gen_agent_skills.py` now gets registered in `scripts/CLAUDE.md`'s inventory table in the same PR, so the `script-inventory` check doesn't warn on it.
- P0.5: added `abicheck/aggregate.py` and a round-trip test — `TargetReport.reason` is itself a bare `str`, and `_load_report_file` flattens a per-target report's structured `reason` object straight into it, so `native-release-compatibility`'s documented `project`/`aggregate` multi-profile path would stay untyped even after every single-comparison producer emits `codes`.
- P0.6: extended its scope to actually deliver the shared-fragment summary-drift check P0.1 only motivated but never wired anywhere real. Rather than inventing a new bespoke checker, each `skills-src/shared/*.md` fragment now registers its `summarizes` claim in the existing `docs/_meta/topics.yaml` registry, reusing `check_docs_contract.py`'s already-enforced round-trip (a page can't claim to summarize a topic it isn't registered against) for free.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 77311a3 addressing this seventh round:

- P0.6: the proposed `topics.yaml` registration wouldn't actually pass `check_docs_contract.py` as written — `task_pages`/`allowed_summaries` entries are hard-checked via `_is_file_under(DOCS, ...)`, and `skills-src/` lives outside `docs/` by design. Added an explicit sub-task extending that validation to accept a repo-relative non-`docs/` path (mirroring how `fact_sources` already tolerates non-doc code paths), plus a test for the extended behavior itself, not just its downstream effect.
- P0.5: made explicit that the `reason.codes` completeness promise is scoped to *within* whichever single exception `check_contracts_comparable`'s gate raises, not across domains — the gate short-circuits at the first failing check (scope before profile), so a pair differing on both dimensions still only surfaces one domain's codes per run. Documented as a deliberate, named boundary in both documents rather than overclaiming full completeness, with tests scoped to match.
- P0.5: added `abicheck/stack_checker.py`/`abicheck/stack_report.py` (`deps compare`'s own `not_comparable_reason: str` field) and a regression test, so `deps compare --format json` isn't left behind the other four producers.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 77fdcda addressing this eighth round:

- P0.6: extended not just `task_pages`/`allowed_summaries` path validation but `_check_front_matter_schema()`'s scan itself — it iterates only `DOCS.rglob("*.md")` today, so a `skills-src/shared/*.md` fragment's `summarizes` claim would stay unchecked regardless of how permissive the registry validation becomes. Added the scan/path-identity extension plus a negative-case test (unknown topic id → checker fails), so the enforcement is proven, not just plumbed.
- P0.5: `_load_report_file()` only recognizes the compare-shaped `verdict: null` + `reason: {...}` object; a `scan --against` not-comparable report is shaped differently (`verdict: "NOT_COMPARABLE"`, reason nested under `diff`), so the aggregate extraction needs its own scan-specific branch or a `scan`-backed project matrix would silently lose the field. Added that plus a dedicated test for the scan-shaped path, and called out that `AGGREGATE_SCHEMA_VERSION` needs its own explicit bump, independent of the compare/scan report schema bumps.

`python scripts/check_ai_readiness.py` still reports 0 errors.

This is round 8 on what's fundamentally a planning document, and the remaining thread (P0.5's `reason_codes`/aggregate plumbing) is converging on genuinely small, correct details rather than surfacing new architectural problems. I'll keep addressing anything substantive that comes in, but flagging this so it's visible if the review loop is expected to continue at this depth.


---

napetrov:
Pushed 49a6b45 addressing this ninth round:

- P0.1: `root-cause-grouping.md` now points skills at `compare --report-mode root-cause --format json`'s existing deterministic `root_causes`/`root_cause_count` fields first, rather than re-deriving a grouping in skill-side prose (which risked a lower-fidelity result diverging from abicheck's own answer).
- P0.4: expanded `abicheck info`'s payload to all six schema-artifact families the `current()` registry already exposes (`snapshot`/`compare`/`scan`/`aggregate`/`build-output`/`run-plan`), not just three — `native-release-compatibility`'s `project plan`/`aggregate` workflow needs the other three too.
- P1.1: fixed the case-indexing source — `validation/data/manifest.json` indexes 11 real-world package pairs keyed by `pair`, not `examples/` case IDs; corrected to `examples/ground_truth.json`, the actual canonical per-case catalog.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 7fc15d1 addressing this tenth round:

- P0.4: `schemas._ARTIFACT_NAMES` is private — `cli_info.py` would have to import it directly, and a future artifact family registered only in `current()`'s internals could silently fall out of `info`'s "all six" payload. Added a small public accessor and a test against it instead of a hand-copied name list.
- P1.2: fixed stale framing — an earlier draft assumed skills group findings themselves, contradicting P0.1's own `root-cause-grouping.md` instruction (added in round 9) to consume `compare --report-mode root-cause`'s existing `root_causes` output first. Narrowed this contingent item to "a genuine coverage gap in that existing output," not "skill-side grouping proved inadequate."
- P1.1: `examples/ground_truth.json`'s per-case entries can't resolve the non-comparable/consumer-scoping/plugin-required-symbol/missing-matrix-target scenarios — they need invocation parameters (`--used-by`, `--required-symbol(s)`, a multi-target matrix) the case catalog doesn't carry. Split those into a second, explicit scenario manifest rather than claiming the case index covers the whole matrix.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 87577f5 addressing this eleventh round:

- P0.3: a copied `shared/*.md` fragment's own outbound links to canonical `docs/` pages weren't rewritten by the generator (only the skill's own `SKILL.md` link to its local fragment copy was) — fine inside this repo's checkout, but dangling in a standalone marketplace/`skills.sh` install where `docs/` isn't present. Now rewritten to the published `abicheck.github.io/abicheck` docs host at generation time, with a test covering the whole generated tree.
- P0.6: extended the "do not hand-edit" generated-file marker to every generated Markdown file under `.agents/skills/**` (skill-specific `references/*.md` and copied `shared/*.md` fragments too), not only each skill's top-level `SKILL.md`.
- P0.5: `deps compare`'s stack JSON output carries no schema version marker at all today (confirmed: no `schema_version` field/constant anywhere in `stack_checker.py`/`stack_report.py`), so the schema-based drift check this item claims couldn't actually cover it. Added a minimal `stack_schema_version` as part of the same change.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 5b762ae addressing this twelfth round:

- P0.5: added a dedicated `dependency_scope_mismatch` code — `comparability.py` already models `dependency_scope` as its own distinct mismatch kind (`_check_dependency_scope_comparable`), separate from scope/profile, but the enum coverage had no matching code for it.
- P0.5: flagged and scoped fixing a real **pre-existing gate bug** that blocks the `frontend_context_kind` code from ever firing — `check_contracts_comparable`'s fingerprint-authenticity re-verification always checks against the 11-key `PROFILE_FIELD_KEYS`, even for a DPC++ contract whose fingerprint was hashed over the 12-key `_FRONTEND_CONTEXT_PROFILE_FIELD_KEYS`, so a DPC++ pair fails authenticity before the correctly-keyed differing-fields computation ever runs. Confirmed against the actual code; this item now includes fixing that gate's key-set selection, since the dedicated code is otherwise unreachable.
- P0.5: the `deps compare` schema fix needs the actual JSON Schema file (`stack_report.schema.json`) and its published mirror, not just a version constant — a bare version number gives P0.7's drift check nothing to validate against.
- ADR: disambiguated "`SCHEMA_VERSION`-gated field" as `REPORT_SCHEMA_VERSION` explicitly, distinct from `serialization.SCHEMA_VERSION` (which versions `AbiSnapshot` files, not reports).

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed eb58649 addressing CodeRabbit's first full review (4 actionable comments) plus this round's Codex findings:

- Softened the "`.agents/skills` is canonical" claim — presented as the right portable default, not a claim that vendor-specific directories (`.github/skills`, etc.) become unnecessary; cross-agent validation is what actually confirms that per client.
- Fixed the MD040 markdown-lint warning (missing language identifier on the directory-tree code fence).
- P0.3: the generator only scanned each skill's `SKILL.md` for shared-fragment references; a fragment referenced only from a skill-specific `references/*.md` file would be silently missed. Now scans every file under a skill's own `references/`, resolves transitively, and rewrites every referencing file's links — plus fixtures for both the indirect- and transitive-reference cases.
- P0.8: stopped duplicating the seven trigger phrases from ADR-058's Product positioning section — now pulled from that one canonical source instead of repeated a second time.
- P0.4: dropped the "all six" schema-artifact count, which was already stale against P0.5's own dependency-stack schema addition (a 7th, independently-landable family). `info`'s payload and test are now described/asserted solely against the public `schemas.artifact_names()` accessor's live output.
- Removed a stale "schema 2.17" literal (current `REPORT_SCHEMA_VERSION` is 2.27) from both documents — refer to the fact-owned constant instead of a volatile number that had already drifted once.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 914f54d addressing this fourteenth round:

- P0.4: `abicheck info` is a small, one-function, no-operand command with no significant helpers — per `AGENTS.md`'s own convention it belongs directly in `cli.py`, not in a sibling `cli_info.py` module (reserved for larger commands/groups).
- P0.5: the same fingerprint-authenticity gate bug fixed for the DPC++ profile case also exists on the scope side — `check_contracts_comparable`'s scope authenticity re-verification always re-hashes against plain `SCOPE_FIELD_KEYS`, even when a contract is manifest-derived (whose `scope_fingerprint` is hashed over the wider `_MANIFEST_SCOPE_FIELD_KEYS`), so a manifest scope mismatch fails authenticity before the real diff runs — leaving the `translation_units` code unreachable the same way. Extended the fix to both scope authenticity call sites too.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed c9e81df addressing this fifteenth round:

- P0.5: neither the release JSON (`_format_release_json`) nor the release summary (`_write_release_summary_file`) carries a schema version marker or has a registered JSON Schema today — same undiscoverable/unvalidatable gap already fixed for `deps compare`'s stack output in an earlier round. Now gives release summaries their own schema version, registered schema, and published mirror as part of the same change.
- P0.5: `SCAN_SCHEMA_VERSION` is a version marker with no schema behind it — no `scan_report.schema.json` exists (only `aggregate_report`/`build_evidence`/`build_source_pack`/`compare_report` do), and `ScanOutcome.diff_summary` is typed as plain `dict[str, Any]`. P0.7's schema-based drift check had nothing to validate scan's `reason_codes` against either — added the actual schema and mirror.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed f2f5a02 addressing this sixteenth round:

- P0.5: fixed the manifest-scope authenticity key-set selection to be per-side, not per-comparison — an earlier draft's "use `_MANIFEST_SCOPE_FIELD_KEYS` when either side is manifest-derived" rule was itself wrong: it would re-hash a legacy-hashed side with a field its own fingerprint was never computed over, newly failing that side's own authenticity check. Each `_fingerprint_matches_fields` call must decide its key set from that call's own side. Added a mixed-sides regression test.
- P0.5: `ProfileMismatchError`/`ScopeMismatchError` are actually defined in `abicheck/errors.py`, not `comparability.py` — added `errors.py` as the owner for the constructor/data-contract change.
- P0.4: stopped treating the `info` admission-bar gap as a formality to flag at merge time. `--version` is an eager `click.version_option` flag, not a root command, so it was never a valid precedent for clearing criterion 2. This item is now explicitly blocked on an upfront maintainer decision (a documented bar exception, or redesigning `info` as a `--version --format json` extension instead of a new verb) — not an implementer's judgment call at PR time. Synced the ADR to the same framing.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 7ddd36a addressing this seventeenth round:

- P0.5: `SCAN_SCHEMA_VERSION` marks three distinct shapes, not one — `ScanOutcome.to_dict()` (`scan_engine.py`) and `ScanResult.to_dict()`/`ScanSetResult.to_dict()` (`service_scan.py`), all structurally different. Testing only `scan --against`'s CLI output would leave the service/MCP and multi-artifact envelopes free to claim the same schema version unvalidated. Now requires the schema (`oneOf` branches or distinct families) and tests to cover all three.
- ADR: fixed acceptance criterion 3, which still hard-required the exact command `abicheck info --format json` even though P0.4 now leaves that spelling open pending a maintainer decision — rephrased around the approved capability-discovery surface, not a specific command name.
- P1.1 eval harness: `examples/ground_truth.json` is not top-level keyed by case directory — its top-level keys are file-wide metadata, and the 195 case records live nested under `ground_truth["verdicts"]`. Corrected the described lookup path and added a fixture-resolution test for it.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 136a495 addressing this eighteenth round:

- P0.5: fixing fingerprint authentication alone isn't sufficient — `unknown_differing` (profile side) and `scope_unknown_differing` (scope side) each independently check against the narrow `PROFILE_FIELD_KEYS`/`SCOPE_FIELD_KEYS`, not the widened per-side key sets, so even once authentication passes, `frontend_context_kind`/`translation_units` would still be classified as unknown and fall into the generic fallback instead of getting their dedicated codes. Extended the per-side widened-key-selection fix through both classification computations, not just the two authenticity call sites.
- P0.5: registered the new release-report schema family (and every new family this item adds) in the public `_ARTIFACT_NAMES`/`current()` catalog, not just the version constants — `abicheck info`'s discovery payload derives exclusively from that registry, so an unregistered schema stays invisible to the exact capability-discovery surface this initiative exists to support.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 55d7b83 addressing this nineteenth round:

- P0.5: define the reason-code enum once, at the Python level (`abicheck/comparability.py`), with every one of its six producers importing and emitting from that single definition instead of hand-copying string values. Added a schema-sync test asserting the Python enum and `compare_report.schema.json`'s own `enum:` list stay in lockstep — the same "one fact, one place, checked automatically" pattern this repo already applies elsewhere.
- P0.5: the release JSON (`_format_release_json`) and release summary (`_write_release_summary_file`) are two genuinely different envelope shapes, not one — required either two distinct schema families or explicit `oneOf` branches, plus real-output validation tests for both shapes.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed 8776634 addressing this twentieth round:

- P0.5: corrected the module ownership for the release-summary functions — `_compare_one_library` and `_write_release_summary_file` are actually defined in `cli_compare_release.py`, not `cli_compare_release_helpers.py`; only `_format_release_json` lives in the helpers module. The file list previously attributed all three to the helpers module, which would have left the compact `summary.json` writer's schema-version marker and `reason_codes` field unimplemented.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Pushed af138a9 addressing this twenty-first round:

- P0.4: the `info` payload test asserted only key equality against `schemas.artifact_names()` — a stale or hard-coded version value for a real key would still pass. Now asserts the full `{name: version}` mapping against `schemas.all_current()`, not just its key set.
- P0.5: the reason-code schema-sync test was scoped to `compare_report.schema.json` alone, but this item's own schema work requires the same enum republished in the scan/release/stack report schemas too — a sync test checking only one would stay green while the others silently drifted. Now either every schema `$ref`s one shared enum fragment, or the sync test explicitly checks every report schema that republishes the enum.

`python scripts/check_ai_readiness.py` still reports 0 errors.


---

napetrov:
Round 22 addressed (commit 8fd81f9) — the first fully non-rate-limited CodeRabbit pass plus three new Codex findings:

1. **ADR-058 (CodeRabbit, major):** Corrected the "Canonical portable default" bullet — Claude Code doesn't read `.agents/skills` natively; its documented convention is `.claude/skills/<skill-name>/SKILL.md`. Fixed the client list and clarified the generator produces that path as Claude Code's own packaging target, same as `.github/skills`/`.cursor/skills` for their clients.
2. **G36 P0.4 (CodeRabbit, minor):** Removed the hand-copied "five of the six criteria" count — named the unmet criterion (2) directly instead.
3. **G36 P1.1 (CodeRabbit, minor):** Removed the redundant hand-copied "195" case-record count.
4. **G36 P0.5 (CodeRabbit, minor):** Fixed the "six different producers" vs. seven-name-list inconsistency; clarified `deps compare`/`stack` as one producer sharing a single schema, for a consistent total of seven.
5. **G36 P0.3 (CodeRabbit, nitpick):** Changed the generator description to read `mkdocs.yml`'s `site_url` dynamically at generation time rather than hardcoding the literal docs URL, with tests deriving expected output from the same source.
6. **G36 P0.7 (Codex):** Narrowed the "must fail CI, not silently rot" claim to what this item actually detects — syntactic drift (renamed/removed commands, flags, field paths) only, explicitly not semantic/prose staleness (unchanged names with changed meaning).
7. **G36 P0.8 (Codex):** Replaced the "manual sync point" acknowledgment with a real automated assertion: the corpus test now extracts the seven canonical prompts from ADR-058's Product positioning section and fails on any mismatch, closing the ADR-vs-corpus drift risk.
8. **G36 P1.4/P1.5 (Codex):** Made a minimum cross-agent validation pass (Codex, Copilot) an explicit precondition of P1.4's public submission step, so skills can no longer go public before ADR-058's cross-agent validation requirement is even minimally satisfied.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 23 (commit 0017b27): CodeRabbit confirmed the round-22 fixes as addressed (Claude Code publication target, hand-copied counts, producer count). Codex's three findings against af138a9 (reference-content overclaim, trigger-corpus sync, P1.4/P1.5 ordering) were already resolved by that same commit (8fd81f9), pushed before this Codex pass ran.

One new, legitimate finding from Codex against 8fd81f9: fixing the ADR to say Claude Code reads the generated `.claude/skills/<skill-name>/SKILL.md` (not `.agents/skills`) left P0.3's own generator scope inconsistent — it only committed `.agents/skills/**`, so Claude Code's actual read location in this repo would have been an untracked manual copy, not a tested, reproducible generation output. Fixed: P0.3 now generates and commits both `.agents/skills/**` and `.claude/skills/**` from the same resolved skill content in one pass, with the drift-check step covering both trees.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 24 addressed (commit 3e3de0b) — three Codex findings against 0017b27:

1. **ADR-058:** Removed the hand-copied `0.5.0` package-version literal from the Versioning and drift model section; it now points at `pyproject.toml`'s `version` key as the fact owner, per the repo's "don't hand-copy a fact with an owner" rule.
2. **G36 P0.6:** The generated-file marker convention and `GENERATED_FILE_MARKERS` registration now explicitly cover both committed output trees (`.agents/skills/**` and `.claude/skills/**`) — after round 23 added `.claude/skills/**` as a second generated tree, the marker/ownership check had been left enumerating only the first. Test plan updated to plant marker-less fixtures in both trees.
3. **G36 P0.7:** Added an explicit version-range enforcement check, distinct from the syntactic tool/API drift check — a skill's declared abicheck version range (in its `metadata` frontmatter) is now checked against the installed package version, failing loudly once exceeded, closing the gap where an unrenamed CLI surface outside a skill's validated range would otherwise pass every planned test.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 25 addressed (commit 12f7982) — four Codex findings against 3e3de0b:

1. **ADR-058:** Stopped claiming `.agents/skills` is read by "every P0 target agent" — the ADR's own ecosystem section already documents Claude Code as the exception. Qualified the authoritative-target claim to the clients that actually honor it (Copilot, Codex, Cursor, Gemini CLI).
2. **ADR-058:** Fixed the "packaging targets" bullet two ways: `.claude/skills/` is case (b) — a required generated copy Claude Code depends on — not case (a) — "unnecessary because the vendor already reads `.agents/skills`"; and removed the "copy/symlink" wording, which contradicted the immediately preceding no-symlinks invariant and P0.3's copy-only generator.
3. **G36 P0.4:** Split the Files list by which discovery design the maintainer selects. The previous unconditional list directed both the `info`-root-command path and the `--version --format json` path to touch `tests/test_cli_root_surface.py` and `README.md`'s command table — only correct for the former; the latter has no root-surface change at all.
4. **G36 P1.1:** Fixed "five-point rubric" to six — the listed grading dimensions (workflow, uncertainty, evidence, root cause, remediation, unsupported-claims) are six independently graded criteria, not five.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 26 addressed (commit c868ea0) — one Codex finding against 12f7982:

- **G36 P0.6:** Fixed — "register that marker pattern in `GENERATED_FILE_MARKERS`" was wrong: that constant is a flat list of exact `(path, marker, generator)` tuples with no glob/tree capability, which is exactly why `docs/reference/examples/case*.md` already needs a *second*, hardcoded glob loop in `check_generated_file_ownership()` rather than per-file registry entries. P0.6 now specifies a third glob loop over `.agents/skills/**/*.md` and `.claude/skills/**/*.md`, mirroring the existing examples-glob loop, instead of a registry entry that couldn't have enumerated either tree.

The other finding in this batch — adding YAML front matter to ADR-058 itself — was already raised earlier in this review and explicitly declined at the time: only 1 of 57 ADRs in this repo carries front matter today; the established convention here is the `**Status:**` line instead (`docs/AGENTS.md`'s front-matter rollout is scoped to `topics.yaml`-registered pages, which ADRs are not). Leaving as-is, consistent with that earlier decision.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 27 addressed (commit 5a83ef3) — one Codex finding against c868ea0, verified against the real file:

- **G36 P0.4:** `abicheck/cli.py` is currently 1,980 of the AI-readiness gate's 2,000-line hard cap (confirmed via `wc -l`) — only ~20 lines of headroom, while `info`'s own body (Click registration, `--format` handling, schema-version discovery, provider/platform probing, JSON/text rendering) needs more than that. Path (a)'s Files entry now calls out that landing `info` there needs a same-PR extraction shrinking `cli.py` first, so the command doesn't itself trip the hard cap.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 28 addressed (commit b9648c6) — two Codex findings against 5a83ef3:

1. **G36 P0.4 path (b):** Fixed the proposed `--version --format json` design — a naive eager `--version` callback runs before `--format`'s non-eager parsing populates `ctx.params`, regardless of argv order, per Click's documented callback evaluation order, so the naive shape couldn't have worked. Specified two correct alternatives (both flags eager with version's callback ordered last, or a non-eager version callback) and required testing the actual combined spelling rather than each flag in isolation.
2. **G36 P0.7:** Added P0.5 as a hard blocker for this item's report-JSON-field-path drift checks. Landing P0.7 before P0.5 would let it validate only `compare_report.schema.json`'s existing fields while silently passing on any skill-cited field path from the scan/release/release-summary/stack surfaces P0.5 defines schemas for — a coverage gap the item's Change section didn't disclose.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 29 addressed (commit 70a6ae8) — 7 findings from Codex and CodeRabbit against b9648c6:

1. **ADR-058 (Codex, CodeRabbit — same underlying gap, two reports):** Aligned the publication-target contract with what P0.3 actually generates. `.github/skills/` and `.cursor/skills/` are not generated — Copilot and Cursor already read `.agents/skills/` directly, so a generated copy would be redundant. `.claude/skills/` remains the one additional generated tree. Removed the earlier claim that generation "produces `.github/skills`/`.cursor/skills`" too.
2. **ADR-058 (CodeRabbit nitpick):** De-duplicated the "Canonical portable default" and "Source-of-truth model" sections — the latter now links back to the former instead of restating the same per-client discovery facts.
3. **ADR-058 (CodeRabbit major):** Separated `skills.sh` from the generated-filesystem-tree rule — it's a distinct external submission (P1.4, its own manifest, possibly a subset bundle), not a third output of `scripts/gen_agent_skills.py`.
4. **G36 P0.1 (CodeRabbit nitpick):** Fixed "copy ... verbatim" — reworded so the same step's link-rewriting follow-on (step 3) reads as a required part of the copy, not an unrelated later pass an implementation could skip.
5. **G36 P0.4 (Codex, second round):** Fixed the `--version --format json` Click design again — round 28's "make both flags eager" doesn't actually work either, since same-eagerness-group params are still processed in command-line order. The correct fix is mixed eagerness (`--format` eager, `--version` non-eager), which Click always processes as a whole group ahead of the non-eager group regardless of argv order.
6. **G36 P0.3 (CodeRabbit major):** Defined the Markdown-to-MkDocs URL mapping explicitly (mkdocs's default `use_directory_urls: true` behavior — `.md` dropped, trailing slash added, anchors pass through unchanged) instead of leaving it implied by "read `site_url`."

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 30 addressed (commit 76d7baf) — one Codex finding against 70a6ae8, verified against the actual dependency chain:

- **G36 P1.4:** Confirmed P1.1's own `Dependencies` line names only P0.1–P0.3/P0.8, so P1.4's precondition of "P1.1, and P1.5" didn't transitively reach P0.4 (capability discovery), P0.5 (typed comparability reasons), P0.6 (AI-readiness coverage of the generated tree), P0.7 (structural/drift gates), or P0.9 (including its dogfooding pass, which P0.9 itself already states must happen "before any external publication"). P1.4's `Dependencies` now names the full required P0 release explicitly as its own precondition, not something inherited transitively through P1.1.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 31 addressed (commit 2d4fb1f) — two Codex findings against 76d7baf:

1. **ADR-058:** Fixed the architecture diagram — it assigned generation to `skills-src/_generator/`, a subtree P0.3 never creates or assigns files to (the actual sole generator is `scripts/gen_agent_skills.py`, deliberately kept outside `skills-src/` since it's Layer C tooling, not editable skill source). The diagram now names the real generator and both its real output trees.
2. **G36 P0.1:** Widened the published-doc-URL rewrite scope. The prior wording only covered copied `shared/*.md` fragments, but a skill's own `SKILL.md` or skill-specific `references/*.md` file can link directly to a canonical `docs/` page too — P1.6's CI-wiring links are exactly this case — so leaving those unrewritten would reproduce the same standalone-install dangling-link gap the fragment-scoped fix was meant to close. The rewrite now runs over every staged Markdown file the generator produces.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 32 addressed (commit 7e4362d) — two Codex findings against 2d4fb1f:

1. **G36 P1.4:** Required P1.5 completion on all four of ADR-058's named minimum targets (Claude Code, Codex, Copilot, Gemini CLI) before publication, not just Codex and Copilot — ADR-058 itself lists only Cursor as conditional ("Gemini CLI at minimum, Cursor if current"); the prior wording incorrectly grouped Gemini CLI with Cursor as optional.
2. **G36 P1.4/P1.5:** Closed an artifact-freshness gap — P1.5 could complete right after P0.3, before P0.4/P0.5's required follow-up commits update skill content to consume capability discovery and typed comparability reasons, so P1.4 checking only "is P1.5 complete" could publish a generated tree different from the one actually validated. P1.5's `Dependencies` now require its publication-relied-on pass to postdate those follow-up commits, not just P0.1–P0.3.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 33 addressed (commit bfcf616) — three Codex findings against 7e4362d:

1. **G36 P0.7:** Extended the self-containment/link-resolution assertion set to run over both committed output trees (`.agents/skills/**` and `.claude/skills/**`), not only the former. A broken relative link or unrewritten doc reference introduced only by the Claude-specific emission path would otherwise pass every planned test indefinitely — drift tests re-check the same faulty output, and idempotent regeneration doesn't imply correct output.
2. **G36 P0.7:** Fixed the version-range enforcement to check containment in both directions, not just "exceeds." A skill declaring a minimum version because it depends on a feature introduced in that release would otherwise silently pass against an older installation that predates the feature entirely.
3. **G36 P0.4:** Narrowed the capability-discovery problem statement — the described payload is a command/schema-family *presence* inventory, not a per-command *option* inventory, so it can't actually answer "is `compare --contract` available" the way the original wording implied. Stated the real scope and named the gap explicitly rather than overclaiming payload coverage.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 34 addressed (commit 7de117c) — two Codex findings against bfcf616:

1. **ADR-058 criterion 2:** Pointed at the concrete artifact "the companion plan's evaluation matrix" refers to (G36 P1.1's `run_skill_evals.py` case selection plus `skill_eval_scenarios.yaml`) instead of an undefined matrix this criterion couldn't be objectively checked against.
2. **ADR-058 criterion 5:** Now requires all four of the Testing section's named minimum cross-agent targets (Claude Code, Codex, Copilot, Gemini CLI) running every P0 skill — the prior wording let the ADR be declared validated without ever exercising Claude Code or Gemini CLI, or three of the four skills, contradicting both the Testing section's own cross-agent list and G36 P1.5's per-target/per-skill procedure. Now matches the same full-coverage bar G36 P1.4's publication gate (round 32) already depends on.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 35 addressed (commit 0b9faa3) — two Codex findings against 7de117c:

1. **G36 P1.1:** Applied the same freshness caveat already added to P1.5 — a P1.1 run completed before P0.4/P0.5's required follow-up commits land never exercised the capability-discovery or non-comparability branches those commits add, so it can't stand in as evidence those branches actually work. The publication-relied-on P1.1 pass must now postdate them.
2. **G36 P1.4:** Added P1.6 (CI integration flow) as an explicit publication prerequisite. ADR-058's own admission-criteria decision states CI setup "ships as a documented action" inside `native-binary-compatibility-review`/`native-release-compatibility`'s first release, not an optional later addition — publishing before P1.6 lands would ship those two skills without content ADR-058 already commits them to carrying from day one.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 36 addressed (commit 96ed9c0) — one Codex finding against 0b9faa3:

- **G36 P1.1/P1.5:** Making P1.6 a P1.4 publication prerequisite (round 35) didn't close the loop by itself — P1.6's own `Dependencies` is only P0.2, so it can still land after P1.1/P1.5 complete, meaning publication could ship `SKILL.md` content (the CI-wiring step and its doc links) that the behavioral evaluation and cross-agent validation passes never exercised. Both P1.1's and P1.5's freshness requirements now also name P1.6's CI-wiring commit as something the publication-relied-on pass must postdate, alongside P0.4/P0.5's follow-ups.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 37 addressed (commit 46433a6) — three Codex findings against 96ed9c0:

1. **ADR-058:** Fixed "Four layers" to "Five layers" — the numbered list right below it (structural, tool/API drift, trigger, behavioral, cross-agent) has five items, and since this section serves as the ADR-level acceptance summary, the mismatch made it ambiguous whether cross-agent validation belongs to the required architecture.
2. **G36 P0.1:** Fixed "structural tests land in P0.5" to P0.7 — P0.5 is the comparability-reason-code product change and defines no skill-structural tests of its own; the structural/drift suite is P0.7's.
3. **G36 P0.4:** Split the Tests section by the same path (a)/(b) discovery-design choice the Files list already splits by. The prior unconditional `tests/test_cli_info.py` entry would test a nonexistent surface under path (b), or worse, prompt silently reintroducing the `info` command path (b) exists to avoid.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 38 addressed (commit 7d5a95e) — two Codex findings against 46433a6:

1. **G36 P0.4:** Applied the same fact-owner discipline already required for schema versions to the root command list. It was left unspecified, which would let an implementation return a stale hand-maintained list while the stated capability-discovery tests stayed green. Now requires click-introspection derivation (same mechanism `gen_cli_reference.py`/P0.7 already use) with an exact-equality test.
2. **ADR-058:** Resolved the safety-invariants two-copies tension explicitly. This ADR's own eleven-item list is frozen decision-record text once accepted, while `skills-src/shared/safety-invariants.md` is the living operational copy a future correction actually lands in — added a note stating the fragment is authoritative for current behavior if the two ever disagree, so a reader following the accepted ADR isn't left applying stale guidance with nothing telling them which copy currently governs.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 39 addressed (commit d0afff2) — two Codex findings against 7d5a95e:

1. **G36 P0.4:** Fixed the `--version --format json` Click design a third time. The mixed-eagerness fix (round 29) solves callback ordering but has a separate, worse problem confirmed against the live CLI: a real root-level `--format` option shadows every subcommand's own `--format` (`abicheck --format json compare OLD NEW` would silently consume the root value while `compare`'s own `--format` stayed at its text default — a correctness regression for every other command). Replaced with a design that registers no second global option at all: `--version`'s own callback manually checks for a `--format` token rather than declaring a Click-level root parameter, plus a required non-version regression test proving no new root option was introduced.
2. **G36 P1.1:** Split the six-dimension evaluation rubric's gating model. Two dimensions (preserved uncertainty, no compatibility claim without sufficient evidence) are this ADR's own safety invariants translated into grading criteria — the `SURVIVOR_BASELINE`-style "establish once, gate on non-regression" model is wrong for these, since a first-run failure would become the accepted floor and P1.4's "acceptable baseline rate" check would never independently catch it. These two now gate at zero tolerance from the first run; the remaining four keep the baseline/non-regression model.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 40 addressed (commit 8d125d1) — one Codex finding against d0afff2, confirmed by them against real Click 8.3.3:

- **G36 P0.4:** Round 39's "parse `--format` out of `ctx.args`/`sys.argv` without declaring the option" design doesn't work at all — Click's parser rejects any undeclared option ("No such option: --format") during argument parsing itself, before any callback runs, for both flag orders. Replaced with a design that actually works within Click's real constraints: declare `--format` as a normal eager root-level option, and enforce the "only valid with `--version`" contract inside `--version`'s own non-eager callback (guaranteed to run after `--format`'s value is populated, per the mixed-eagerness ordering that's still correct). `--version`'s callback now raises a `UsageError` when `--format` was given without `--version`, closing the original shadowing gap with a loud, clear failure instead of a workaround that doesn't parse at all.

This CLI design question has now gone through four review rounds (28, 29, 39, 40) — each catching a real, distinct Click semantics issue the prior fix missed. I'm fairly confident in the current design (declared eager `--format` + enforcement in `--version`'s non-eager callback), but flagging the history here in case a reviewer wants to double check before this lands as real code.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 41 addressed (commit c44b371) — 4 CodeRabbit findings against d0afff2, one verified against Gemini CLI's real documentation before fixing:

1. **ADR-058 (Major):** Fixed a real factual error — confirmed against Gemini CLI's own docs (`docs/reference/tools.md`, `activate_skill`: "Loads specialized procedural expertise from the `.gemini/skills` directory") that Gemini CLI reads skills from `.gemini/skills`, **not** `.agents/skills` as this ADR previously claimed. Claude Code and Gemini CLI are both documented exceptions needing a generated vendor-specific copy — not Claude Code alone. Updated the "Canonical portable default" bullet, the "packaging targets" bullet, and the source-of-truth diagram, which now shows all three generated trees (`.agents/skills/`, `.claude/skills/`, `.gemini/skills/`).
2. **ADR-058 (Minor, "complete the publication tree"):** The diagram was still missing `.claude/skills/`'s own output tree from an earlier round that only added the generator's directory entry — fixed as part of the same edit above.
3. **G36 (cascading from #1):** Extended every `.claude/skills/**` reference added in prior rounds (P0.1, P0.3, P0.6, P0.7) to also cover `.gemini/skills/**` — generator, marker/ownership check, and structural self-containment assertions now name all three trees consistently.
4. **G36 P1.4 (Major):** Clarified that P0.8 (trigger tests) is already a publication prerequisite, reached transitively through P1.1's own dependency list — not omitted, just not previously called out alongside the items P1.4 adds directly.

**One finding not applied, with reasoning:** CodeRabbit's request to fully replace ADR-058's own eleven safety invariants with a link-only summary. An ADR's decision section stating its actual decision is normal practice here (every other ADR in this repo does the same), and the two-copies tension already has an explicit resolution note from round 38 — the operational fragment is authoritative if the two ever disagree. Full replacement would remove the decision record's own content, not just deduplicate it.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 42 addressed (commit cf0993a) — three Codex findings against c44b371:

1. **G36 P0.6:** Fixed the marker-ownership fixture count from six to nine — round 41 extended coverage to three generated trees (`.agents/skills/`, `.claude/skills/`, `.gemini/skills/`), but the test prescription still said "plant in both trees... six is flagged," leaving one tree without negative marker coverage in the test plan.
2. **G36 P0.4/P1.1:** Generalized "updates `SKILL.md`s to use `info`" to consume whichever capability-discovery surface the maintainer decision actually selects. Path (b) has no `info` command at all, so a follow-up written unconditionally against path (a)'s spelling would reference a command that was never built.
3. **G36 P0.3:** Read `use_directory_urls` from `mkdocs.yml` itself (defaulting to `true` only when absent, mirroring mkdocs's own resolution) instead of hardcoding today's absence as a permanent assumption — closes the same "second un-synced fact owner" class of bug the `site_url` fix already closed, now for the URL-shape half of the mapping too.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 43 addressed (commit b7f010e) — one Codex finding against cf0993a:

- **G36 P0.7:** The version-range check only compared this repository's own dev/CI abicheck version against committed `SKILL.md` files — it never reaches a real user's own installation once a skill is later installed standalone from `skills.sh` alongside an out-of-range abicheck. Added the missing runtime half: each skill's own tool-selection step must independently read the installed `abicheck_version` via the capability-discovery surface and stop if it falls outside the skill's declared range, closing the gap between "CI catches stale commits" and ADR-058's actual "fail loudly" requirement at runtime.

The ADR-058 front-matter finding recurred again this round — already declined twice with the same rationale (only 1 of 57 ADRs in this repo carries front matter; the convention is the `**Status:**` line instead). Not re-applying.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 44 addressed (commit c8e516f) — two Codex findings against b7f010e:

1. **G36 P0.9:** Applied the same freshness caveat already required for P1.1/P1.5 to the dogfooding pass. P0.9's `Dependencies` only named P0.1–P0.3, so a dogfooding pass could complete before P0.4/P0.5's required skill-content follow-ups and P1.6's CI-wiring commit change the generated skills — letting P1.4 accept a stale dogfooding record while publishing materially different content.
2. **G36 P0.7:** Widened the report-JSON field drift check beyond `report-interpretation.md`. P0.1's `root-cause-grouping.md` fragment names `root_causes`/`root_cause_count`, and P0.5 adds `reason_codes` references elsewhere — restricting the check to one fragment would leave those real field references unprotected by the promised drift gate.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

napetrov:
Round 45 addressed (commit 2f43923) — one Codex finding against c8e516f:

- **G36 P0.9/P1.1/P1.5:** P0.6 (which adds ownership markers/front matter across all three generated trees) was still missing from the freshness requirements — the second round in a row where a closed, itemized "commits that change generated content" list turned out to be missing something (P1.6 was missing in round 36, now P0.6). Rather than patch in P0.6 and risk a third missing item next round, generalized all three freshness notes to state the actual requirement directly: postdate *every* commit that changes the generated skill trees' content, with the currently-known set (P0.4, P0.5, P0.6, P1.6) given as examples, not an exhaustive closed list.

`scripts/check_ai_readiness.py`: 0 errors, 96 warnings (unchanged baseline).

---

PR #686, opened 2026-08-09
feat: ADR-058 native compatibility Agent Skills — P0 portfolio and generator

napetrov:
## Summary

Implements the core of [ADR-058](https://github.com/abicheck/abicheck/blob/main/docs/contribute/adr/058-native-compatibility-agent-skills.md) (G36 phases P0.1–P0.3, P0.6–P0.9): four portable Agent Skills that let a coding agent answer a native-compatibility question in the user's own words, backed deterministically by the abicheck CLI.

## Motivation

ADR-058 (already accepted, previously unimplemented) decided the architecture for publishing abicheck as a coding-agent Skill portfolio, organized around the user's real question ("will this break existing consumers?") rather than around abicheck's own command vocabulary. This PR builds the skill portfolio, the generator that publishes it self-contained into every target agent's read path, and the gates that keep it from silently drifting.

## Changes

- **`skills-src/`** — the one hand-authored source: four `SKILL.md` files (`native-binary-compatibility-review`, `native-api-evolution`, `native-release-compatibility`, `native-consumer-compatibility`) plus eleven shared Layer-B domain fragments (compatibility contracts, evidence/depth, baselines and comparability, public-surface scoping, compiler/build profiles, consumer scoping, policies/suppressions, report interpretation, root-cause grouping, remediation catalogue, safety invariants) and a CI-wiring fragment cited by the two skills ADR-058 names for it (folds in P1.6).
- **`scripts/gen_agent_skills.py`** — publishes `skills-src/` into three self-contained generated trees (`.agents/skills/`, `.claude/skills/`, `.gemini/skills/`): resolves each skill's cited shared fragments transitively, copies (never symlinks) them into the skill's own `references/shared/`, and rewrites every `docs/` link to the published site using `mkdocs.yml`'s own `site_url`/`use_directory_urls` rather than a hardcoded literal. Wired into `scripts/verify.py` as the `agent-skills-generated` step.
- **AI-readiness gate**: `skills-src/` registered in `REQUIRED_CLAUDE_MD_DIRS`; `generated-file-ownership` now walks all three generated skill trees for the "do not hand-edit" marker, scoped to the directories the generator actually owns so a repository's own hand-authored skill sharing an output root is never flagged.
- **docs-contract gate**: `task_pages`/`allowed_summaries` now accept a repo-relative path outside `docs/` the same way `fact_sources` already does, and `_check_front_matter_schema` actually scans those registered external pages — so a `skills-src/shared/*.md` fragment's `summarizes` claim is enforced, not just recorded. Eleven fragments registered in `topics.yaml`.
- **Tests**: `tests/test_gen_agent_skills.py` (generator mechanics — fragment resolution, transitivity, orphan/missing-fragment detection, link rewriting, idempotency), `tests/test_agent_skills_structural.py` (frontmatter validity, self-containment, abicheck version-range containment in both directions), `tests/test_agent_skills_drift.py` (every cited CLI command/option and report-JSON field checked against the live Click tree and report schema), `tests/test_agent_skills_triggers.py` (a labelled trigger corpus extracted from and verified against ADR-058's own Product-positioning prompts, plus a lexical discriminability check against five out-of-scope negative prompts).
- **Docs**: new `docs/use/agent-skills.md` catalog page (nav + README pointer); ADR-058 and its index row updated to "partially implemented"; G36 plan items marked done as completed.

## Explicitly out of scope for this PR

P0.4 (`abicheck info`) is blocked on an explicit maintainer decision between two design paths G36 records (a new root command vs. extending `--version --format json`) — not attempted here. P0.5 (typed comparability `reason.codes`) is its own independent product-surface change per G36's sequencing. All of P1 (behavioral evaluation against the examples corpus, cross-agent validation log, public publication channels) remains.

## Checklist

- [x] Tests added/updated for new behaviour (4 new test files, ~85 new test cases)
- [x] Changelog fragment added (`changelog.d/20260809_000000_claude_adr058_native_agent_skills.md`)
- [x] Docs updated (`docs/use/agent-skills.md`, nav, README, `AGENTS.md`, `docs/AGENTS.md`, `scripts/CLAUDE.md`)
- [x] `scripts/check_ai_readiness.py`, `scripts/check_docs_contract.py`, `mkdocs build --strict`, `mypy abicheck/`, `ruff check` all pass with 0 errors
- [x] Full fast unit lane (23,847 tests) passes with 0 failures

---
_Generated by [Claude Code](https://claude.ai/code/session_01SzHf9vfN3MMUnzknWHwWaX)_


napetrov:
Fixed in 45956ec — all four Codex findings accepted, each verified against the live CLI or implementation rather than taken on faith, plus four found in a self-review of the generator.

**Codex findings**

- **`aggregate` missing `REPORTS_DIR`** — confirmed: it's a required `click.Argument`, so the example would have exited 64. Fixed to `abicheck aggregate release-reports/ --format json -o aggregate.json`, with `--manifest` kept as the way to surface a target that produced no report at all.
- **`project plan` at a pre-build stage** — confirmed against `--help`: a cell resolves only for a profile given a `--build-output`, and zero resolved checks exits 1. Step 2 now stops at `project validate`; `project plan` moved after the builds exist, with the `--build-output PROFILE=DIR` form shown. A profile with no build output is now explicitly matrix state **not run**.
- **Added overloads are not unconditionally source-compatible** — correct, and it mattered because this skill explicitly supports preserving both contracts. The table now splits "new function (new name)" from "overload of an existing name", and both it and the shared remediation catalogue name the real mechanisms (an existing call reaching the old function through an implicit conversion, an existing `&f`) and recommend a distinct name when source compatibility is in the contract.
- **Per-consumer attribution under multiple `--used-by`** — confirmed in `cli_compare_helpers._apply_used_by_scoping`: the per-app summaries are per app, but `relevant_changes_by_id` is a deduplicated union keyed by finding id, with no app association. The skill previously said "report per consumer" while pointing at the merged finding list — which would misattribute one app's break to another, the exact provenance failure its own safety invariant 11 forbids. It now scopes one run per consumer whenever findings must be named per consumer, and limits the multi-app run to the coarser "which of these is affected at all".

That last one also exposed a gap in the drift gate: it checked command and option *names* but not required positional arguments, which is why the `aggregate` defect got through. It now validates operand counts too, asking Click which options consume a following word so an option's value isn't miscounted as an operand, and scoped to runnable invocations so prose that merely names a command isn't flagged.

**Also fixed (self-review)**

- Stale generated skill directories were never detected or removed — renaming a skill left the old one published in all three trees while `--check` reported "in sync". Ownership is now read off the emitted marker rather than the current render; hand-authored skills sharing an output root carry no marker and stay untouched.
- `agent-skills-generated` was in the `pr` profile but in no CI job's `--only` list, so it never ran (and `scripts/CLAUDE.md` claimed it did). Wired in, and a new structural test asserts every `pr` step is CI-reachable. That test records one pre-existing gap it surfaced: `fmt-check` runs in `--profile pr` and pre-commit but in no CI job — `ruff format --check abicheck/ tests/` reports 530 files on a clean checkout of `main` too, so it's untouched here rather than silently fixed.
- `render_all(src_dir)` honoured `src_dir` for discovery but resolved fragments against the module globals, mixing an alternate tree with this repo's own.
- A missing `mkdocs.yml` `site_url` silently produced root-relative links that resolve nowhere from an installed skill; now a hard error.

Full fast lane green (23875 passed), `ruff check` clean, ai-readiness and docs-contract at 0 errors.

---

napetrov:
Fixed in 1bddf2b — all four accepted; each checked against the real loader/CLI semantics rather than restated.

- **Mach-O `compatibility_version`** — correct, and my table had it backwards. dyld records in each client the compat version of the dylib it linked against and refuses one whose version is *lower*, so the check guards downgrades; raising it does not turn away older clients. As written it would have led someone to ship an incompatible dylib under an unchanged install name believing the bump protected them. The install name is now stated as the macOS ABI epoch, with compat-version bumps reserved for backward-compatible evolution.
- **Baseline named by the consumer** — right, and the row was self-contradictory: the question names 1.2 while the answer said "oldest supported". A symbol introduced in 1.2 and removed by 2.0 is invisible against 1.0. Now the baseline is the version consumers were built against, with a range requiring one comparison per supported release (consistent with the matrix dimension added to the release skill last round).
- **`--diagnostic-comparison` for profile-change analysis** — confirmed against the flag's own help and `docs/reference/exit-codes.md`: the pair is refused with exit 16, so the branch was unusable as written. It now directs to the escape hatch with the two obligations attached — carry `assurance: "none"` and the mismatch warning into the summary, and don't use it to get past a mismatch that was accidental rather than the subject.
- **pImpl stack allocation** — correct, I overstated it. `Session` keeps one `Impl*`, so it stays complete and pointer-sized; that stable size is the point of the pattern. What's actually lost is triviality of the special members, which is what the cost line now says.

Two follow-ons: exit `16` is now named wherever the skills branch on a not-comparable result (they treat it as a distinct outcome, so the code a user actually sees should be stated), and the drift gate correctly flagged `compatibility_version` as absent from the report schema — it's a Mach-O load-command field, so it went into a narrow `NON_REPORT_IDENTIFIERS` escape mirroring the existing `FOREIGN_LONG_OPTIONS`, asserted disjoint from real schema property names so an entry can never shadow a genuine field reference.

Gates green: fast lane 23404 passed, `ruff check`, ai-readiness and docs-contract at 0 errors, `mkdocs build --strict` clean, generated trees in sync.

---

napetrov:
Fixed in 85a475d — all three accepted.

- **`COMPATIBLE_WITH_RISK` in the release matrix** — the sharpest of the three. It was unrepresentable: four states covered pass/break/not-comparable/not-run, so a risk verdict landed in "pass", and because it exits `0` exactly like `COMPATIBLE` (confirmed in `docs/reference/exit-codes.md` and `learn/verdicts.md`), an exit-code reading of the matrix couldn't see it either. A release that raised a dependency floor could get a patch/minor with the deployment risk never surfaced. Now five states keyed off the report's own `verdict`, with a **risk** cell making the recommendation conditional on reviewing those findings, and the termination criteria requiring they be surfaced.

- **Defaulted parameter** — right, my claim was over-broad. It holds only for ordinary calls that omit the argument; taking the function's address, using the old function-pointer type, or overriding it as a virtual all fail to compile. Both the design table and the remediation catalogue now qualify it and recommend keeping the old signature alongside a new entry point when source compatibility is in the contract.

- **Appended enumerator** — correct, and it's the subtler one: without a fixed underlying type the implementation picks one wide enough for the enumerators it can see, so a new out-of-range value can widen it and change `sizeof` plus every containing layout. The design table, remediation catalogue, and semver table now gate "minor" on a fixed underlying type or an in-range value, and the reserved-slots pattern gained the from-day-one advice to pin the type — it's the same class of decision as reserving fields, and equally free if made early.

Gates green: 23404 fast tests, `ruff check`, ai-readiness and docs-contract 0 errors, `mkdocs build --strict`, generated trees in sync.

---

napetrov:
Fixed in fbb2a77 — all three accepted, and the version-range one is the most consequential finding of this whole review cycle.

**Version range.** Confirmed and worse than "predates": the shipped 0.5.0 CHANGELOG section mentions none of the five surfaces, and each appears only in unreleased `changelog.d/` fragments — `aggregate` (29), `--contract-evaluation` (17), `--report-mode root-cause` (9), `project plan` (7), `--diagnostic-comparison` (5). A user on the published release would pass the mandatory preflight and then fail on unknown commands: exactly the failure the range exists to prevent. Retargeted to `>=0.6.0,<0.7.0`.

Worth recording *why* it was wrong, because one of my own tests caused it. `test_installed_abicheck_is_inside_each_declared_range` asserted the installed version fell inside the range — but in an unreleased tree `importlib.metadata` reports the last *published* version, with no dev marker to distinguish it. Satisfying that assertion forced the minimum down onto the release lacking the surface. The test looked like a guard and was in fact producing the defect. It's replaced by one asserting the minimum is newer than `repo_facts.json`'s `latest_release` while unreleased fragments are pending (and equals-or-exceeds it after), verified in both directions.

**`aggregate` expected-target mode.** Right, and my previous round's fix was insufficient — I added the required `REPORTS_DIR` but the example still exited 64. Now uses `--manifest`, with the reasoning that an undeclared target should be reported missing rather than silently absent (step 4's "not run" state), and why `--discovered-only` is the wrong default for a release decision. Since that requirement is enforced in the command body rather than the Click parser, neither the option-name nor the positional-argument check could see it — added a test that executes each cited `aggregate` invocation against a real empty reports directory, scoped to `aggregate` as the one command here gating on an option *group*.

**Severity resolution.** Confirmed by running `compare` with no severity flag against a `.abicheck.yml` that supplies one — the block is present. My "only when a `--severity-*` flag was passed" would have made an agent report "there is no gate" in a repo that has one. Now described as present whenever severity-aware grading is resolved from any source (flag, config, profile, or pack).

Also scoped the drift gate to the files the generator actually publishes — `skills-src/CLAUDE.md` is the contributor contract, never shipped into a skill, and discusses repo internals that aren't report fields.

Gates green: 23400 fast tests, `ruff check`, ai-readiness and docs-contract 0 errors, strict mkdocs, trees in sync.

One note on cadence: this is the fifth push in ~45 minutes, and each has cancelled the previous CI run via the workflow's concurrency group, so no run has completed yet. I'll hold off on further pushes unless something blocking appears, to let this one finish.

---

napetrov:
Fixed in 6819022 — all three accepted, and the first is the worst inversion found so far.

- **Scoped verdict read backwards.** Confirmed twice over: `compare_report.schema.json`'s own `full_verdict` description says "verdict becomes the scoped, gate-relevant one; full_verdict preserves the original", and `tests/test_cov95_cli.py:1693-1695` asserts a `--used-by` run with `full_verdict: BREAKING` / `verdict: COMPATIBLE`. The skill called `verdict` "the library's global answer — context, not the answer here", which would have led an agent to discard the per-consumer answer and report the global one. In the skill whose entire purpose is the per-consumer answer, that's about as wrong as it gets. Corrected in both the skill and the shared fragment, including that `full_verdict`'s *presence* is itself the divergence signal.

- **Runtime floor.** Right — `runtime_floor_raised` is a real `ChangeKind`, emitted as a risk and promoted to a break against declared floors by `apply_runtime_floor_contract`. "Runtime isn't `compare`'s job" would have had an agent ignore a finding sitting in the report it was already reading. Now separates the symbol-version floor `compare` covers from the dependency-graph checks `deps compare`/`deps tree` add.

- **Gate-failing cells marked pass.** Correct, and it is the same shape as the `COMPATIBLE_WITH_RISK` gap from the previous round — I fixed the verdict axis and left the other two. `severity.blocking` and `contract_coverage_exit_contribution` fail independently on the same report and each exit nonzero. Pass now requires all three axes clear; a cell failing either of the others becomes **risk** with the same conditional handling.

That last one is worth naming as a pattern rather than a one-off: three separate findings now have been "the matrix/summary keys off one axis while another can fail independently". The orthogonal-axis design is easy to state and easy to under-apply.

Gate gap closed alongside: the drift check rejected `runtime_floor_raised` as not-a-report-field. Rather than allowlist it, `ChangeKind` values are now a recognized vocabulary checked against the live enum — so change-kind citations are validated too, and a renamed kind fails here. Verified by renaming a cited kind and confirming the failure.

Gates green: 23400 fast tests, `ruff check`, ai-readiness and docs-contract 0 errors, strict mkdocs, trees in sync.

---

napetrov:
Fixed in e6d14b1 — all four accepted, plus one more the sweep turned up.

Three of the four were **leftover instances of corrections I had already made** — primary statement fixed, sibling left saying the opposite. Rather than patch them individually again, I swept every previously-corrected claim across `skills-src/`, which found a fourth instance the review hadn't flagged (`native-binary-compatibility-review`'s step 4 still had the stale severity claim).

- **`severity` conditional-field table** — right, and worse than a leftover: the page contradicted itself, accurate prose at lines 57–68 and the stale rule in the table below it. Both that row and the swept sibling now describe resolution from flag, `.abicheck.yml`, run profile, or gate pack.
- **Remediation catalogue runtime-floor row** — correct, and it's the row that matters most: an agent looking up a remediation lands there, not on the fragment I fixed. Now states the finding is visible as `runtime_floor_raised`, with `deps compare` scoped to the wider dependency graph.
- **`coverage_warnings` "always present"** — correct, and instructive about my own method. That claim came from over-generalizing a single synthetic probe that happened to have warnings; `reporter.py` emits the key only when non-empty. I re-checked `evidence_tier`/`evidence_tiers` the same way while there — those two really are unconditional, so those claims stand.
- **SONAME for `API_BREAK`** — correct, and it contradicted this file's own later source-only exception. Already-compiled consumers cannot observe an `API_BREAK`, so prescribing an epoch bump imposes a needless ecosystem-wide rebuild. `BREAKING` and `API_BREAK` are now split, with the epoch requirement attached only to the former.

Two things worth recording about the review pattern, since they affect what to trust going forward:

1. **Empirical verification against one input is not verification.** The `coverage_warnings` error came from exactly that, and it's the same shape as an earlier one where a single `compare` run misled me about field presence. Reading the emission condition in `reporter.py` is what settles it.
2. **Fix propagation is its own failure mode.** Three findings this round existed only because I corrected the canonical statement and not its siblings — which is ironic in a portfolio built on "one fact, one place". The shared fragments reduce this but don't eliminate it, because the same fact legitimately appears as prose in one place and a table row in another.

Gates green: 23400 fast tests, `ruff check`, ai-readiness and docs-contract 0 errors, strict mkdocs, trees in sync.

---

PR #700, opened 2026-08-10
feat: honor severity/exit-code-scheme in `scan --against`

napetrov:
## Summary

`scan --against` now honors `--severity-preset`/`--severity-*`/`--exit-code-scheme` (and `.abicheck.yml`'s `severity:`/`exit_code_scheme` keys) the same way `compare` already does, closing an asymmetry between the two commands' exit-code contracts.

> **Note on scope.** The core change is small. Most of this PR is the consequences of *widening a command's exit-code space*, found across seven automated review rounds — each fix exposing the next boundary. **Two of those were regressions this PR itself introduced, both of the "CI goes green on a real failure" kind.** They are called out below; a reviewer short on time should look at those first.

## Motivation

`scan --against` computed its exit code from the overall verdict alone (`legacy_exit_code`: `BREAKING`→4, `API_BREAK`→2, else→0), completely ignoring any severity setting — unlike `compare`, where `--severity-preset info-only` can leave a `BREAKING` verdict at exit `0`. The severity CLI/config surface was already partially plumbed into `scan` (`resolve_compare_config` was called with the severity fields hardcoded to `None` and the result discarded), so this closes a real, previously-documented gap rather than adding machinery from scratch.

## Changes

**The feature**
- Added `--severity-preset` (+ the hidden per-category `--severity-*` overrides) and `--exit-code-scheme` to `scan`, gated as comparison-only flags — a usage error (exit `64`) without `--against`, matching the rest of that family.
- Threaded the resolved `SeverityConfig`/scheme through `run_scan_core` → `_run_baseline_compare`, which computes the exit via `severity.compute_exit_code` under the `severity` scheme. The default (`legacy`) scheme is byte-for-byte the prior behavior.
- `--budget` overflow / `NOT_COMPARABLE` / evidence-contract exit codes are unaffected — they're returned before the comparison runs.

**Making the new exit code explicable** (a non-zero exit on an otherwise *compatible* diff was indistinguishable from ADR-049 §7's orthogonal contract-coverage `1`)
- The report now publishes the gate that produced the code: a `diff.severity` block in JSON (`scan_schema_version` 1.9, built by the same `reporter._build_severity_json` that writes `compare`'s, so the two are comparable field-by-field) and a `severity gate: exit N — blocking: <cats>` line in the **default text** output.
- The gate is the scan's *compatibility* gate: a `--crosscheck KEY=error` promotion raises it too (adding `promoted_crosscheck`, a floor — it never lowers the gate).
- The ADR-049 coverage diagnostic now explains itself against the real pre-coverage base rather than re-deriving a verdict-based one.

**Regressions this PR introduced, and fixed** ⚠️
- `aggregate.GateInfo.from_scan_report` separated the coverage axis from the compatibility gate using an explicit premise in its own comment — *"scan … has no native 1, so a raw 1 can only be the orthogonal contribution."* Giving scan a severity gate falsified it: an error-level addition **is** a native `1`. Folded with a coverage `1`, a genuinely-failing target was attributed entirely to coverage and **silently dropped from `blocking_targets`**. Fixed by reading the nested gate through `GateInfo.from_report_data` — the same strict, fail-closed reader `compare` reports use.
- Making that block authoritative then meant `check_report._neutralize_gate` (`gate-mode: advisory`) had to zero it too, or an **explicitly advisory check would block**. Fixed following the two rules that function already documents from hitting them before: the traversal is *imported* (new `aggregate.scan_severity_gate_paths`, sibling of `contract_coverage_block_paths`) rather than re-derived, and each container is rebound to a copy before writing.

**Correctness and docs**
- Fixed a verdict-relabel bug: a severity-demoted `BREAKING` diff could be reported as the *less* severe `API_BREAK` by a promoted cross-check.
- `docs/reference/exit-codes.md`: new "`scan --against` and severity" section; corrected the summary table's demoted-break rows — including a **pre-existing** error (compare-severity fixed `BREAKING` at `4`, but `info-only` yields `0`) that this PR made load-bearing by pointing readers at that column. Verified against the implementation, since the default preset isn't uniform (`default` leaves `potential_breaking` at `warning`, so `API_BREAK` → `0` until `strict`).
- Registered the new surface in `docs/_meta/topics.yaml`; regenerated `cli-reference.md`; corrected `scan --help` and `--pack`'s help, which still stated scan "follows its verdict directly and has no gate".
- `--pack` gate-severity folding is **not** extended to `scan` (still `gate_supported=False`); the rejection message now describes that narrower gap accurately. Likewise the ADR-049 contract-evaluation receipt still blanks its severity fields — it is a separate resolution from the one that scores the run, and unifying them needs `compare`'s combined `resolve_and_apply` machinery, not a drive-by field addition.

## Testing

- New coverage for: severity-gated `BREAKING` → `0` and default-preset → `4`; comparison-only-flag rejection; the JSON gate block and its absence under legacy; the text gate line and its absence under legacy; `TestScanPreCoverageBaseExit` (promoted/demoted bases, legacy fallback, malformed/absent summaries); the aggregate blocking/non-blocking/fail-closed cases; advisory neutralization *and* `deferred` preservation; the cross-check promotion in both directions.
- Full fast suite green on each commit (23342 → 23354 passed, 0 failures). `mypy abicheck/` clean, `ruff check` clean, `check_ai_readiness` 0 errors, `check_docs_contract` 0 errors, `mkdocs build --strict` clean.
- CI on this branch has repeatedly been auto-cancelled by supersession as each review round landed; pushing has now stopped so a full pass can complete.

## Checklist

- [x] Docs updated (`exit-codes.md`, `cli-reference.md` regenerated, `AGENTS.md`, `topics.yaml`)
- [ ] `pre-commit run --all-files` passes locally — not run: `ruff format --check` shows pre-existing drift across ~530 files in this environment. Confirmed against `origin/main` that every file this PR touches was already in that set (or is still clean), so the change introduces no new formatting deviation.
- [x] No new `mypy` or `ruff check` errors introduced


napetrov:
Re: the P1 on `abicheck/diff_types.py:1170` ("Do not treat the vtable-derived vptr as independent evidence") — confirmed and reverted in b581383. It was worse than the comment states, and the reasoning is worth recording.

Reproduced with producer-shaped records:

```python
def producer(vtable):  # exactly dwarf_snapshot.py:774 / dumper_castxml.py:1517
    return RecordType(name="A", kind="class", size_bits=64, vtable=list(vtable),
                      vptr_offset_bits=0 if vtable else None)

ev("A", producer([]), producer(["A::f()"]), {}, {})  # -> True, both directions
```

Because `(old.vptr is None) != (new.vptr is None)` is *identical* to the empty↔non-empty vtable transition on both backends, that branch was true by construction for every input reaching it. So it did not merely leave one false positive standing — it made the whole guard a **no-op**, restoring the original capture-gap FP in full.

Of the two remedies offered, neither was available as stated:

- *"Stop deriving the vptr witness from the vtable list"* — `diff_layout._check_vptr_introduced` keys on exactly that `None → 0` transition, so changing the producers would break a working detector in order to fix its consumer.
- *"Require a genuinely independent layout fact"* — there is none to require. Only the optional `ABICHECK_CLANG_LAYOUT_TOOL` path computes a real vptr, and nothing in the model distinguishes that value from the derived one.

So the field is simply not read here anymore. The pure-virtual case rejoins the polymorphic-through-a-base case as an accepted false negative. Neither hides the break: `_check_vptr_introduced` fires independently and the verdict stays `BREAKING` — verified end to end, and now pinned by its own test rather than asserted in prose, since the guard's docstring leans on that neighbour.

The test gap was the more important half of the finding, and you identified it exactly. The helper built `RecordType`s with `vptr_offset_bits` left `None` on both sides — a record no backend can emit — so the entire file, including the pre-existing test for the guard's whole purpose, passed against a guard that suppressed nothing on real input. The helper now derives the field the way the producers do, and one test pins that derivation in the producers' own source so this reasoning fails loudly if either ever computes a real vptr. Five tests in the file fail against the reverted revision.

Shadow-measurement `exports` baseline 31 → 30, attributed in place. Full suite 23641 passed / 0 failed; FP-rate gate 41 cases, 0 FP / 0 FN; ai-readiness 0 errors.

---

PR #693, opened 2026-08-09
refactor: cut CodeFactor "Complex Method" complexity in buildsource/ (14 of 68)

napetrov:
## Summary

First batch of the CodeFactor **Complexity** cleanup: **14 of the board's 68** `Complex Method` findings — every one in `abicheck/buildsource/`. Behaviour-preserving extraction only; no logic changes.

Rebased onto `main` after #688 was squash-merged, so this now shows only its own two commits (12 files).

## Motivation

Reading the actual CodeFactor board changed the picture from what #688 assumed. Its 69 issues were:

| Category | Count | Severity | Engine |
|---|---|---|---|
| Complexity — `Complex Method` | 68 | Minor | CodeFactor's own metric (threshold ~16) |
| Security — unsafe yaml load | 1 | **Major** | bandit |

CodeFactor does **not** report pylint for this repo. The single Major (`abicheck/impact/use_cases.py:284`) went in with #688. Everything remaining is complexity, spread across 60 files.

The metric is not reproducible locally — I fitted candidate formulas against all 59 Python findings and the best fit matched only 8 — so `mccabe` is used as a proxy: it identifies the **same functions with the same line ranges**, running 0–9 points below CodeFactor's number. Targets below are quoted as mccabe.

## Changes

| Function | mccabe | CodeFactor (before) |
|---|---|---|
| `type_graph._walk_types` | 26 → **11** | 35 |
| `type_graph._index_declared_entities` | 24 → **10** | 33 |
| `toolchain_probe._check_one_overlay` | 24 → **8** | 33 |
| `inputs_emit.compact_inputs_pack` | 26 → **10** | 25 |
| `inline.BuildConfig.from_dict` | 20 → **<9** | 16 |
| `inline.BuildConfig._validate_structure` | 18 → **<9** | 22 |
| `source_link.link_source_abi` | 18 → **<12** | 21 |
| `header_graph.build_header_only_graph` | 17 → **9** | 16 |
| `crosscheck_coherence._check_compile_context_conflict` | 16 → **<9** | 18 |
| `project_targets._target_issues` | 16 → **<12** | 17 |
| `l2_seed.derive_l2_include_dirs` | 15 → **<9** | 17 |
| `inputs_validate.validate_inputs_pack` | 14 → **<9** | 16 |
| `check_report.augment_report` | 14 → **<9** | 17 |
| `include_graph.depfile_args_from_argv` | 13 → **<9** | 17 |

Shapes worth calling out:

[...]

PR #701, opened 2026-08-10
refactor: split the SYCL AST document scanner and the C++ cv-qualifier scanner

napetrov:
## Summary

Batch 5b of the CodeFactor "Complex Method" series: the two hand-rolled character-level scanners, both behaviour-preserving and both verified differentially against their pre-refactor selves.

## Motivation

Continues #693 / #694 / #695 / #696 (merged) and #697 (open). These two are the highest-mccabe functions left on the board, and both are stateful parsers where the ordinary "read it and check the branches" review is weak — so each ships with an exhaustive differential check instead of an argument.

`name_classification._strip_cv_in_segment` in particular was **deferred earlier in this series** as too risky to touch. What changed is not the risk but the verification: the differential technique used on `_looks_like_itanium_encoding` in #697 applies directly, which turns the risk from assumed into measured.

## Changes

|---|---|---|---|
| `name_classification._strip_cv_in_segment` | 27 | 20 | <8 |
| `sycl_context._iter_json_documents` | 22 | 23 | <8 |

**`_iter_json_documents`** held the whole document-boundary scanner in one generator: a nested chunk-refill closure over three `nonlocal`s, the whitespace skip, the bracket/string/escape state machine and the piece collection, all sharing `cur`/`pos` directly. The state now has an owner — `_ChunkCursor` (`ensure_char`/`skip_whitespace`/`scan_document`) and `_BracketState`, fed one character at a time. The generator is left as what it always described: skip whitespace, check the document opens with an object/array, decide whether this one is wanted, scan, yield.

This also gives the two performance invariants a place next to the code implementing them rather than in one long docstring: why a growing buffer with `raw_decode` retries was quadratic (on `_ChunkCursor`), and why an unwanted document must not accumulate chunks even transiently (on `scan_document`).

**`_strip_cv_in_segment`** carried every syntactic case of C++ declarator parsing in one loop — template-argument skipping, array-decay pointer-boundary tracking, declarator-grouping vs. parameter-list parens, per-parameter comma scoping, sigil tracking and cv-token recording — over three variables shared with a nested closure. Each case is now a method on `_CvSegmentScan` reached through a leading-character dispatch table, so the rule for a case and the evidence for it (which real g++/clang mangling confirmed it, which review round found it) sit together instead of in one 130-line comment block. `_strip_cv_in_segment` stays as the entry point, so callers and the recursion are unchanged.

## Verification

Both were checked against their pre-refactor implementations, comparing outputs *and* raised exception types/messages:

- **`_iter_json_documents`** — 72,198 runs: 4,011 document streams (nested objects, brackets inside strings, escaped quotes and backslashes, malformed and truncated input) × chunk sizes 1/2/3/7/64/100000 × three `want_text` policies. **0 differences.**
- **`_strip_cv_qualifiers`** — 98,463 spellings: every combination of cv-qualifier × base type × pointer/reference/array suffix; function pointers across return type × declarator grouping × parameter list × trailing cv (including nested callbacks and cv-qualified callback return types); template arguments; pointer-to-array declarators; plus 120k random token-soup strings from the grammar's own alphabet. **0 differences.**
- **`cv_qualifiers_only_differ`** — the real public entry point — all **1,283,689** ordered pairs of the realistic subset. **0 differences.**

Plus the full fast suite (23334 passed), `ruff check`/`format`, `python -m mypy abicheck/` clean across 337 files, and `scripts/check_ai_readiness.py` at 0 errors.

## Checklist

- [x] Tests added / updated for new behaviour — n/a, refactor-only; the differential checks above plus the existing suites for each module are the regression check

---

### DIFF
--- a/abicheck/name_classification.py
+++ b/abicheck/name_classification.py
@@ -47,6 +47,8 @@ checks and is left as a follow-up.
 from __future__ import annotations
 
 import re
+from collections.abc import Callable
+from typing import ClassVar
 
 __all__ = [
     "ITANIUM_RTTI_PREFIXES",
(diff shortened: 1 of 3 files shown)


napetrov:
@coderabbitai review

---

napetrov:
@coderabbitai review

---

PR #703, opened 2026-08-10
refactor: split the aggregate text renderer and the ADR-049 contract evaluator

napetrov:
## Summary

Batch 5d of the CodeFactor "Complex Method" series: two findings, both pure extractions with no reordering.

## Motivation

Continues #693 / #694 / #695 / #696 / #697 / #701 (merged) and #702 (open). Touches no file the open PR does, so it lands independently.

## Changes

|---|---|---|---|
| `aggregate.AggregateResult.render_text` | 22 | 19 | <8 |
| `contract_evaluation.evaluate_change_contract_relevance` | 18 | 15 | <8 |

**`render_text`** rendered the whole aggregate text report — header, per-target lines, compatibility, contract coverage, the profile matrix, and the closing `Coverage:`/`Gate:` blocks — appending to one `lines` list throughout, so every section's branching counted against the same function. Each block that can be absent or take several shapes now returns its own lines: `_render_contract_coverage_lines`, `_render_profile_matrix_lines`, `_render_coverage_and_gate_lines`, with `_render_profile_entry_line` taking the four mutually exclusive per-target shapes and their two qualifying suffixes.

**`evaluate_change_contract_relevance`** is a chain of early-return authorities — kind, contract domain, decisions an earlier pipeline step already made, then a from-scratch header-surface classification. Each carries the review round that established it, so the guards had grown a long way from the code they guard. Three helpers, one per kind of authority:

- `_mode_dispatch_decision` — what the finding's kind and the selected `--contract` domain settle on their own;
- `_pipeline_authority_decision` — what `FilterNonPublicSurface` and friends already decided unconditionally, which a recomputation here cannot see and could contradict;
- `_surface_classification_decision` — the from-scratch classification, reached only once every earlier authority has declined.

The entry point is left as the order those authorities apply in.

## Verification

Both are extractions with no reordering, and both have strong direct coverage, so neither needed the differential treatment the hand-rolled parsers in #697/#701 got:

- `render_text`'s output is asserted directly in **27 places** across `tests/test_aggregate.py`, `tests/test_aggregate_findings.py` and `tests/test_cli_scan_helpers_coverage.py`.
- `evaluate_change_contract_relevance` is covered by **1,639** contract tests.

Plus the full fast suite (23340 passed), `ruff check`/`format`, `python -m mypy abicheck/` clean across 338 files, and `scripts/check_ai_readiness.py` at 0 errors.

## Checklist

- [x] Tests added / updated for new behaviour — n/a, refactor-only; the existing direct coverage above is the regression check

---

### DIFF
--- a/abicheck/aggregate.py
+++ b/abicheck/aggregate.py
@@ -953,19 +953,35 @@ class AggregateResult:
         lines.append("Compatibility:")
         lines.append("  " + self._render_compatibility_line())
 
+        lines.extend(self._render_contract_coverage_lines())
+        lines.extend(self._render_profile_matrix_lines())
+        lines.extend(render_finding_matrix_lines(self.finding_matrix))
+
+        lines.extend(self._render_coverage_and_gate_lines())
+
+        return "\n".join(lines)
+
+    def _render_contract_coverage_lines(self) -> list[str]:
+        """The ADR-049 contract-coverage block, empty when every domain closed.
+
+        Named separately from the required-target coverage line for the same
+        reason the JSON block is: a bare exit 1 with neither of them explained
+        is the failure this axis most easily causes.
+        """
+        out: list[str] = []
         # Named separately from the required-target coverage line above for
         # the same reason the JSON block is: a bare exit 1 with neither of
         # them explained is the failure this axis most easily causes.
         incomplete = self.contract_coverage_targets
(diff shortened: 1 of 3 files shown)


napetrov:
@coderabbitai review

---

PR #707, opened 2026-08-10
refactor: split the DWARF struct-layout detector and the stdlib spelling helper

napetrov:
## Summary

Batch 5e of the CodeFactor "Complex Method" series: two findings, both in dense, heavily review-hardened code, so both ship with a differential check rather than an argument.

## Motivation

Continues #693 / #694 / #695 / #696 / #697 / #701 / #702 (merged) and #703 (open). Touches no file the open PR does.

## Changes

|---|---|---|---|
| `diff_platform._diff_struct_layouts` | 17 | 15 | <8 |
| `type_reachability.directly_referenced_stdlib_type_spellings` | 17 | 15 | <8 |

**`_diff_struct_layouts`** held four numbered sections inside one per-struct loop: whole-struct size and alignment, the added-field-by-offset index, the removed-field matching (reserved-field activation → pure rename → genuine removal), and existing-field offset/type drift. Each is now its own function, so the five rounds of review fixes recorded in the removed-field matching comments sit next to the matching they constrain rather than in the middle of an unrelated loop.

`FieldInfo`/`StructLayout` move to a `TYPE_CHECKING` import — the runtime function-local import existed only to annotate locals, and the helpers need the names in signatures.

**`directly_referenced_stdlib_type_spellings`** ran four distinct passes in one body: resolving the exact set, resolving the trusted set, counting stripped-spelling collisions across every stdlib identity in the snapshot, and assembling the answer.

The exact and trusted passes were *structurally identical* — differing only in which scan collections they start from, never in the rule — so they now share `_alias_confirmed_identities`. The "one genuinely unambiguous route is real proof" rule is stated once instead of twice. `_stripped_spelling_index` and `_assemble_spellings` take the other two passes.

## Verification

Both checked differentially against their pre-refactor implementations, comparing full outputs:

- **`_diff_struct_layouts`** — 40,002 runs: randomized field sets over overlapping names (including reserved-prefixed ones), byte offsets, type spellings, byte sizes and bit offsets/widths, so the reserved-activation, rename and removal paths all fire; plus the struct-missing and empty-struct cases. **0 differences.**
- **`directly_referenced_stdlib_type_spellings`** — 12,000 runs: randomized snapshots over identities chosen to exercise every hazard the function's own comments document — a user `api::vector<int>` whose bare spelling collides with `std::vector<int>`, an inline-ABI-tagged `__cxx11::basic_string`, a global `duration` alongside `std::chrono::duration`, resolved and ambiguous typedef aliases, and qualified typedef keys whose *derived* suffix is what collides — across both `exclude_export_only_roots` values. **0 differences.**

The second check was also confirmed non-vacuous: **34% of sampled inputs returned non-empty results** (both the qualified identity and its derived bare form), so it compared real output rather than empty sets on both sides.

Plus the full fast suite (23468 passed), `ruff check`/`format`, `python -m mypy abicheck/` clean across 340 files, and `scripts/check_ai_readiness.py` at 0 errors.

## Checklist

- [x] Tests added / updated for new behaviour — n/a, refactor-only; the differential checks above plus the existing suites are the regression check

---

### DIFF
--- a/abicheck/diff_platform.py
+++ b/abicheck/diff_platform.py
@@ -17,7 +17,7 @@
 from __future__ import annotations
 
 import re
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from .binary_utils import strip_vendor_hash
 from .checker_policy import ChangeKind
@@ -71,6 +71,11 @@ from .model import (
     stdlib_namespaces_excluded,
 )
 from .name_classification import RTTI_DATA_PREFIXES
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    from collections.abc import Set as AbstractSet
+
+    from .dwarf_metadata import FieldInfo, StructLayout
 
 
 def _pe_export_id(e: Any) -> str:
@@ -1497,8 +1502,6 @@ def _normalize_type_name(name: str) -> str:
 
 
 def _diff_struct_layouts(o: object, n: object) -> list[Change]:
-    from .dwarf_metadata import FieldInfo, StructLayout
-
     old_structs: dict[str, StructLayout] = getattr(o, "structs", {})
     new_structs: dict[str, StructLayout] = getattr(n, "structs", {})
     changes: list[Change] = []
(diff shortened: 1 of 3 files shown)


napetrov:
@coderabbitai review

---

napetrov:
@coderabbitai review

---

PR #708, opened 2026-08-10
feat: G29 Phase 5 item 2 — macro/config dependency graph (MACRO_CONTROLS_DECL/DECL_USES_MACRO)

napetrov:
## Summary

Continues [G29 — Impact-Analysis Layer](https://github.com/abicheck/abicheck/blob/main/docs/contribute/plans/g29-impact-analysis-layer.md): implements Phase 5 item 2 ("Macro/config dependency"), the next unimplemented item in review-stated priority order after items 1 (template instantiation) and 6 (archive/link provenance), both merged in #691.

Adds a new Clang-backed L5 graph pass, `abicheck/buildsource/macro_graph.py`, populating two new edge kinds:

- **`MACRO_CONTROLS_DECL`** (`CONF_HIGH`) — a declaration is compiled only under a `#ifdef X`/`#ifndef X`/`#if defined(X)`/`#if !defined(X)` conditional region. Exact for these simple single-macro forms; compound conditions (`#if defined(X) && defined(Y)`, `#elif`) are deliberately unmodeled (nesting depth is still correctly tracked across them).
- **`DECL_USES_MACRO`** (`CONF_REDUCED`) — a declaration's own text references a macro `#define`d earlier in the same file. A textual heuristic (word-boundary regex, same-file only), documented as such.

Both edges join only onto `macro`/`source_decl` nodes the graph already carries (no new node kind — same "join-only-onto-an-existing-node" discipline `archive_graph.py`/ADR-057 D1 already establish for this phase). `MACRO_EXPANDS_TO_VALUE`/`MACRO_EXPANDS_TO_TYPE`/`MACRO_CONTROLS_EDGE` remain reserved, unpopulated vocabulary — see the module docstring for why each is deferred.

### Architecture

Two-pass extractor, mirroring `type_graph.py`'s shape:
- **Pass A** (`parse_clang_ast_decl_ranges`): pure `clang -ast-dump=json` parser producing each declaration's identity + `(file, begin_line, end_line)` span.
- **Pass B** (`scan_conditional_regions`, `find_decl_macro_uses`): pure raw-text scan for preprocessor conditionals and `#define` lines — clang's AST has no representation of preprocessor conditionals at all.

**Load-bearing empirical finding**: a declaration's `range.begin`/`range.end` don't reliably carry a `"line"` key — clang's JSON AST dumper threads one whole-document-wide sticky `(file, line)` cursor across every node's `loc`/`range.begin`/`range.end`, printing a field only when it changes. This is stricter than `type_graph.py`'s existing file-only sibling tracking, so `parse_clang_ast_decl_ranges` walks every AST node (not just declaration-shaped ones) to stay in sync — verified against nested structs, multi-statement bodies, and sibling declarations.

### Wiring

- `graph_facts.MACRO_DEP_EDGE_KINDS` (2 populated + 3 reserved), unioned into `source_graph.EDGE_KINDS` (that file sits at its 2000-line hard cap — a docstring paragraph elsewhere was trimmed to make room, net 1997 lines).
- `inline_graph_fold.fold_macro_graph`, wired into `fold_semantic_graphs()` after `fold_template_graph`.
- Docs: `buildsource/CLAUDE.md` module table, `docs/reference/source-graph-schema.md` (new section), `docs/contribute/plans/g29-impact-analysis-layer.md` (Phase 5 item 2 + G29.6 marked DONE).
- Changelog fragment under `### Added`.

### Deliberately out of scope for this slice

- Macro-expansion value/type tracing (`MACRO_EXPANDS_TO_*`).
- Per-edge (rather than per-declaration) conditional attribution (`MACRO_CONTROLS_EDGE`).
- Compound `#if`/`#elif` conditions.
- Cross-file macro-use tracking.

## Tests

`tests/test_macro_graph.py` (36 tests: conditional-region scanning incl. nested/sibling/deep-nesting cases, AST decl-range parsing incl. no-range skip, textual macro-use detection incl. defined-after-use and heuristic-overmatch cases, graph augmentation incl. join-only-onto-existing-node and diagnostic accumulation, plus one `@pytest.mark.integration` real-clang end-to-end case) + 2 new `fold_macro_graph` wiring tests in `tests/test_inline_changed_paths.py`.

## Verification

- `ruff check` / `ruff format --check` — clean
- `mypy abicheck/` — clean (0 errors, matches baseline)
- `python scripts/check_ai_readiness.py` — 0 new errors
- Fast suite (`pytest tests/ -m "not integration and not libabigail and not abicc and not slow and not golden"`) — 23504 passed, 0 failures

---
_Generated by [Claude Code](https://claude.ai/code/session_01QfxZYUuhjYZbBBvaMvoWEv)_

## Summary by CodeRabbit

* **New Features**
  * Added source-graph analysis for macro dependencies, virtual dispatch, vtables, and callback/function-pointer relationships.
  * Source graph collection now includes these analyses with confidence and coverage information.

* **Bug Fixes**
  * Improved member-call resolution, inherited virtual dispatch, qualified calls, operators, callback registration, and compact macro parsing.
  * Added graceful handling when compiler analysis is unavailable.

* **Documentation**
  * Expanded source-graph schema and analysis guidance.

* **Tests**
  * Added comprehensive unit, regression, and integration coverage.


napetrov (commented):
Triaged: this one is accepted as a documented, known gap rather than fixed. Minting a speculative `macro` node for an undefined guard macro would override the join-only-onto-an-existing-node invariant this module's own test suite pins down (`test_augment_never_mints_new_nodes`) — that's its own scoped design decision (macro nodes are cheap/bounded unlike the thousands of internal-only symbols `archive_graph.py`'s identical rule guards against, so the tradeoff may genuinely differ there), not something to override under a review comment. Documented in the `EDGE_MACRO_CONTROLS_DECL` docstring entry in `abicheck/buildsource/macro_graph.py` and in the G29 plan doc, with a regression test (`test_augment_negated_guard_produces_no_edge_when_macro_node_absent`) pinning the current, intentional behavior. Every other finding in this review round has a corresponding fix pushed.

---

napetrov (commented):
Replying to two review comments that are already fixed by pushed commits.

---

**"Resolve live member calls before filtering virtual edges" (virtual_dispatch_graph.py:212):** Fixed at the root in `call_graph.py` itself (commit 8c0d401) — `_find_referenced_decl` now resolves a `MemberExpr`'s string `referencedMemberDecl` via a new `member_index`, so `p->f()` correctly classifies as `call_kind="virtual"` before it ever reaches `virtual_dispatch_graph.py`'s `DECL_CALLS_DECL` filter. Verified against real Clang 17 output. Resolving.

**"Parse compact function-like macro definitions" (macro_graph.py:594):** Fixed in commit 4f1f4f8 — `_DEFINE_RE` (shared with `preprocessor_scan.py`) now accepts replacement tokens immediately after `)` with no required whitespace. Resolving.

---

napetrov (commented):
All fixed in bb7d139, pushed just now.

---

**"Pre-index member declarations before walking call sites" (call_graph.py:297):** Fixed with a new whole-AST pre-pass (`_index_member_decls`) that indexes every function-decl node's `id -> node` before `_walk_calls` starts resolving any call site, independent of visit order. Re-verified against the exact repro (`struct A { virtual void f(){ g(); } virtual void g(); };`) with real Clang 17 — the edge now resolves. Regression test added (`test_parse_member_call_to_later_declared_sibling_still_resolves`).

**"Recognize inherited virtual methods when classifying calls" (call_graph.py:297):** Fixed with a new `_ref_is_virtual()` helper — checks the resolved declaration's own `OverrideAttr`/`FinalAttr` child (present whenever `override`/`final` is written, confirmed against real Clang 17 for both keywords) in addition to `ref["virtual"]`. Kept self-contained within `call_graph.py`: an earlier attempt threaded a precomputed virtual-identity set from `override_graph.parse_clang_ast_virtual_methods` in, but that makes `call_graph.py` import from `override_graph.py`, which already function-locally imports from `call_graph.py` — the AI-readiness `import-cycle-growth` gate correctly flagged the resulting cycle, confirmed locally before reverting to the self-contained fix. Re-verified against the repro with real Clang 17 — now classifies `virtual`/`overapprox`. Regression test added (`test_parse_call_to_override_without_own_virtual_flag_is_still_virtual`).

**"Preserve callback-only declarations as graph endpoints" (callback_graph.py:747):** Fixed — `augment_graph_with_callback_registrations` now mints a missing `source_decl` endpoint (using this module's own fully-resolved AST edge information) instead of skipping the edge, mirroring `override_graph.py`'s established minting precedent in this same family. Tests updated to assert minting.

**"Qualify callback slot identities before joining them" (callback_graph.py:680):** Not fixed this pass — a real fix needs a scope-qualified identity in `call_graph.py` itself (shared infrastructure every consumer of `DECL_CALLS_DECL` depends on), which is a larger, separately-scoped change. Documented explicitly in the module docstring (new paragraph) and pinned by a dedicated regression test (`test_bare_name_collision_across_unrelated_functions_is_a_known_false_positive`) so the current, known-incorrect behavior can't silently drift further. Leaving open.

**"Preserve narrowed coverage for derived virtual passes" (inline_graph_fold.py:448):** Fixed — `fold_virtual_dispatch_graph` now propagates `narrowed_passes`/`narrowed_scope`/`degraded_passes` from its three prerequisite passes (`call_graph`/`type_graph`/`override_graph`) instead of always stamping full coverage. Regression tests added.

**"Add the new graph producers to topic ownership" (docs/reference/source-graph-schema.md:281):** Fixed — registered `macro_graph.py`, `virtual_dispatch_graph.py`, and `callback_graph.py` as `fact_sources` for the `impact-analysis` topic in `docs/_meta/topics.yaml`.

Verification: `ruff check`/`ruff format --check` clean, `mypy abicheck/` clean (0 errors), `python scripts/check_ai_readiness.py` 0 errors, full fast suite (`pytest tests/ -m "not integration and not libabigail and not abicc and not slow and not golden"`) 23580 passed / 0 failed.

---

napetrov (commented):
Both fixed in 1920010, pushed just now.

---

**"Keep qualified calls out of virtual dispatch" (call_graph.py:412):** Fixed with a new `_member_expr_is_qualified()` helper. Confirmed clang's `-ast-dump=json` carries no qualifier information at all — the text `-ast-dump` shows a sibling `NestedNameSpecifier TypeSpec 'B'` node for `obj.B::f()` that is completely absent from the identical AST's JSON form (byte-for-byte identical `MemberExpr` key sets between `obj.B::f()` and `obj.f()`). Derived instead from source-range arithmetic: an unqualified access's member-name token begins immediately after the receiver's own end plus one operator character; a qualifier occupies the extra bytes in between. While verifying this against the common "call the base implementation from an override" pattern (`struct D : B { void f() override { B::f(); } };`), found a second real gap: clang anchors a genuinely implicit `this` receiver's synthesized position at the member name itself (not before a written qualifier), so the receiver-to-member gap always reads as zero for that shape — added a separate branch (`_is_implicit_this_receiver`) that instead measures the `MemberExpr`'s own begin-to-end span. Verified against 7 real Clang 18 ASTs (explicit/implicit receiver × qualified/unqualified × dot/arrow, plus the three earlier repros to confirm no regression). Regression tests added.

**"Unwrap explicit casts around callback addresses" (callback_graph.py:491):** Fixed — `_address_taken_function` now also unwraps `CStyleCastExpr`/`CXXStaticCastExpr`/`CXXReinterpretCastExpr`/`CXXConstCastExpr`/`CXXFunctionalCastExpr`/`CXXDynamicCastExpr` around a `&func`/decay shape, not just `ParenExpr`. Verified against real Clang 18 output for both a C-style cast and `static_cast`/`reinterpret_cast`. Regression tests added (parametrized over all 5 cast kinds).

While verifying against CI, also found and fixed two pre-existing, unrelated test bugs the new code surfaced on Windows: `test_callback_graph.py`'s missing-clang test used an empty `BuildEvidence()`, which hits `extract_from_build`'s own "nothing to do" early return before its availability check ever runs (now uses a real compile unit, matching the established sibling-test convention); `test_macro_graph.py`'s real-clang end-to-end test filtered by an Itanium-only mangled prefix, which never matches a Windows runner's MSVC-targeted mangling (now matches by substring, which both schemes carry). A third Windows-only integration-test failure (`override_graph`/`virtual_dispatch_graph` showing as degraded rather than fully covered on that runner) is investigated but left open — the derived `virtual_dispatch_graph` is correctly propagating `override_graph`'s own degraded state (this session's own earlier fix working as designed), but the underlying cause is a real compiler behavior on the Windows CI runner I can't reproduce or safely guess at from this environment.

Verification: `ruff check`/`ruff format --check` clean, `mypy abicheck/` clean (0 errors), `python scripts/check_ai_readiness.py` 0 errors.

---

napetrov (commented):
Two fixed, one investigated and deliberately left open, in 2114cab, pushed just now.

---

**"Recognize cv-qualified function-pointer slots" (callback_graph.py:304):** Fixed — `_FUNCTION_POINTER_TYPE_RE` now accepts a top-level `const`/`volatile` between `*` and `)`. Confirmed against a real Clang 18 AST for `using H = void (*)(int); void reg(H const h);`: the desugared parameter type is exactly `"void (*const)(int)"`, which the previous regex (whitespace-only between `*` and `)`) rejected outright. Regression test added.

**"Do not stamp narrowed virtual passes as fully covered" (inline_graph_fold.py:465):** Fixed — this was a real bug in my own previous fix: I added the narrowed/degraded propagation but left the prior unconditional `extractor_passes["virtual_dispatch_graph"] = True` in place alongside it, so a narrowed/degraded run stamped both keys simultaneously. Converted to the same mutually-exclusive `if fully_covered: ... elif narrowed: ... elif degraded: ...` shape every sibling clang-backed pass already uses. (Confirmed this was live, not just theoretical: the Windows CI log I read while triaging the earlier `override_graph` degradation already showed exactly this — `virtual_dispatch_graph` present in both `extractor_passes` and `degraded_passes` at once.) Regression tests added/strengthened.

**"Ignore whitespace when detecting qualified member calls" (call_graph.py:516):** Confirmed real — `obj . f()` (legal whitespace around `.`) does misfire the same way a real `Base::` qualifier would, verified against real Clang 18 output. **Not fixed this round, deliberately**: closing this needs the actual source text between the two AST offsets to tell whitespace apart from a real qualifier's bytes, and `parse_clang_ast_calls` is a pure function over the AST dict alone (by design, for compiler-free unit testing) — it has no source-file access, unlike `macro_graph.py`'s own Pass B. No fixed numeric threshold soundly separates "a couple of stray spaces" from "a real single-letter class name plus `::`" in the general case, so I didn't patch it with an unsound cutoff. Documented explicitly in `_member_expr_is_qualified`'s own docstring and pinned by a dedicated regression test recording the current, accepted behavior. In practice this only misfires on non-idiomatic whitespace styling no common formatter produces. Leaving open.

Verification: `ruff check`/`ruff format --check` clean, `mypy abicheck/` clean (0 errors), `python scripts/check_ai_readiness.py` 0 errors.

---

napetrov (commented):
All fixed in b603d76, pushed just now.

---

**"Add an #ifdef/#ifndef fallback so nesting depth cannot desync" / "Add an `#ifdef`/`#ifndef` fallback" (macro_graph.py):** Fixed exactly as proposed — `_IF_RE` now also matches `#ifdef`/`#ifndef`. Verified against a real `#ifdef OUTER` / `#ifdef INNER extra` / `#endif` / decl / `#endif` repro (matches the suggested regression test). Regression test added.

**"Check `degraded` before `narrowed`" / "Require completed prerequisites before stamping full coverage" (inline_graph_fold.py):** Both fixed — reordered to check `degraded`-or-`missing` before `narrowed`, folding the "some prerequisite never ran at all" case (clang unavailable) into the same `degraded` bucket as a real per-TU diagnostic. `cli_buildsource_helpers.py` doesn't actually carry a second copy of this logic (it only calls `fold_virtual_dispatch_graph()`), so there was only the one place to fix. Regression test added for the mixed narrowed+degraded case.

**"Correct the stale pass counts in the docstring" (inline_graph_fold.py):** Fixed, count-free wording as proposed.

**"Correct the callback endpoint contract" (docs/reference/source-graph-schema.md):** Fixed — now describes the minting behavior.

**"Update the stale comment about skipped registration edges" (tests/test_callback_graph.py):** Fixed — comment now explains `my_handler(0);`'s real remaining purpose (exercising `call_graph.py`'s own `DECL_CALLS_DECL` edge) rather than a stale "correctly skipped" claim.

**"Patch or mark the remaining Clang-backed graph passes before they run" (tests/test_inline_changed_paths.py):** Fixed — now fakes `ClangOverrideGraphExtractor`/`ClangTemplateGraphExtractor`/`ClangCallbackGraphExtractor` too.

**"Assert the `clang` exit status before parsing its output" (tests/test_macro_graph.py):** Fixed as proposed.

Also applied, unprompted by a specific thread but from the same review round: `callback_graph.py`'s and `virtual_dispatch_graph.py`'s own edge-folding loops now iterate a `list(graph.edges)` snapshot rather than the live list before calling `graph.add_edge()` inside the loop.

**Deliberately not applied** (CodeRabbit's own "🔵 Trivial | 💤 Low value" tags): the `pytest.mark.parametrize` refactor suggestions for `test_macro_graph.py`'s simple-guard tests and `test_virtual_dispatch_graph.py`'s identity-helper/diamond/cycle tests, the `node_by_id` map perf optimization in `callback_graph.py`, and the lazy-import suggestion in `inline_graph_fold.py` — all stylistic/low-value per the reviewer's own assessment, not correctness issues.

Verification: `ruff check`/`ruff format --check` clean, `mypy abicheck/` clean (0 errors), `python scripts/check_ai_readiness.py` 0 errors.

---

napetrov (commented):
Replying to the 5 open Codex threads on this PR (pushed as `c4b3bb7`):

**"Degrade callback coverage with an incomplete call graph"** (`inline_graph_fold.py`) — Fixed. `fold_callback_graph`'s coverage stamp now combines its own extractor-run state with `call_graph`'s recorded coverage state worst-wins (`degraded`/missing > `narrowed` > `full`), mirroring the propagation `fold_virtual_dispatch_graph` already applies for its own three prerequisites. Covered by a new `TestFoldCallbackGraphPropagatesCallGraphCoverage` test class (all four coverage-state combinations).

**"Register the vocabulary's authoritative fact source"** (`docs/_meta/topics.yaml`) — Fixed. Added `abicheck/buildsource/graph_facts.py` to the `impact-analysis` topic's `fact_sources`.

**"Propagate callbacks from parameters into stored slots"** (`callback_graph.py`) — Investigated and confirmed real (reproduced end to end against real Clang 18: `reg(my_handler)` → `stored = h;` → `stored(1);` produces no `CALLBACK_MAY_INVOKE` edge for the real `invoke -> my_handler` relationship). Deliberately not fixed in this pass: closing it needs two new pieces — a new intra-procedural "slot aliases slot" edge for a function-pointer-to-function-pointer assignment, and a cycle-guarded transitive closure in Part A's join — a scoped follow-up of its own, not a drive-by extension. Documented in the module's own docstring ("Deferred") and pinned by `test_callback_propagated_through_stored_parameter_slot_is_not_joined`. Leaving this thread open.

**"Qualify callback slot identities before joining them"** and **"Ignore whitespace when detecting qualified member calls"** — both are the same class of pre-existing, documented, deliberately-unfixed limitations from earlier rounds on this PR (bare-name identity collision across unrelated functions; whitespace/comments between receiver and member widening the same source-range gap arithmetic real qualification uses). Both are pinned by dedicated regression tests (`test_bare_name_collision_across_unrelated_functions_is_a_known_false_positive`, `test_parse_call_with_whitespace_around_dot_is_a_known_false_positive`) documenting the current, known-limited behavior rather than being silently accepted. Leaving these threads open too.

---

napetrov (commented):
Fixed in `7ac893f`: rewrote both coverage paragraphs (`virtual_dispatch_graph` and `callback_graph`) to describe the actual mutually-exclusive `extractor_passes`/`narrowed_passes`/`degraded_passes` stamps derived from their clang-backed prerequisites, instead of the stale single-flag description.

---

napetrov (commented):
Fixed in `5605d4c`. Confirmed the exact repro end to end against real Clang 18 (`struct Base { virtual ~Base(); };` + a derived class produced no `TYPE_HAS_VTABLE` edge for either side before the fix). Added `override_graph.parse_clang_ast_virtual_destructor_owners` — a narrower pure function reading just the `virtual: true` flag clang stamps directly on a destructor's own AST node, deliberately not extending `_OVERRIDE_CANDIDATE_KINDS` (that would reopen the Itanium D1/D2 dual-mangling concern this module's docstring already explains for why destructors are excluded from override-*edge* matching). Its output feeds a third, independent vtable-presence seed stamped directly on the owning class's own `record_type` node.

One caveat worth flagging: this closes the case where something else in the graph (inheritance, a call) already minted the class's bare `record_type` node — a fully isolated class with nothing else referencing it doesn't get one at all (verified this is a pre-existing limitation shared by the other two vtable-presence seeds too, not new here).

---

napetrov (commented):
Replying to the three open Codex threads on this PR (pushed as `46c7aa2`):

**"Ignore comment delimiters inside string and character literals"** (`macro_graph.py`) — Fixed. Confirmed the exact repro (`const char *token = "/*";` before a real `#ifdef` silently swallowed the guard). `_lines_starting_inside_block_comment` now tracks ordinary `"..."`/`'...'` string/char-literal state alongside its existing `/* */` cross-line tracking, with escape-character handling. A raw string literal (`R"(...)"`) stays a documented, unhandled residual case — its arbitrary delimiter is a materially different parsing problem and I have no empirical evidence of that shape mattering here.

**"Infer virtuality for overrides without the override keyword"** (`call_graph.py`) — Confirmed real, already documented in `_ref_is_virtual`'s own docstring before this comment (a derived method redeclaring a virtual signature with neither `override`/`final` nor a repeated `virtual: true` is a known false negative). The fix needs the same whole-hierarchy signature-matching walk `override_graph.py` already performs; threading it into `call_graph.py` would form an import cycle with the reverse, already-established direction (`override_graph.py` already imports from `call_graph.py`). Not attempting a restructure here. Now pinned by a dedicated regression test (`tests/test_call_graph_extra.py`, a sibling split of `test_call_graph.py` which sits at its own 2000-line hard cap) rather than left untested. Leaving this thread open.

**"Preserve virtual facts for isolated implementation classes"** (`override_graph.py`) — Confirmed as the same pre-existing, shared limitation I flagged in my own reply on the destructor-seed fix: a fully isolated class (nothing else in the graph — no `TYPE_INHERITS` edge, no override edge — already minted its bare `record_type` node) has nothing for any of the three vtable-presence seeds to stamp onto. Minting a node here purely from an observed fact is riskier than it looks: `type_graph.py`'s own class-node minting has an independent identity quirk for a fully isolated class (its `CXXRecordDecl` walk mints only the class's injected-class-name identity, e.g. `record::Leaf::Leaf`, not the bare `Leaf` every seed's owner-recovery logic uses) — verified empirically. Minting a *second*, differently-identified `record_type` node from this module risks a silent duplicate-identity split for the same real class rather than closing the gap cleanly, and untangling that needs its own investigation into `type_graph.py`'s own record-identity computation, not a same-PR extension of this fix. Already pinned by `test_has_virtual_destructor_fact_without_a_record_node_mints_nothing` (and the analogous existing test for the leaf-virtual-method seed). Leaving this thread open.

---

napetrov (commented):
Both fixed in `2a17a11`:

**"Strip leading comments before matching directives"** — Confirmed real, including the worse nested-desync consequence (verified with an `#ifdef OUTER` / `/* note */ #ifdef INNER` / `#endif` / `#endif` repro: `OUTER`'s region was truncated 3 lines early before the fix). New `_strip_leading_inline_comment()` strips a same-line-closed leading `/* ... */` before every directive-family regex is tried, applied identically to `scan_conditional_regions` and `_macro_definition_lines`.

**"Give unnamed callback parameters a stable slot identity"** — Confirmed real end to end with a hand-built two-callee fixture (both `reg_a`'s and `reg_b`'s own unnamed slots collapsed onto one node before the fix). The slot identity now falls back to `f"{callee_identity}#param{position}"` when the parameter's own identity is empty, exactly as suggested.

---

napetrov (commented):
Both fixed in `d20e0ff`:

**"Process directives after multiline comment closures"** — Confirmed real (verified with a nested `#ifdef OUTER` / multi-line `/* opening\n*/ #ifdef INNER` / `#endif` / `#endif` repro: `OUTER`'s region was truncated before the fix, same desync shape as the earlier same-line leading-comment fix). New `_line_after_carryover_comment_closes()` finds where the carried-over comment ends and resumes matching against the live remainder, applied identically to `scan_conditional_regions` and `_macro_definition_lines`.

**"Unwrap list-initialized callback slots"** — Confirmed end to end against real Clang 18 (`handler_t slot{my_handler};` produced no `DECL_TAKES_ADDRESS_OF` edge before the fix). `_address_taken_function` now also unwraps a single-element `InitListExpr`, scoped to one element since a scalar type type-checks to exactly one initializer in valid C++.

---

napetrov (commented):
**"Classify virtual operator calls as virtual"** — Fixed in `d40fd15`. Confirmed against real Clang 18 (`B &b; b();` classified `direct`/`exact` before the fix). Two pieces were needed: `_classify_call` now also checks `CXXOperatorCallExpr`, and — the deeper part — `_find_referenced_decl` now upgrades a `DeclRefExpr`'s compact stub to the full declaration node via `member_index` when indexed, mirroring the resolution the string-shaped `referencedMemberDecl` path already had. Without that second piece, `_ref_is_virtual` was reading a stub with neither `virtual` nor `inner` populated, so extending the call-kind check alone wouldn't have been enough.

**"Honor line splicing before scanning line comments"** — Confirmed real (reproduced empirically: a nested guard's frame gets popped early by a `#endif` that's actually spliced into a still-open `//` comment). Deliberately not fixed this pass: unlike the four other comment-tracking fixes on this PR, correctly handling splicing needs a genuinely different per-line carry-over mechanism than the existing `in_block` state, threaded through every line-number-keyed caller without disturbing their physical-line-number contract (clang's own AST ranges are physical-line-based too) — a same-slice fix risks getting that fidelity subtly wrong, worse than the current, honest gap. Documented in the module's own docstring and pinned by a dedicated regression test. Leaving this thread open.

---

napetrov (commented):
Replying to CodeRabbit's first full review (and the duplicate Codex finding on the same issue), pushed as `f82c414`:

**"Preserve callback slots in dispatch-edge identity" (Codex) / same finding from CodeRabbit** — Fixed. Confirmed empirically (a hand-built two-slot fixture collapsed 2 expected edges into 1 before the fix). `CALLBACK_MAY_INVOKE` now also stamps the joined slot id as `role`, giving the two edges distinct relation keys (ADR-046 D1) instead of colliding.

**"Index fields in unnamed records"** — Fixed, exactly as suggested. Confirmed with a hand-built anonymous-union fixture.

**"Use `.get` for `narrowed_scope["call_graph"]`"** — Fixed, exactly as suggested.

**"Catch `TypeError` so a malformed AST location degrades to a diagnostic"** — Fixed at the one real call site (`abicheck/buildsource/macro_graph.py:989`); confirmed with a hand-built malformed-`line` fixture.

**"Add the missing SPDX identifier line"** — Fixed.

**"Skip guards check `clang++`, but both tests run `clang`"** — Investigated, not a real gap: every `fold_*` function in `inline_graph_fold.py` normalizes `clang_bin="clang"` to `"clang++"` internally before use (`clang_bin if clang_bin != "clang" else "clang++"`, applied at all 8 call sites) — so `_collect_source_graph(..., clang_bin="clang")` genuinely invokes `clang++` under the hood, and the existing `shutil.which("clang++")` guard is correct as written. No change made.

**"Correct the documented `fold_semantic_graphs()` order"** — Investigated, not accurate as stated: the `CLAUDE.md` line describes `fold_semantic_graphs()`'s own internal call order (verified correct: call→type→override→[virtual_dispatch]→template→macro→callback→include), not `cli_buildsource_helpers.py`'s separate out-of-band `collect` path, which does use a different internal ordering — but that's a distinct code path with its own already-documented history of falling behind `fold_semantic_graphs`'s list (see the inline comments there), not a doc inaccuracy in the line flagged. No change made.

**"Use scope-qualified callback slot identities"** — This is the same bare-name-identity-collision limitation already flagged by Codex earlier on this PR and already documented/pinned (`callback_graph.py`'s own docstring, "A real consequence of this weak identity..."; `test_bare_name_collision_across_unrelated_functions_is_a_known_false_positive`). Deliberately deferred for the reason already recorded there — fixing it needs a scope-qualified identity in `call_graph.py` itself (shared infrastructure, its own scoped follow-up), not a same-PR fix. Leaving this thread open alongside the original.

Nitpicks (parametrizing the simple-guard tests, extracting `override_graph.py`'s shared `bases_of`/`declared` setup, deriving node-id prefixes from `source_graph` helpers, a nested-record test) are reasonable but optional style/refactor suggestions with no functional impact — not applied in this pass.

---

napetrov (commented):
Fixed in `2f7335c`. Confirmed real (`#if defined FEATURE_X` fell through to the unmodeled fallback before the fix). Widened `_IF_DEFINED_RE`/`_IF_NOT_DEFINED_RE` to a proper alternation between the parenthesized and bare forms — deliberately not independently-optional `\(?`/`\)?`, which would have accepted a mismatched operand (`defined(X` or `defined X)`) as valid; a malformed one still falls through to the unmodeled fallback, verified with a dedicated test.

---

napetrov (commented):
**"Treat C++23 elifdef branches as unmodeled"** — Fixed in `c804fa3`. Confirmed real and worse than a silent miss: `#if defined(A) ... #elifdef B ... #endif` attributed the `#elifdef B` branch's declarations to `A` with `negated=False` — a wrong edge, not just a missing one. `_ELIF_RE` now also matches `#elifdef`/`#elifndef`.

**"Strip block comments throughout directive lines"** — Confirmed real (both `#/**/ifdef X` and `#if/**/defined(X)` reproduced as unrecognized). Deliberately not fixed this pass: closing it needs a systematic rewrite replacing every whitespace gap across the whole directive-regex pattern set, and mid-token block comments inside a directive are exceedingly rare, deliberately obfuscated styling — unlike the realistic comment placements already fixed elsewhere in this review cycle. Documented in the module's own docstring and pinned by a dedicated regression test. Leaving this thread open.

---

napetrov (commented):
`integration-tests (windows-latest)` failure on `c804fa3`: `test_collect_source_graph_folds_override_and_macro_graph_passes` fails because `override_graph`/`virtual_dispatch_graph` are in `degraded_passes` instead of `extractor_passes` on that Windows runner.

This is the same pre-existing, Windows-only environment issue already investigated earlier in this PR's history, not a new regression: `virtual_dispatch_graph`'s own coverage-propagation logic is working exactly as designed here (correctly reporting itself degraded because `override_graph` — one of its three prerequisites — degraded), and `override_graph` itself degrading traces to a real per-TU Clang diagnostic on that specific Windows toolchain, which I cannot reproduce or safely diagnose from this Linux sandbox (no Windows+clang environment available). Confirmed the only change to `override_graph.py` since this failure was last seen (`5605d4c`, the virtual-destructor vtable-presence seed) is purely additive — a new, independent third seed — and doesn't touch the override-edge extraction path this failure is actually in, so it isn't the cause.

Not fixing blind here, per this repo's own precedent for exactly this failure. If a Windows+clang repro surfaces (e.g. via the `contrib/abicheck-clang-plugin`'s own CI matrix or a maintainer's local Windows box), it'd be worth its own scoped investigation — not a guess from this session.

---

PR #710, opened 2026-08-10
fix: verdict and gate are different axes in the composite Action

napetrov:
Follow-up to #700, from Codex review on its final commit.

The Action's verdict mapping read the report's compatibility verdict only for exit 0. A severity policy that demotes `abi_breaking` below `error` while something else still gates exits non-zero with a report that still says `BREAKING`, and the Action published the *gating* tier instead — so a workflow branching on `verdict` acted on the wrong tier.

Fixed for the class rather than the reported case, across three review rounds:

- **exit 2** (an error-level `--crosscheck`, or `potential_breaking: error`) published `API_BREAK`
- **exit 1** (an error-level `addition`/`quality_issues`) published `SEVERITY_ERROR` — the same gap one exit code over, which the first fix missed
- `compare` carried both gaps although the reviews named only `scan`; every new test fails against the previous revision in both modes

**The gate deliberately does not move with the verdict.** `GATE_TIER` keeps the tier the exit code gated at, because the severity policy switching a break's gate off is precisely what the user asked for — letting an escalated `BREAKING` reach `fail-on-breaking` (default `true`) would re-gate the very finding the policy demoted.

Only a *break* may escalate: ranking alone let a `COMPATIBLE` report overwrite `COVERAGE_INCOMPLETE`, the inverse of the point. That was a bug I introduced and the exit-1 tests caught.

`action.yml` now documents `verdict`, `exit-code`, and the step's own outcome as the three different things they are — the raw gate tier is not the same as whether the composite step failed, since the `fail-on-*` inputs decide that.

## A second change was attempted here and reverted

This branch also carried a weak-export/COMDAT demotion (`func_export_dropped_inline_available`), reverted in `d582e92` before merge. Recording why, since the analysis is kept in `AGENTS.md`:

The demotion's invariant is *"the consumer already emitted its own COMDAT copy"*, and the only evidence available for it was `Function.is_inline` — which proves the `inline` **specifier**, not that a definition is present. Verified against real clang:

```
declared_only    inline=True  has_body=False
defined_here     inline=True  has_body=True
```

Both AST parsers assign that field straight from the specifier attribute with no body check, and `Function` carries no definition fact. So a header declaring a function `inline` while defining it elsewhere would have had a genuine load-time break demoted to a risk.

Reverted rather than fixed again because the missing fact cannot be supplied where it matters: `has_inline_definition` is cheap on direct-clang and derivable from DWARF, but **castxml — the default L2 header backend — exposes only declaration attributes**. Gating on it correctly ships a feature inert by default while carrying a schema bump.

`AGENTS.md` keeps the empirical disproof, what a real fix needs, the castxml blocker, and the three constraints the attempt did establish (old-side evidence is load-bearing since `WEAK` does not imply vague linkage; the predicate must be a leaf both `diff_symbols` and `diff_platform` consume; the demotion may replace only the removal finding, never the signature comparison).

## Verification

- Full suite: **24145 passed, 502 skipped, 4 xfailed, 0 failed**
- FP-rate gate: 41 cases, 0 FP / 0 FN; shadow contract measurement on baseline
- `check_ai_readiness`: 0 errors; `ruff check` clean
- Every new test confirmed failing against the revision before it

https://claude.ai/code/session_01FcGP8rSpY8YK9jjCDd9xJW

## Summary by CodeRabbit

* **New Features**
  * Added COMDAT and weak-linkage analysis to improve visibility into symbols that may remain available through inline definitions.
  * Preserved detailed compatibility findings while identifying the policy that actually blocks a check.
* **Bug Fixes**
  * Improved consistency across scan and comparison results, severity policies, and coverage policies.
* **Documentation**
  * Clarified the independent meanings of verdicts, exit codes, and workflow outcomes.
  * Documented current limitations around linkage-blind removals and COMDAT evidence.
* **Tests**
  * Expanded coverage for linkage analysis, verdict reconciliation, and policy-gating scenarios.


napetrov:
Both findings on `6fe3bdf` were right and are fixed in `46dcc1e`.

**P1 — old-side inline evidence (`symbol_linkage.py`).** This was a genuine hole in the argument, not just a missing guard. I had reasoned "WEAK ⇒ vague linkage ⇒ the consumer carries its own copy", and the first half of that is simply false: an ordinary out-of-line function marked `__attribute__((weak))` is `WEAK` too, and a consumer compiled against *that* declaration holds only an undefined dynamic reference. Making it inline on the new side and dropping the export breaks it at load time — exactly the case I was demoting. And the direction matters for the reason you give: the consumer was built against the **old** headers, so only the old side can establish that it emitted its own copy. The predicate now requires `is_inline` on both sides — old for already-compiled consumers, new for newly-compiled ones.

That narrows the demotion: a template instantiation or implicit special member whose producer didn't set `is_inline` now falls back to being reported as a removal. That's the safe direction and the same absence-of-evidence discipline the binding check already uses, so I took it rather than trying to infer vague linkage some other way.

Guarded by a new FP-corpus FN sentinel. Restoring the old predicate makes it fail as a **false negative** — "real break scoped away" — which is the precise failure mode:

```
false negatives (real break scoped away): ['out_of_line_weak_export_dropped_stays_breaking']
ERROR: false negatives 1 exceed baseline 0
```

**P2 — positional-argument shift (`model.py`).** Correct, and `hidden_friend_owner` right there in the same class already documents the rule I broke. Both fields are appended after all pre-existing fields now. Rather than assert it by eye I diffed the field lists against `origin/main`:

```
Function: prefix preserved (31 fields), appended ['elf_binding']
Variable: prefix preserved (13 fields), appended ['elf_binding']
```

Full suite 24218 passed / 0 failed; FP-rate gate 45 cases, 0 FP / 0 FN; shadow measurement on baseline; ai-readiness 0 errors.

---

napetrov:
Third P1 (`symbol_linkage.py` — signature checks swallowed by the demotion) fixed in `04dc6fa`. Confirmed before fixing:

```
kinds  : ['func_export_dropped_inline_available']
verdict: Verdict.COMPATIBLE_WITH_RISK      # on a func_return_changed break
```

The mistake was in what the demotion is *for*. It replaces the **removal** finding, not the comparison — the declaration is still present on the new side, which is the entire basis for demoting, so it can also have changed. This turns out to be the one path where a symbol leaves `_public_functions` yet still has a new-side declaration worth comparing, which is why `_check_function_signature` had never run on it. Both findings are emitted now; an unchanged declaration still yields the lone risk finding.

I checked the sibling early return for the same shape rather than assuming it was fine. `FUNC_VISIBILITY_CHANGED` also holds a new-side declaration but is **not** the same bug and deliberately keeps swallowing the signature check: a function that went hidden isn't callable by anyone, so how its signature changed is moot. A demoted export is still *available* through the header, which is exactly why a change to it matters. That distinction is recorded in the code so it doesn't read as an oversight.

Guarded by two tests plus an FP-corpus FN sentinel that fails as "real break scoped away" when the early return is restored — 46 cases, 0 FP / 0 FN.

### On the verification

The first full-suite run after this change reported one failure, `test_diff_functions_builds_type_map_not_naive_dict`. That was my own doing, not the change: I edited `diff_symbols.py` *while the suite was running*, adding 31 lines at line 803, above the function at line 892 that this test reads via `inspect.getsource`. The code objects still held pre-edit offsets, so it inspected the wrong span. Two independent facts agree — the line arithmetic, and that the new code is gated on a ChangeKind only the new predicate can produce, so every pre-existing path returns exactly what it did before.

Rather than close on that reasoning I re-ran the suite with the tree untouched: **24233 passed, 0 failed**. Worth flagging for anyone reading the history that a mid-run source edit silently invalidates a run of this suite.

---

napetrov:
The displaced-gate finding on `action/run.sh:1110` is fixed in `92a28ee`.

You're right, and it's the same asymmetry that caused the original problem: the `API_BREAK` summary branch carried a severity note and `BREAKING` carried none, so escalation turned that gap into a failing summary saying only "Binary ABI break detected" on a step that an `addition: error` gate had failed.

Fixed as one shared `_blocking_gate_note` called from both branches rather than a third copy of the text — a per-branch copy is what created the gap, so duplicating it again would have been fixing the case rather than the class. It states which severity categories blocked, and, when the verdict was escalated above the tier that gated, which tier `fail-on-*` actually applies to.

Worth recording one thing from writing the tests. My negative assertion `"escalated" not in summary` failed against correct code — pytest's `tmp_path` embeds the test's own name, `test_an_un**escalated**_breaking_s0`, into the summary through the library paths. Both assertions now match the literal note text. A bare substring check against output containing temp paths is a trap, and it would have been easy to "fix" the code to satisfy a broken test.

Verification: full suite **24148 passed, 0 failed**; the two positive tests fail against the previous revision; `ai-readiness` 0 errors; `ruff` and `bash -n` clean.

---

napetrov:
Addressed in `d512614`.

**COMDAT scan** — three findings, all real:

- **Raw paths.** `CompileUnit.output` is normalized for *persistence*, not for reading back (`~/…` redaction per ADR-032 D7; Ninja/Make/Bazel outputs relative to `directory`). `_resolved_object` now expands `~`, joins a relative label onto the unit's own directory, and skips a label naming no existing file.
- **Preserving a loaded scan.** On the `base_build` path an unconditional rescan over persisted labels could replace a valid `comdat` with an empty, unresolvable one. A scan that established nothing no longer displaces one that did, and a build where nothing resolves leaves the field untouched.
- **Gating.** The sweep is opt-in (`ABICHECK_COLLECT_COMDAT=1`) — parsing every object's symbol table is real I/O and nothing consumes the result yet. The gate lives in `build_evidence` rather than at the call site because `inline.py` sits exactly on the 2000-line hard cap (the readiness gate caught the overflow).

**Action severity note.** Correct that `scan` is the second case bypassing the `fail-on-*` flags: its final branch sets `FINAL_EXIT=1` on a configured severity category at *every* tier, so the "still follows the flags" wording was the opposite of what happens there. Fixed, with a test pinning both the compare and scan directions.

**Windows `integration-tests` is failing on the base branch, not from this PR.** `test_collect_source_graph_folds_override_and_macro_graph_passes` fails identically on `main` at `7c59880c` — same assertion, same `degraded_passes={'override_graph', 'virtual_dispatch_graph'}` (job [93566336746](https://github.com/abicheck/abicheck/actions/runs/31420962799/job/93566336746) vs. this PR's [93581259412](https://github.com/abicheck/abicheck/actions/runs/31426992767/job/93581259412)). `main` was green at `73b5576`, so it regressed between. I'll re-run against the base once it recovers.

Local: 24054 passed in the fast lane, `ruff check` clean, AI-readiness 0 errors.

---

PR #713, opened 2026-08-10
feat: G31 Phase C/D — clang `Param.is_restrict`, and the header-backend capability matrix

napetrov:
## Summary

Continues G31: closes one more Phase C fact-completeness gap on the direct-clang backend, and ships Phase D's documentation deliverable — the CastXML vs. direct-clang vs. hybrid capability matrix — as a generated, drift-checked artifact rather than prose.

## Motivation

Phase D asks for "a backend capability matrix (which facts/edges each can and cannot see)", plus "why CastXML can't do all graph edges" and a "how to extend clang parsing" roadmap. Building the matrix as something *checked against the parsers* rather than hand-typed immediately paid for itself: its drift check reported a divergence this plan had never recorded, and that divergence turned out to be a live false positive.

**`Param.is_restrict` was castxml-only.** It has been populated by `dumper_castxml._resolve_cv_restrict` since it shipped; `dumper_clang.py` never populated it; and unlike `deprecated`/`is_scoped` before this phase, `_diff_param_restrict` compares the two bools **directly, with no producer gate**. So comparing a CastXML-parsed side against a clang-parsed side of *unchanged* headers reported `param_restrict_changed` for every `restrict`-qualified parameter.

## Changes

**Phase C — `Param.is_restrict` on the direct-clang backend (schema v22)**

- `dumper_clang_qualifiers._clang_param_is_restrict` extracts the qualifier, following this phase's established direction (extract on clang rather than gate, since a plain bool is directly cross-comparable). Verified against real clang 18 output for: the C `restrict` spelling, the C++ `__restrict`/`__restrict__` spellings clang normalizes to one form, typedef indirection (visible only in `desugaredQualType`), and the `int *restrict *` (pointee-qualified) vs `int **restrict` (parameter-qualified) split. Builds on `_last_top_level_ptr_end`, the top-level-qualifier helper the cv-qualifier work already needed, plus `_declarator_group`/`_follows_type_operator_keyword` for the cases where C's declarator precedence puts the parameter's own pointer inside parentheses (`int (*restrict p)[3]`) and where a `typeof`/`decltype` operand only looks like it does.
- **Two gates**, both following precedent already set in this phase:
  - a **header-tier gate** (`_both_header_aware`, as `param_defaults` uses) — DWARF, PDB and the symbol-table paths never populate this fact, so their `False` meant "not collected", not "not qualified". This gate can only remove findings that were false, since a non-header side is unconditionally `False`.
  - schema **v22**'s `clang_restrict_facts_reliable` — the fourth instance of the v19/v20/v21 shape, covering `"hybrid"` alongside `"clang"` for the same reason v20 did (a merge keeps castxml's `params` verbatim for a matched function, but appends a clang-only function's parameters as they are).
- Snapshot disk-cache version bumped to `10` (v7/v8 precedent) so a warm cache re-extracts.

**Phase D — `docs/reference/header-backend-capabilities.md`**

- 95 rows, one per field of the six declaration dataclasses the two backends build, generated by `scripts/gen_backend_capability_matrix.py` from `scripts/backend_capabilities.py` (the `gen_platform_matrix.py`/`platform_capabilities.py` splice pattern, one sibling over). The **hybrid column is derived** from `dumper_hybrid.py`'s real backfill lists, not hand-typed.
- Plus the two explainers Phase D names: why CastXML structurally cannot supply the graph's body-dependent edges (its XML has no statements at all, so abicheck sources every edge from clang regardless of snapshot backend), and the five-route clang-extension roadmap (JSON AST dump / `clang.cindex` / LibTooling / `PPCallbacks` / `VTableContext`) with the rule the matrix bears out: if clang *prints* the fact, close it in `dumper_clang.py`; if clang only *computes* it, it belongs in the LibTooling companion.
- `tests/test_backend_capability_matrix.py` re-derives every published claim by reading both parsers with `ast` — asking not "is this keyword passed?" but "is it passed a real expression, or a placeholder like `size_bits=None`/`is_opaque=False`?" — and fails when claim and code disagree.

**Gaps documented rather than fixed** (castxml was not installable in this environment, and this phase's bar is verify-before-claiming): `EnumType.underlying_type` is clang-only, so a castxml/hybrid enum keeps the default `int`; a hybrid merge drops clang-only facts it does not explicitly backfill for any declaration both backends saw; an opaque handle type is *absent* from a clang snapshot rather than opaque; and `Variable.value`, `Variable.access`, `Param.is_va_list` have no producer on any layer (making `param_became_va_list` unreachable on real input).

**Two splits, forced by the hard cap.** `dumper_clang.py` (1970) and `diff_symbols.py` (1999) were both within 30 lines of the 2000-line limit, so this lands with `dumper_clang_qualifiers.py` (the top-level-qualifier helpers, re-exported so existing `dumper_clang._name` import paths resolve) and `diff_param_qualifiers.py` (the `param_restrict` + `param_va_list` comparison loops). The two `@registry.detector` registrations deliberately stay at their original positions in `diff_symbols.py` and delegate: `registry.detector()` stamps an incrementing counter that fixes the order findings appear in every report, so registering from the new module moved both detectors and reordered the coverage-gap rows.

## Checklist

- [x] Tests added / updated for new behaviour — `tests/test_clang_param_restrict.py` (69 cases: the predicate, the same spellings against a live clang dump — including a systematic fifteen-parameter C declarator corpus whose expected answers are encoded in the function names — and both gates, each with a positive control), `tests/test_backend_capability_matrix.py` (18), plus a new round-trip class in `tests/test_serialization_roundtrip.py`
- [x] Changelog fragment added — `changelog.d/20260810_g31_param_restrict_and_backend_matrix.md`
- [x] Docs updated — new reference page (in `mkdocs.yml` nav, `scripts/CLAUDE.md` inventory, `check_ai_readiness.py`'s generated-file registry, `docs/AGENTS.md`'s regeneration list); `snapshot-format.md` updated for v22; the G31 plan doc appended to per its own convention
- [x] Full fast+golden suite green (24,854 passed); `ruff check` clean; `mypy abicheck/` clean (0 errors, baseline held); `check_ai_readiness.py` 0 errors; `check_docs_contract.py` 0 errors; `mkdocs build --strict` passes

Four schema-version assertions in existing tests were re-expressed against the `SCHEMA_VERSION` constant (they assert "a fresh serialize stamps the *current* version"); the one deliberate literal pin was bumped. G20 fixtures regenerated — the diff is exactly the version bump plus the new field.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SgYuhpzuxRejWb53EQbd1e)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of parameter `restrict` and `va_list` qualifier changes.
  * Reduced false positives across header-analysis backends.
  * Correctly handles callbacks, references, parenthesized pointers, function pointers, and typedefs.
  * Avoids relying on unreliable legacy snapshot data.

* **Documentation**
  * Added header-backend capability documentation and updated snapshot format guidance.
  * Documented backend limitations, hybrid behavior, and capability-matrix generation.

* **Compatibility**
  * Updated snapshot schema and cache versions to preserve qualifier-data reliability.


napetrov:
Pushed `0a48047` addressing both review findings and the CI failures.

**Codex's callback-parameter finding was real — thanks, and it was worse than described.** Reproduced against real clang 18: for `void (*cb)(int *restrict)` the parameter's own `*` sits inside parentheses, so `_last_top_level_ptr_end` finds no depth-0 pointer, `_field_own_cv_source`'s no-pointer fallback returned the whole spelling, and the regex matched the *callback argument's* `restrict`. castxml answers False for the same declaration, so this would have introduced a fresh cross-backend false positive of exactly the kind the PR removes.

Fixed by requiring a depth-0 pointer before searching at all, rather than falling back to the whole spelling. One correction to the suggested framing: this needs no function-declarator detection or AST-structure inspection, because the requirement is semantically exact rather than a heuristic — C11 6.7.3p2 allows `restrict` only on a pointer to an *object* type, so a spelling with no top-level pointer can never be restrict-qualified. I verified that rather than citing it: `void (*restrict cb)(int)` is rejected outright by clang in C and C++ alike (`pointer to function type 'void (int)' may not be 'restrict' qualified`), which also retired an "accepted false negative" note an earlier draft of the fix had written for that shape. There's now a test asserting the compiler really does reject it, alongside live-clang coverage of the callback spellings.

**The unit-test failures were mine, and a good catch by the golden tests.** Registering `diff_param_qualifiers` from `checker.py`'s top import block pulled `diff_symbols` (and `diff_symbols_renames`, which owns `fingerprint_renames`) forward from its own import much further down the file, moving those detectors ahead of `diff_elf_layout`'s and reordering the coverage-gap rows in every report. Fixed by importing the two shared helpers inside the detectors — the remedy the root `AGENTS.md` already names for this coupling. The full fast suite *including* the `golden` marker is green locally (24,738 passed); the everyday fast command excludes that marker, which is the argument for `verify.py --profile pr` before opening a PR.

**`integration-tests (windows-latest)` is failing on `main` too**, at this PR's own base commit (`7c59880`, run [31420962799](https://github.com/abicheck/abicheck/actions/runs/31420962799)) — `test_collect_source_graph_fold1`, on `override_graph`/`virtual_dispatch_graph` being degraded. Unrelated to this diff; not something I've touched here.

**Codecov's 77.63% patch figure** was measured on the failing run. Both new modules measure 100% line+branch locally, so I expect it to resolve on this run; I'll follow up if it doesn't.

---

napetrov:
`1addf10` — the second Codex finding was right, and fixing it properly turned up a third case neither round named.

`int (*restrict p)[3]` is parenthesized for the same declarator-precedence reason a callback parameter is, but it *is* a legal restrict-qualified object pointer that castxml sees, so requiring a depth-0 `*` turned the round-1 false positive into a false negative. Confirmed against clang 18 (`int (*restrict)[3]`).

The rule that holds in both directions is **a parenthesized declarator group wins over depth 0**: when C forces the parameter's own `*` into parentheses, everything at depth 0 belongs to something else — the pointee, or the callback's parameter list. `_declarator_group` finds the group holding the parameter's pointer, identified as the first depth-0 parenthesized group whose contents begin with `*`; that distinguishes it from a function pointer's trailing parameter list (begins with a type name) and from an array extent (not parenthesized at all).

Twelve spellings now verified end to end against real clang, including **`int *(*)(int *restrict)`** — a function pointer with a pointer *return* type. That one has a depth-0 `*` from `int *`, so the round-1 rule would have searched the callback's parameter list and matched its `restrict`: a false positive of the original kind that the depth-0 requirement did not actually remove. It's covered by both a unit case and a live-clang case now.

On the suggestion to distinguish function-pointer declarators from object-pointer ones: not needed, and I checked rather than assumed. C11 6.7.3p2 allows `restrict` only on a pointer to an *object* type, so a function-pointer declarator group can never legally carry the qualifier — clang rejects both `void (*restrict cb)(int)` and `int *(*restrict cb)(void)` in C and C++ alike. One rule covers both declarator kinds.

Full fast suite including the `golden` marker is green locally (24,753 passed), and both new modules measure 100% line+branch under the full suite. Codecov has since updated to "All modified and coverable lines are covered by tests" — the earlier 77%/78% figures were from the runs whose canonical lane had failed.

---

napetrov:
`bdcb199` — both findings from this round were right, and the first one was right in a way my previous fix wasn't.

**Registration order.** Making the shared-helper imports function-local restored *`diff_symbols`'* detectors (what the golden tests caught) but not the two **moved** ones. Measured against a worktree at the base commit: `param_restrict` went 16 → 5 and `param_va_list` 20 → 6. One methodology note worth recording, since it nearly sent me the wrong way — that comparison is only valid with `PYTHONPATH` pointed at the worktree; the editable install otherwise shadows it and reports a confident, false "identical".

Fixed by keeping the **registrations in `diff_symbols` at their original source positions** and moving only the loop bodies. Each detector applies its snapshot-level gates and hands `diff_param_qualifiers` the already-selected public-function maps, so the new module takes `dict[str, Function]` and imports nothing from `diff_symbols`. That direction is forced rather than chosen: once `diff_symbols` imports the new module, a back-import is a real cycle, and the AI-readiness import-cycle gate walks *every* AST import — function-local ones included — so that escape hatch isn't available here. `checker.py`'s import is dropped entirely; `ensure_loaded`'s prefix discovery imports the module anyway and it now registers nothing.

Re-measured: all 65 detectors in the same order as base, `param_restrict` back at 16, `param_va_list` at 20. `tests/test_detector_discovery.py` now pins the relative order that encodes (after `functions`/`fingerprint_renames`, before `constants`) rather than literal indices, which any unrelated new detector would shift. That guard demonstrably fails against the pre-fix state: at index 5 with `functions` at 9, `param_restrict > functions` is false.

**Reference wrappers.** Also correct: `int *__restrict &` is a legal C++ reference to a restrict-qualified pointer whose qualifier clang still prints, while castxml's walk follows only `CvQualifiedType`/`Typedef`/`ElaboratedType` and stops at the outer `ReferenceType`. Matching the visible qualifier re-created the cross-backend disagreement from the other side, so a top-level `&` after the parameter's own pointer now answers False — deliberately matching castxml rather than the spelling. Confirmed against real clang for the lvalue, rvalue, plain-reference and reference-to-array-of-pointers shapes.

Full fast suite including `golden` is green (24,761 passed); ruff, `mypy abicheck/`, ai-readiness and docs-contract all clean. `diff_symbols.py` ends at 1987 lines — below both the cap and its original 1999, so the split still buys headroom.

`integration-tests (windows-latest)` is still red for the pre-existing reason already noted above (fails on `main` at this PR's base); not re-reporting it per event.

---

napetrov:
`2a2a3e4` — Codex round 4 (both valid) and all five CodeRabbit actionable items, plus both nitpicks.

**The declarator predicate, done differently this time.** Hand-picking spellings had failed three rounds running, so I built a systematic corpus instead: every declarator shape, compiled twice — once with `restrict` on the parameter's own pointer and once without — with the expected answer encoded in the function name (`_r`/`_n`). Fifteen parameters, exactly one wrong, and it pointed at the real rule rather than another patch:

**the parameter's own declarator is the INNERMOST parenthesized `(*…)` group.** `int (*(*restrict p)[3])[2]` (p *is* restrict) and `int (*restrict (*p)[3])[2]` (p is plain; the qualifier is on the array's element type) print as `int (*(*restrict)[3])[2]` and `int (*restrict (*)[3])[2]` — taking the outer group answers both alike and gets the second wrong. Descending fixes it and keeps the other fourteen. The `decltype` case was narrower: that `(` opens a *dereference*, not a declarator, and a declarator group's `(` is never immediately preceded by an identifier character — which separates the two exactly. The corpus is now a live-clang test, so the next shape arrives with its expected answer instead of being argued about.

**CodeRabbit's five** were all real: two stale "four helpers" counts; a plan-doc claim that the predicate reuses `_field_own_cv_source` (that was the *discarded* first cut — the same document's own findings section contradicted it); "Two model fields" over a list of three; `_parse` converting a clang failure into a skip *after* clang was found (so a future clang rejecting a fixture would turn the class green — fixed to skip only for the missing binary); and hybrid reliability tested at legacy-unreliable and clang-current-reliable but never current-hybrid-reliable.

Both nitpicks too: the summary counted a companion-tool row as "only direct-clang populates", contradicting that row's own note — now counted by own-parse capability with the sentence saying so. And the evidence scanner's attribute-assignment upgrade was unscoped across classes; since it changes no verdict for either backend today, I removed it rather than elaborating it. A post-construction assignment now surfaces as a loud claim-vs-evidence test failure instead of silent contamination.

Not adopted: the docstring-coverage pre-merge check is CodeRabbit's own threshold rather than a repo gate — ai-readiness and docs-contract both pass clean.

**On CI:** `integration-tests (ubuntu-24.04)` failed on `bdcb199` where it had passed on `0a48047`, which I can't yet explain — the log window available to me only reaches the coverage-upload step, not the pytest output. The Windows lane's failure is the known pre-existing `test_collect_source_graph_folds_override_and_macro_graph_passes` (`override_graph`/`virtual_dispatch_graph` degraded), and that test is a plausible candidate for ubuntu too, since nothing in `bdcb199` touches source-graph collection. I'm watching this run to see whether it reproduces; if it does I'll dig into it rather than assume it's flaky.

Full fast suite including `golden` is green locally (24,772 passed); ruff, mypy, ai-readiness, docs-contract and `mkdocs build --strict` all clean.

---

napetrov:
Diagnosed the `integration-tests (ubuntu-24.04)` failure I flagged as unexplained above — it is **not** related to this PR, and it's a different cause from the Windows one.

```
FAILED tests/test_perf_dump_scaling.py::test_dwarf_parse_scaling_stays_subquadratic
AssertionError: DWARF parse scaling exponent 1.98 regressed toward O(n^2)
assert 1.9787116500114856 < 1.9
```

That test times a DWARF parse at two sizes and derives a scaling exponent from the single ratio `log(t2/t1) / log(n2/n1)`, asserting `< 1.9`. It measured 1.98 — a marginal overshoot of a wall-clock measurement taken from two data points on a shared runner, which is about as noise-sensitive as a gate gets.

This PR touches no DWARF code. The complete list of `abicheck/` files in the diff is `diff_param_qualifiers.py`, `diff_symbols.py`, `dumper_clang.py`, `dumper_clang_qualifiers.py`, `model.py`, `serialization.py`, `snapshot_cache.py` — the DWARF parse path (`dwarf_snapshot.py`, `dwarf_metadata.py`, …) is untouched, so there is no mechanism by which this diff could change that timing.

So the two integration lanes are failing for two independent, pre-existing reasons: Windows on `test_collect_source_graph_folds_override_and_macro_graph_passes` (`override_graph` degrades when its extractor emits diagnostics — a toolchain condition), and ubuntu-24.04 on this timing threshold. I'm deliberately not touching either: widening a perf threshold or altering an unrelated gate as a drive-by from this PR is exactly the kind of change that should be its own, deliberate decision. Flagging them for a maintainer instead.

Correcting my earlier note in passing: I'd guessed ubuntu might be the same source-graph test as Windows. It isn't — worth stating plainly rather than leaving the guess standing.

---

napetrov:
Rebased onto `main` (`e3dfedb`) and pushed `ac3fc5b`. One conflict, in the G31 plan doc — additive on both sides (main's tenth review round vs. this branch's `Param.is_restrict` section), resolved by keeping both in chronological order. Re-checked after the rebase: the capability matrix is still in sync against main's updated `dumper_clang.py` (95 facts), and detector registration order is unchanged (65 detectors, `param_restrict` at 16, `param_va_list` at 20).

## Review round 5 — `typeof` operand misread as a declarator group

Fixed, and the near-miss is the useful part.

Clang prints `typeof` **spaced** whatever the source wrote — `typeof(*gp)` comes back as `typeof (*gp)`, confirmed against clang 18:

```
typeof (*gp) *restrict      # source wrote `typeof (`
typeof (*gp) *restrict      # source wrote `__typeof__ (`
typeof (*gp) *restrict      # source wrote `typeof(`
```

So the round-4 rule ("the `(` directly follows an identifier character") saw a space, took `(*gp)` for the declarator group, searched `*gp`, and answered `False`. `decltype(` passed only because clang happens to print *it* unspaced.

The obvious repair — skip whitespace, then require an identifier — is wrong in the other direction, which I checked before applying rather than after:

```
'int (*restrict)[3]'       char before, skipping spaces = 't'  -> "expression operand"
'typeof (*gp) *restrict'   char before, skipping spaces = 'f'  -> "expression operand"
'void (*)(int *restrict)'  char before, skipping spaces = 'd'  -> "expression operand"
```

It would have regressed rounds 1 and 2. The declarator spellings read as `<identifier><space>(` too; the *specific keyword* is the only thing separating the two families. The rule is now a closed set (`decltype`, `typeof`, `__typeof`, `__typeof__`, `typeof_unqual`) — these being the only constructs clang prints in a type spelling whose parenthesized operand is an expression.

Re-verified against real clang, 25 cases, no regressions: all three `typeof` source spellings plus the unqualified control, the `decltype` pair, the fifteen-parameter `_r`/`_n` declarator corpus, and the reference wrappers. Added as both unit rows and a live-clang test; the test's docstring records why the whitespace-skipping variant is not the fix, so it doesn't get "simplified" back later.

## CI

Both `integration-tests` failures are pre-existing and unrelated to this diff:

- **ubuntu-24.04** — `test_perf_dump_scaling.py::test_dwarf_parse_scaling_stays_subquadratic`, a two-point wall-clock timing assertion (`exponent 1.98 < 1.9`). This PR touches no DWARF file. (Correcting my earlier guess in this thread that it was the same test as the Windows lane — it isn't.)
- **windows-latest** — `test_collect_source_graph_folds_override_and_macro_graph_passes`, failing on `main` as well.

Everything else is green locally: 24854 passed on the fast+golden lane, `ruff check`, `mypy abicheck/` (0 errors), AI-readiness and docs-contract both at 0 errors.

---

PR #719, opened 2026-08-11
feat: G31 Phase C fact-completeness — EnumType.underlying_type on castxml

napetrov:
## Summary

Closes one item of Phase C's fact-completeness backlog: `EnumType.underlying_type` was clang-only — a castxml (and therefore hybrid) enum silently kept the model default `"int"` regardless of the header's real underlying type.

## Motivation

No diff detector reads this field directly (`enum_underlying_size_changed` is computed from DWARF instead), but `tu_merge.py`'s cross-TU ODR-conflict check for enums does — so it was trivially passing on castxml input even when two translation units genuinely disagreed on an enum's underlying type. A pinned castxml build became available in this environment, making the fact newly verifiable (it wasn't previously, per `AGENTS.md`'s Known Gaps section).

## Changes

- `abicheck/dumper_castxml.py`: `parse_enums()` now resolves the `<Enumeration type=...>` id through the same `_underlying_type_name` helper `parse_typedefs()` already uses (following a typedef chain to its concrete base type). Verified against real castxml 0.6.3 output: `enum class Color : short` reads `short int` (fixed underlying type); a plain `enum Status { OK, FAIL }` reads `unsigned int` (compiler-chosen from the member value range).
- `scripts/backend_capabilities.py`: `EnumType.underlying_type`'s castxml capability moves from `NONE` to `FULL`; `docs/reference/header-backend-capabilities.md`'s generated matrix regenerated to match.
- `tests/test_backend_capability_matrix.py`: the "hybrid drops an unbackfilled clang-only fact" regression fixture moves to `RecordType.is_template_pattern`, which still has that castxml-`NONE`/clang-`FULL`/not-backfilled shape (this field no longer does).
- `tests/test_castxml_enum_underlying_type.py` (new): synthetic-XML unit coverage plus a live, real-castxml-invoking `integration`-marked test confirming the fixed and compiler-chosen cases.
- `docs/contribute/plans/g31-header-graph-default-on-followup.md`: records the closure alongside the other Phase C fact-completeness closures already documented there.

## Checklist

- [x] Docs updated (generated backend-capability matrix + plan doc)
- [x] `ruff check` / `ruff format --check` pass on touched files
- [x] `mypy abicheck/dumper_castxml.py` clean
- [x] `pytest tests/test_dumper_unit.py tests/test_castxml_enum_underlying_type.py tests/test_backend_capability_matrix.py tests/test_tu_merge.py` — 254 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01HtfBbjbcqLNeaxZ4sCKwPq)_


napetrov (commented):
Good catch — fixed in a2bf834: the unfixed-enum assertion is now target-aware (`"int"` on `sys.platform == "win32"`, matching the MSVC ABI; `"unsigned int"` elsewhere), with a comment explaining why only the unfixed case is target-dependent (a fixed underlying type is a literal language feature, so `Color`/`Big` stay unconditional).

---

napetrov (commented):
Good catch — fixed in 8d92fbb: bumped `snapshot_cache._SNAPSHOT_CACHE_VERSION` (`"12"` → `"13"`) and `source_extractors/castxml.CASTXML_EXTRACTOR_VERSION` (`"0.1"` → `"0.2"`), each with an inline comment following the existing v7–v12 convention in `snapshot_cache.py`, so a warm cache from before this change is invalidated on upgrade instead of replaying the old `"int"` default.

---

napetrov (commented):
Fixed in a296cfe, discussed with the maintainer given the scope. `_diff_generated`/`_diff_typedefs`/`_diff_macros`'s content-change loops are now gated on a new `FactCompatibility.structured_content_comparable` flag (`fact_set.py`), which requires matching producer/producer_version (same rule set `opaque_hashes_comparable` already uses for body/template hashes) before trusting a `type_hash`/`value` diff as a real content change — a persisted 0.1-produced baseline compared against a fresh 0.2 extraction now suppresses the false `GENERATED_HEADER_CHANGED` instead of trusting the stale-vs-real `type_hash` diff.

This needed one more piece underneath it: the castxml extractor never stamped a `fact_set`/`coverage` block at all before this, so it was previously treated as a "pre-C.8 producer" and exempt from every comparability rule in this file regardless of `CASTXML_EXTRACTOR_VERSION`. It now stamps both (`producer="castxml-source"`, `producer_version=CASTXML_EXTRACTOR_VERSION`, and an honest `compiler_family`/`unsupported`-family coverage), so the new gate actually has data to act on for the concrete case you flagged.

Accepted trade-off: a genuine content change spanning a real producer/compiler upgrade may now go unreported until both sides are re-collected with the same producer — the same trade-off `opaque_hashes_comparable` already makes for body/template hashes. Recorded as a C.11 amendment in `docs/contribute/adr/038-build-integrated-fact-collection-variants.md`.

---

napetrov (commented):
Correct — fixed in d8eac8e. `check_fact_compatibility` now distinguishes a symmetric fact_set absence (both sides `{}`, the genuine pre-C.8 forward-compat case) from an asymmetric one (exactly one side `{}`, exactly the already-persisted-pre-0.2-baseline shape you described): only the symmetric case keeps the "nothing to check, so nothing is gated" behavior. `structured_content_comparable`/`opaque_hashes_comparable`/`source_edges_comparable` are all `False` now for `check_fact_compatibility({}, new_fs)` and its reverse. New regression tests cover both directions plus the concrete castxml-extractor-produced case.

---

napetrov (commented):
Fixed in 226c7d8: added `_castxml_tool_version()` (cached `castxml --version` probe, mirrors `clang.py`'s `_clang_compiler_version`) and wired it into `default_fact_set(..., compiler_version=...)`. Verified against real castxml — the integration test now asserts a non-empty `compiler_version`.

---

napetrov (commented):
Fixed in 226c7d8: `_diff_fact_coverage()`'s description now names `generated_header_changed`/`public_typedef_target_changed`/`public_macro_value_changed` explicitly when `structured_content_comparable` is false, and the `structured_facts_comparable` branch's now-inaccurate "content-change findings are unaffected" claim is removed.

---

napetrov (commented):
Fixed in fb6aa5b: `parse_types()`'s identity-dedup now canonicalizes a `class`/`struct` opaque-redeclaration tie via `min(kind)` (mirroring `tu_merge._record_kinds_compatible`), so `class Handle; struct Handle;` (either order) always resolves to `kind="class"`. Added `test_parse_types_opaque_redecl_kind_canonicalizes_regardless_of_order` (both orderings) as regression coverage.

---

napetrov (commented):
Fixed in fb6aa5b: `deprecated` is now merged across every redeclaration sharing an identity (first non-empty `[[deprecated]]` message wins, independent of which node the kind tie-break keeps), so `struct Handle; struct [[deprecated("old")]] Handle;` now surfaces the message regardless of declaration order. Added `test_parse_types_opaque_redecl_merges_deprecated_from_a_later_decl` as regression coverage.

---

napetrov (commented):
Fixed in fb6aa5b: `CastxmlSourceExtractor.cache_identity_extra()` now returns the probed castxml/bundled-Clang identity (mirrors `ClangSourceExtractor.cache_identity_extra()`'s existing `--gcc-path` fold), so `source_replay._extractor_version()` folds it into the D8 per-TU cache key and a warm cache no longer replays a stale `SourceAbiTu` after the castxml binary at the cached path is upgraded or swapped. Added `test_castxml_extractor_cache_identity_extra_folds_probed_tool_version` as regression coverage.

---

napetrov (commented):
Fixed in 362f479: the merge now distinguishes "no `DeprecatedAttr`" (`None`) from "attribute present, empty message" (`""`) throughout — `parse_types()`'s per-identity merge and both `_build_record()` branches. `struct Handle; struct [[deprecated]] Handle;` now correctly emits `deprecated=""` instead of `None`. Added `test_parse_types_opaque_redecl_merges_bare_deprecated_marker` as regression coverage.

---

napetrov (commented):
Fixed in 362f479: `_CASTXML_BUNDLED_COMPILER_RE` now matches `(?:clang|LLVM) version` (same spelling variance `dumper_castxml_probe._CLANG_VERSION_RE` already documents/handles for the identical bundled frontend). Added `test_castxml_tool_version_recognizes_llvm_spelled_banner` as regression coverage.

---

napetrov (commented):
Already addressed by 27c4053: `structured_content_agrees` in `check_fact_compatibility()` is computed without the `same_recipe`/`hash_recipe_id` term — only `opaque_hashes_comparable`/`source_edges_comparable` honor that override, per the function's own docstring ("No `hash_recipe_id` override here -- see this function's own docstring"). Resolving; flag if this doesn't match what you're seeing on the current head.

---

napetrov (commented):
Already addressed by 27c4053 (and extended by 362f479 for the LLVM-spelled banner variant): `_castxml_tool_version()` combines both the `castxml version X.Y.Z` line and the bundled-compiler identity line (`_CASTXML_BUNDLED_COMPILER_RE`, now matching `clang|LLVM version`) into one `compiler_version` string, so two installs sharing a castxml release but bundling a different frontend no longer read as the same identity. Resolving; flag if this doesn't match what you're seeing on the current head.

---

napetrov (commented):
Fixed in cad289d: verified empirically against real clang (`-Wdeprecated-declarations` on `struct [[deprecated]] H; struct [[deprecated("old")]] H;` warns `"old"`, pointing at the second decl; reversed order warns the bare form) — clang's own diagnostic uses the most recent redeclaration's marker, not the first. Switched the merge from `setdefault` to always-overwrite so a later marker wins. Added `test_parse_types_opaque_redecl_uses_most_recent_deprecated_reason` (both orderings) as regression coverage.

---

napetrov (commented):
Fixed in e441730: added `fact_set_rollup_is_inconsistent()` to compute the distinguishing bit alongside `rollup_fact_set()`'s own `{}` collapse, `link_source_abi()` stamps it onto `surface.coverage["fact_set_inconsistent"]`, and `check_fact_compatibility()` gained `old_inconsistent`/`new_inconsistent` keyword parameters so an inconsistent side's `{}` is treated the same as an asymmetric absence (not forgiven) rather than a genuine symmetric pre-C.8 pair — this now suppresses `structured_content_comparable`/`opaque_hashes_comparable`/`source_edges_comparable` correctly. Added unit coverage for the helper, `check_fact_compatibility`'s new gating, `link_source_abi`'s stamping, and an end-to-end `diff_source_abi` suppression test.

---

napetrov (commented):
Fixed in 9872a6a: `_castxml_tool_version()` now reads `f"{r.stdout}\n{r.stderr}"` and `_CASTXML_VERSION_RE` is case-insensitive, mirroring `dumper_castxml_probe.py`'s own combined-transcript normalization. Added `test_castxml_tool_version_reads_stderr_banner` and `test_castxml_tool_version_matches_banner_case_insensitively` as regression coverage.

---

napetrov (commented):
Fixed in 9872a6a: `_surface_fact_set_inconsistent()` now requires `cov.get("fact_set_inconsistent") is True` rather than `bool(...)` truthiness. Added `test_diff_still_reports_content_changes_when_inconsistent_flag_is_string_false` as regression coverage.

---

napetrov (commented):
Fixed in fe79764: `parse_types()` now decides kind (`min(kind)` over every non-definition redeclaration sharing an identity, via a new `opaque_kinds` tracking dict) and location (the most-public declaration wins; ties keep whichever was already kept, via `_decl_is_public`) independently — mirroring `tu_merge.py`'s own location-preservation principle for its cross-TU merge. Added `test_parse_types_opaque_redecl_keeps_public_location_over_private` (both declaration orderings) as regression coverage.

---

napetrov (commented):
Fixed in 3bf0ce3: `override_kind` is now only honored when the surviving record is actually opaque (`is_opaque and override_kind`) — a definition's own kind always wins. Added `test_parse_types_definition_kind_not_overridden_by_opaque_redecl` as regression coverage.

---

napetrov (commented):
Fixed in 3bf0ce3: `structured_facts_comparable` is now also gated by `inconsistent`, not just the recipe-dependent categories — an inconsistent side's `{}` cannot establish absence either. The pre-existing asymmetric-absence exemption (one side simply never stamped a fact_set) stays untouched, per the function's own docstring. Updated the existing regression test to assert the new gating.

---

napetrov (commented):
Fixed in 3bf0ce3: `_castxml_tool_version()` now goes through `run_bounded_for_extraction()` (the same deadline-aware bounded path every other subprocess this extractor runs already uses) instead of a bare `subprocess.run(timeout=5)`. Added `test_castxml_tool_version_is_bound_by_deadline_run_bounded` as regression coverage.

---

napetrov (commented):
Fixed in 8d4003c: added `_executable_stat_key()` (dev/ino/mtime_ns/size via `os.stat`, mirroring `dumper_toolchain._executable_sha256`'s own stat-based cache key) and folded it into `_castxml_tool_version()`'s `lru_cache` key at both real call sites (`cache_identity_extra`, `_stamp_fact_set_and_coverage`), so a same-path executable swap within one long-lived process gets a fresh probe. Added `test_castxml_tool_version_refreshes_when_executable_is_swapped` as regression coverage.

---

napetrov (commented):
Investigated and confirmed real — but it's a pre-existing architectural gap, not something this PR introduces. `diff_filtering._find_opaque_types()` (and `_find_by_value_types()`/`_downgrade_opaque_type_changes()`) already key by bare `RecordType.name`, and `_root_type_name()` matches against `Change.symbol`, which `diff_types.py` also stamps with the bare name (`name = t_old.name`) for these change kinds. This collision risk already existed identically for castxml's own `is_opaque=True` types before this PR — this PR's fix only makes it newly *reachable* on the direct-clang backend, since that backend previously silently dropped every forward-decl-only type instead of emitting a stub at all (so it never populated `_find_opaque_types()`'s set from clang input).

A correct fix needs qualified identity threaded consistently through several call sites in `diff_filtering.py` (`_find_opaque_types`, `_find_by_value_types`, `_downgrade_opaque_type_changes`, `_root_type_name`'s ~5 call sites), none of which have existing test coverage for the cross-namespace-collision case to validate a change against. That's a systematic, cross-cutting rework of shared bare-name-keyed logic spanning `diff_types.py` and `diff_filtering.py` together — well beyond the safe scope of a reactive fix under this PR's review-comment loop, and risks a regression in already-shipped opaque-suppression behavior if done hastily.

Recorded in `AGENTS.md`'s "Known gaps" section (`0041a10`) with the full investigation, per this repo's own established convention for exactly this situation, rather than attempting a rushed broad rewrite here.

---

napetrov (commented):
Fixed in 0041a10: the probe now re-checks the deadline directly on any failure; a still-exhausted budget re-raises the real `deadline.DeadlineExceeded` — deliberately NOT wrapped into `SourceExtractionError`, matching the un-swallowed exception `_replay_cache_lookup()`'s own bare `deadline.check()` already lets propagate for this same phase (and, via `extract()`'s per-TU path, letting `_extract_one`'s narrow `except SourceExtractionError` miss it lets it correctly reach `scan_engine.py`'s outer `except deadline.DeadlineExceeded` handler instead of being silently absorbed). Added `test_castxml_tool_version_reraises_deadline_exceeded_under_active_budget` as regression coverage.

---

napetrov (commented):
Fixed in 95e9665: `cache_identity_extra()` now falls back to the executable's own stat signature (`dev:ino:mtime_ns:size`, already computed via `_executable_stat_key` for the `lru_cache` key but previously discarded) when the probed version string is empty, so two distinct broken/unparseable installs at the same path no longer collapse to the same `""` identity. Scoped to `cache_identity_extra()` only — `compiler_version`'s own provenance value stays `""` on the same failure. Added `test_cache_identity_extra_falls_back_to_stat_when_version_unparseable` and `test_cache_identity_extra_stays_empty_when_binary_unresolvable` as regression coverage.

---

napetrov (commented):
Fixed in 55cedd0. `_stamp_fact_set_and_coverage()` now stamps `fact_set.compiler_version` with the same stat-fallback identity `cache_identity_extra()` already used, via a new shared `_castxml_identity_with_stat_fallback()` helper, instead of the version probe's bare `""` on failure. `compiler_version` is only ever used for opaque string-equality comparison and diagnostic display (`fact_set.py`'s `compiler_version_mismatch` issue), never parsed as a real version, so the `stat:...` fallback is exactly as valid a value there as in the cache key. Added `test_parse_root_stamps_stat_fallback_compiler_version_when_probe_fails` verifying two differently-broken installs at the same path now persist different `compiler_version`s.

---

napetrov (commented):
Fixed in 71fb49d: `same_recipe` now also requires `not inconsistent` — a matching `hash_recipe_id` no longer overrides an `old_inconsistent`/`new_inconsistent` flag. Added `test_check_fact_compatibility_matching_recipe_id_does_not_override_inconsistent`.

---

napetrov (commented):
Fixed in 71fb49d: `has_signal` now also counts a bare `fact_set_inconsistent: true` flag, not just non-empty `fact_set`/`fact_family_states`. Added `test_diff_reports_coverage_incomplete_from_bare_inconsistent_flag`.

---

napetrov (commented):
Fixed in a89da7e. `check_fact_compatibility()` now appends an explicit `fact_set_inconsistent` `FactSetIssue` whenever `old_inconsistent`/`new_inconsistent` is set — `check_fact_set_compatibility()` itself only ever compares the two rolled-up dicts it's given and has no visibility into those flags, so two identical fact_sets previously produced no issue at all. Added `test_check_fact_compatibility_reports_issue_for_inconsistent_matching_fact_sets` and `test_diff_reports_coverage_incomplete_when_inconsistent_flag_has_matching_fact_sets`.

---

napetrov (commented):
Fixed in a89da7e. `class`/`struct` are now collapsed to one fixed spelling (`"struct"`) before folding into `opaque_kinds`, so the canonicalized value depends only on the identity's real kind category, not on which subset of legally-compatible redeclarations happens to be present in a given snapshot. `union` stays distinct (mirrors `tu_merge._record_kinds_compatible`). Added `test_parse_types_opaque_redecl_kind_stable_when_a_compatible_redecl_is_added` and updated the existing canonicalization tests' expected value.

---

napetrov (commented):
Fixed in 94b96c2. Confirmed this was a real regression the previous fix introduced: unconditionally folding every opaque record to a fixed `"struct"` spelling relabeled an unambiguous single redeclaration away from its real class-key. Now tracks the full set of raw kinds observed per identity (`opaque_kind_sets`) and only folds to a fixed spelling once genuine ambiguity (2+ distinct raw kinds) is observed — an identity with exactly one raw kind keeps it unchanged, so it matches a later same-key complete definition exactly. New `_reduce_opaque_kind_set()` in `dumper_clang_qualifiers.py` (`dumper_clang.py` sits at its 2000-line hard cap). Added `test_parse_types_opaque_kind_matches_later_complete_definition_of_same_key`.

---

napetrov (commented):
Fixed in 94b96c2. `scripts/backend_capabilities.py`'s `EnumType.underlying_type` row now marks direct-clang `_PARTIAL` (`⚠️ Partial`) rather than `_FULL` — `_enum_underlying()` correctly reads a fixed enum's real underlying type but hard-codes `"int"` for the unfixed case, which castxml's own extraction correctly resolves. Regenerated `docs/reference/header-backend-capabilities.md` via `python scripts/gen_backend_capability_matrix.py`.

---

napetrov (commented):
Partially fixed in 36b2a82. The **persisted** half is fixed: `fact_set.compiler_version` now folds in a stat signature of the resolved `cc_bin` (`_stamp_fact_set_and_coverage` already has the per-TU `compile_unit` in scope). The **cache-key** half (`cache_identity_extra()`) is deliberately deferred: that hook is a zero-arg, once-per-extractor-instance call (`source_replay._extractor_version()`), while the resolved compiler genuinely varies per compile unit in the common case (no explicit `--gcc-path`-style override) — closing it needs a wider signature change to `source_replay.py`'s shared cache-key infra (also used by `ClangSourceExtractor`), which I didn't want to rush reactively. Recorded in `AGENTS.md`'s Known gaps section; leaving this thread open for visibility on the remaining half.

---

---

napetrov (commented):
Investigated and confirmed reproducible directly against `_ClangAstParser` — an unrelated forward-only tag `struct Foo;` alongside a separate anonymous-struct typedef `typedef struct { int x; } Foo;` collide into one `identity` key, and the tag is silently dropped from the snapshot. Deliberately **not fixed** in this PR: a correct fix needs the tag-namespace vs. ordinary-namespace distinction threaded through `identity` and every downstream consumer of this same function's merge maps, and `parse_types()` has already been revised three times this session (the opaque-stub fix, the kind-canonicalization fix, and that fix's own regression fix) — each individually re-verified against `dumper_clang.py`'s exact 2000-line hard cap. A fourth same-session change to the same function is the risk pattern this repo's own "known gaps over risky reactive patches" convention exists to avoid. Recorded in `AGENTS.md`'s Known gaps section; leaving this thread open.

---

napetrov (commented):
Confirmed and reverted in f9a5457. You're right — the stat-based fold was actively worse than the pre-existing gap: (1) two TUs resolved to `gcc` vs `g++` (an ordinary mixed C/C++ build) get different stat suffixes purely from being different files, and (2) a baseline built on one CI worker compared against another never shares device/inode at all — both spuriously trip `fact_set_inconsistent` and suppress every structured/opaque/source-edge finding on completely healthy comparisons. Reverted `compiler_version` back to castxml-only identity rather than patch further under review pressure. Updated `AGENTS.md`'s Known gaps entry to record that a correct fix needs a portable semantic identity (a real `cc_bin --version` probe, normalized) for both halves, not filesystem stat fields. Added `test_parse_root_compiler_version_ignores_resolved_emulated_compiler_path` covering exactly the mixed-driver case you flagged.

---

napetrov (commented):
Investigated and confirmed reproducible directly against `_ClangAstParser`. Deliberately **not fixed**: this isn't fixable at the extraction layer — an opaque snapshot has no way to know in advance which of the ambiguous, compatible keys a later definition will use, so no fixed canonicalization choice can match every possible future definition. A correct fix needs new provenance carried on `RecordType` itself (e.g. a `kind_ambiguous`-shaped field, schema-versioned, populated by both header backends, read by `diff_types_abicc_parity._diff_type_kind_changes()` to skip exactly this one transition shape) — a cross-cutting model change, not a scoped fix. `dumper_clang.py`'s `parse_types()` has already been revised four times this session for adjacent findings in this exact area; recorded in `AGENTS.md`'s Known gaps section rather than attempted under continued review pressure.

---

napetrov:
CI note: `Validate examples (gcc, debug-headers, shard 3/3)` failed on `f9a5457` due to `apt-get update` hitting `403 Forbidden` on `packages.microsoft.com` (azure-cli repo) — an external infra hiccup unrelated to this PR's diff, not a code issue. Watching for recovery; will re-run when the upstream repo is reachable again.

---

PR #724, opened 2026-08-11
scan --against: make the 20-finding report cap overridable, add per-kind truncation counts

napetrov:
## Summary

`scan --against`'s report cap on embedded findings (`_MAX_BASELINE_FINDINGS = 20`) was hard-coded with no CLI flag or env var override, and `suppressed_truncated: true` gave no way to see the *shape* of what was cut without rerunning at a higher cap.

- `cli_scan_baseline.py`: `_MAX_BASELINE_FINDINGS` is now a resolved default via `_resolve_max_baseline_findings()` — an explicit override wins, else the `ABICHECK_MAX_BASELINE_FINDINGS` env var, else the built-in default (20). `_baseline_summary`/`_add_severity_blocking_compatible_findings`/`_run_baseline_compare` all take a `max_findings` parameter, threaded from one resolved value so the two truncation stages can't disagree on the cap.
- Whenever the `findings` or `suppressed` list is truncated, the summary now also reports `findings_truncated_kinds`/`suppressed_truncated_kinds` — a `ChangeKind -> count` breakdown of what was cut.
- `scan_engine.py`/`cli_scan.py`: threaded `max_findings` through `run_scan_core` and added `scan --max-findings N` (a positive int, `click.IntRange(min=1)`). It's rejected without `--against`, mirroring the existing `_COMPARISON_ONLY_FLAGS` usage-error pattern the other comparison-only flags already use.
- Regenerated `docs/reference/cli-reference.md` (`scripts/gen_cli_reference.py`).

**Deliberately out of scope:** threading `max_findings` into `ScanRequest`/`service_scan.py` (the typed Python API). That file sits exactly at the repo's 2000-line AI-readiness hard cap, and any addition trips the file-size gate — a same-PR drive-by extension isn't worth splitting that file for. The CLI/env-var path covers the reported DX pain point.

## Testing

- `pytest tests/test_cli_scan_baseline.py tests/test_pr494_scan_regressions.py tests/test_scan_baseline_headers.py tests/test_cli_scan.py tests/test_scan_compare_parity.py tests/test_cli_contract.py -q` — all passing, including new tests for `_resolve_max_baseline_findings` and the per-kind truncation breakdown.
- `ruff check` / `ruff format --check` clean.
- `mypy abicheck/` — 0 errors.
- `python scripts/check_ai_readiness.py` — no new errors (file-size gate unaffected).

---
_Generated by [Claude Code](https://claude.ai/code/session_012MPZTRXMP4HyuAPcKzcQHS)_


napetrov:
Addressed all three Codex findings in 4c376ba:

- **P2 "Expose the override on `ScanRequest`"**: added `ScanRequest.max_findings`, forwarded through `run_scan()` → `run_scan_core`, and validated by the existing `_reject_comparison_only_fields` guard (baseline-only, same as the other compare-only fields). `service_scan.py` sits exactly at the repo's 2000-line AI-readiness hard cap, so I compacted two pre-existing multi-line comments in the same class rather than adding an allowlist entry — net file size stayed at/under 2000 lines.
- **P2 "Bump the scan schema version"**: `SCAN_SCHEMA_VERSION` → `1.10`, with a version-history entry in `abicheck/schemas/__init__.py` matching the existing style for the two new additive `diff` keys.
- **P1 "Register the new report fields in the topic registry"**: added `abicheck/scan_engine.py` to the `verdicts` topic's `fact_sources` in `docs/_meta/topics.yaml`, alongside the already-registered `cli_scan.py`/`cli_scan_baseline.py`. `reference/exit-codes.md` (that topic's `reference_page`) and `reference/cli-reference.md` (`cli-surface`'s `reference_page`, regenerated in the first commit) already cover this surface, so no new topic was warranted per docs/AGENTS.md's "when does a new fact need a new page" criteria.

All three plus new/existing tests, `ruff check`, `mypy abicheck/` (0 errors), and `check_docs_contract.py` (0 errors) pass locally.


---

napetrov:
Addressed the two follow-up findings in d70e2c8:

- **P2 "Validate the API cap before starting the scan"**: `run_scan()` now rejects a non-positive `ScanRequest.max_findings` up front (as `errors.ValidationError`), before any scanning happens. `_resolve_max_baseline_findings()` now raises a plain `ValueError` instead of `click.ClickException` — it's a shared engine-layer helper reached from both the CLI (where `click.IntRange(min=1)` already rejects a bad value before this is ever reached) and the click-free Python API.
- **P1 "Register truncation as its own documentation topic"**: rather than a brand-new topic (this is a report-*shape* concern, not a new mental model, per docs/AGENTS.md's own "when does a new fact need a new page" criteria), I registered it under `output-formats` — the topic that already owns scan's JSON report-shape narrative (it's where `SCAN_SCHEMA_VERSION` is documented) — and added real prose to `docs/use/output-formats.md` covering `--max-findings`/`ABICHECK_MAX_BASELINE_FINDINGS` and the two new truncation fields, with a worked JSON example. `fact_sources` now includes `schemas/__init__.py`, `cli_scan_baseline.py`, and `scan_engine.py`.

Tests, `ruff check`, `mypy` (0 errors), `check_docs_contract.py` (0 errors), and `mkdocs build --strict` all pass locally.


---

PR #731, opened 2026-08-12
docs: fix Mach-O/DWARF, no-headers, auto-frontend, and config-plan claims

napetrov:
## Summary

Documentation correctness hotfix (PR1 scope from an external documentation review), each item independently verified against source before editing.

## Motivation

A recent documentation review of `main` flagged several front-door pages making claims that contradict either the actual code or a dedicated reference page that already gets the same fact right. Before touching anything, every claim was verified against source (`macho_metadata.py`, `dwarf_metadata.py`/`dwarf_unified.py`, `cli_options.py`, `dumper.py`, `action.yml`, `cli_project.py`) — two of the review's claims turned out to need a different diagnosis than "the docs are wrong" (see below), so this PR fixes what's actually inaccurate rather than following the review verbatim.

## Changes

- **Mach-O/DWARF**: `README.md`, `docs/start/getting-started.md`, and `docs/learn/architecture.md` no longer claim DWARF debug-info cross-check on macOS. abicheck has no Mach-O DWARF/debug-map reader (`macho_metadata.py` has none; `dwarf_metadata.py`/`dwarf_unified.py` are ELF-only via `pyelftools`) — `docs/reference/platforms.md` already documented this correctly, these pages didn't match it.
- **No-headers wording**: `--header`'s `help=` text in `abicheck/cli_options.py` (and the generated `docs/reference/cli-reference.md` it feeds) no longer says a headerless native binary "falls back to symbols-only mode" unconditionally. Without headers, abicheck actually tries DWARF (ELF) / PDB (PE) first, only degrading to symbols-only when neither headers nor debug info are present. Same correction applied to `docs/use/cli-usage.md` and `docs/use/tool-modes.md`, which repeated the stronger, inaccurate claim.
- **`--ast-frontend auto` opt-in**: `docs/reference/header-backend-capabilities.md` and `docs/use/github-action.md` now state that `auto` never silently falls back from castxml to clang without the explicit `--allow-ast-frontend-fallback` opt-in (or `ABICHECK_ALLOW_AST_FALLBACK=1`), matching `action.yml` and `cli_options.py`/`dumper.py`, which already enforce this.
- **Config File stale claim**: `docs/reference/config-file.md` no longer says the `targets:`/`bundles:`/`profiles:` run-plan generator is "planned but not built yet" — `abicheck project plan` implements it (`cli_project.py`). Also cross-references why an explicit `--config` fails loudly on a malformed file while an auto-discovered one only warns and falls back, instead of leaving them as two disconnected-looking rules.
- Regenerated `docs/reference/cli-reference.md` via `scripts/gen_cli_reference.py`.
- Added a `changelog.d/` fragment.

**Two review claims were investigated and found to need a more careful diagnosis, not a literal fix:**
- `--ast-frontend auto`'s resolver logic is actually *consistent* across `action.yml`/`cli_options.py`/`dumper.py` — only the two summary docs above were missing the opt-in caveat, not a genuine three-way disagreement.
- The config-discovery "contradiction" the review described is two correctly-documented rules for two different code paths (explicit `--config` vs. auto-discovered config), not an actual conflict — addressed by cross-referencing them instead of picking one to be "right".

## Checklist

- [x] Tests added / updated for new behaviour — n/a (docs + `help=` text only; existing `cli_reference`/`fair_metadata`/`cli_options` suites pass)
- [x] `pre-commit run --all-files` passes locally — ran the equivalent (`ruff check`, `ruff format --check`, `mypy abicheck/cli_options.py`) directly

## Verification

- `mkdocs build --strict` — clean
- `python scripts/check_docs_contract.py` — 0 errors (same pre-existing warnings as `main`)
- `python scripts/check_ai_readiness.py` — same 2 pre-existing, unrelated `adr-status-sync` errors as `main` (confirmed via `git stash`)
- `ruff check`/`ruff format --check`/`mypy` on `abicheck/cli_options.py` — clean
- `pytest -k "cli_reference or fair_metadata or cli_options or header_help or symbols_only"` — all pass

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01TnUW9GU5pAaQF13BynYq2x)_

## Summary by CodeRabbit

- **Documentation**
  - Clarified platform-specific analysis capabilities, including Mach-O/macOS limitations and headerless binary behavior.
  - Expanded guidance for AST frontend selection, fallback rules, and incompatible configurations.
  - Documented project configuration validation, generated run plans, and GitHub Actions usage.
  - Added an unreleased-version announcement banner showing the latest published release.
  - Improved navigation and added documentation topics for agent skills and frontend resolution.
  - Clarified agent skill version requirements and installation guidance.


napetrov:
Both Codex findings verified against source and fixed in 9d764ba:

- **`auto`'s device-context exception** (`header-backend-capabilities.md`): confirmed `_resolve_single_ast_backend` in `dumper.py` only rejects a non-`host` `--frontend-context` for an *explicit* `--ast-frontend castxml` (or env-pinned castxml), not plain `auto` — `auto` skips castxml and routes to clang instead (`test_header_ast_parser_auto_device_routes_to_clang_not_rejected`). Added this as a documented exception rather than folding it into the general opt-in row, since it's a different mechanism (castxml can't satisfy the request at all) from the toolchain-error fallback the opt-in flag gates.
- **Malformed-config scope** (`config-file.md`): confirmed `cli_compare_helpers.py`'s compare project-config load always raises `UsageError` on parse failure, explicit or auto-discovered — the explicit-vs-auto-discovered leniency is specific to the *separate* `compile:` block load (`merge_compile_config` in `cli_options.py`, shared by `compare`/`dump`/`scan`'s L2 compile context). Rewrote the section to name which loader each rule applies to instead of implying it uniformly.

---

napetrov:
Four more findings, all verified against source and fixed in c19a389:

- **`config-file.md`**: confirmed `targets:`/`bundles:`/`profiles:`/`baseline:` are recognized top-level keys in `BuildConfig._KNOWN_TOP_KEYS` but absent from `_KNOWN_BLOCK_KEYS`, so `from_dict()`'s `_scalar_findings` never inspects their contents — `compare`'s hard-error rule stops at "is this a known top-level key," not the block's contents. Narrowed the claim to what `BuildConfig.from_dict` itself actually checks; only `project validate`/`project plan` validate the contents.
- **`architecture.md`**: the intro paragraph still said all three platforms have "implemented debug-info cross-check support," contradicting the corrected diagram/DWARF section further down the same page. Fixed to match.
- **`cli-usage.md`/`tool-modes.md`**: confirmed in `service.py` that only `_dump_elf` calls `notify()` with a no-headers warning — `_dump_pe`/`_dump_macho` have no equivalent. Scoped the "prints a warning" claim to ELF.
- **`github-action.md`**: added the same `--frontend-context device` exception documented in `header-backend-capabilities.md`, linking back to it.

---

napetrov:
Three more findings, all verified and fixed in 428da01:

- **Changelog fragment**: repeated the same "auto-discovered config warns and falls back" overgeneralization already fixed in `config-file.md` itself. Scoped it to name the `compile:`-block loader specifically.
- **`--header` help text**: confirmed `_diff_macho` (`diff_platform.py`) has seven sub-detectors beyond exports (install name, arch drift, compat version, dependencies, reexports, loader facts, weak exports), all fed from `_dump_macho()`'s persisted load-command metadata — "export-only" understated it. Reworded to "binary metadata: exports plus load-command facts", matching what the architecture page already says. Regenerated `cli-reference.md`.
- **`config-file.md`**: confirmed via grep across `cli.py`/`cli_options.py`/`cli_scan.py` that only `--config` is a real registered option — `--build-config` only appeared in a docstring, never as a CLI flag. Removed the incorrect second spelling.

---

napetrov:
Two more findings, verified and fixed in 714de1b:

- **Device-context exception scope**: confirmed via `test_header_ast_parser_env_pinned_castxml_rejects_device_context` that `ABICHECK_AST_FRONTEND=castxml` pins `auto` the same as an explicit `--ast-frontend castxml` — it still rejects a device context instead of routing to clang. Qualified the exception in `header-backend-capabilities.md` (and the `github-action.md` summary) to unpinned `auto` only.
- **L0-only scope**: confirmed `service_input_resolution.resolve_side_snapshot` still calls `embed_side_build_source` after binary/debug-info resolution, so `--sources`/`--build-info` can attach L3–L5 evidence to a headerless Mach-O input regardless of platform. Scoped `architecture.md`'s claim to the binary/debug-info evidence specifically, not the whole scan.

---

napetrov:
One more finding, verified and fixed in 0a614d6:

- **`tool-modes.md`**: the Headers requirement row said the headerless fallback is "symbols-only mode" unconditionally. Same evidence as the earlier `cli_options.py` fix — `_dump_macho()`/`_diff_macho()` retain and diff install-name/dependency/arch/compat-version/reexport/rpath/loader metadata with no headers, so for Mach-O the fallback is binary metadata (exports + load-command facts), richer than a symbol list. Reworded to distinguish ELF/PE's genuine symbols-only fallback from Mach-O's.

---

napetrov:
Two more findings, verified and fixed in 9206cc4:

- **Changelog L0 claim**: repeated the same "headerless Mach-O scan is always L0 only" overclaim already scoped correctly in `architecture.md`. Scoped it to match.
- **Agent Skills banner**: confirmed against the page's own "Installing them" section that installation (copying a directory) is version-independent — only execution is gated by each skill's own preflight check. Reworded the banner to distinguish install-time from run-time so it doesn't contradict the page's later instructions.

---

napetrov:
Three more findings, verified and fixed in 2d8a271:

- **"Unpinned" definition**: confirmed `_resolve_single_ast_backend` only rejects a device context when the literal backend argument is `castxml` or the env var pins it — an explicitly spelled `--ast-frontend auto` still routes to clang the same as the bare default. Reworded "unpinned" to mean "not pinned to castxml" in both `header-backend-capabilities.md` and `github-action.md`.
- **Missing fallback trigger**: confirmed `dumper._castxml_fallback_reason()` has a third path — `UnsupportedCastxmlVersionError` → `castxml-unsupported-version` — independent of the toolchain-mismatch/direct-include-guard pair already listed. Added it as a third trigger in both pages.
- **Skill version range**: confirmed every `skills-src/*/SKILL.md` declares `abicheck-version-range: ">=0.6.0,<0.7.0"`, a bounded range, not "0.6.0 or later". Reworded the `agent-skills.md` banner accordingly.

---

napetrov:
One more finding, verified and fixed in 277ce41, plus the same fix applied proactively where I found it repeated:

- **`tool-modes.md`**: confirmed `diff_platform.py`'s `_diff_elf`/`_diff_pe` cover SONAME/dependencies/rpaths (ELF) and machine type/imports/delay-load/hardening/version-floor (PE) independently of headers/debug info — "symbols only" understated the headerless fallback the same way the earlier Mach-O finding did. Reworded to "L0 binary-metadata analysis" with per-platform detail.
- **`cli-usage.md`**: found the identical "falling back further to symbols-only analysis" claim in two places, not flagged this round but the same bug. Fixed proactively for consistency.

---

napetrov:
Two more findings, verified and fixed in 375b661:

- **Agent Skills version check**: confirmed every skill's own preflight (`skills-src/*/SKILL.md`) runs `abicheck --version` — the executable on `PATH`, not a source checkout's `pyproject.toml`, which can differ. Reworded the banner to name the actual check.
- **ADR exemption overclaim**: confirmed by reading `scripts/adr_status_sync.py` that only `Status` (and `Verified`, when present) is machine-checked — `Date`/`Decision maker` are unenforced convention. Corrected `docs/AGENTS.md`'s exemption text to scope "validated" accurately and flag the Date/Decision-maker gap as separate and still open.

---

napetrov:
Two more findings, verified and fixed in 18babce:

- **`cli-usage.md` snapshot workflow**: confirmed `compare`'s `resolve_input()` loads a JSON snapshot verbatim — no headers flag or DWARF/PDB re-resolution happens at compare time for a snapshot input. The generic native-binary no-headers paragraph was misplaced right after an example where both sides were already dumped with headers. Replaced with a snapshot-specific note.
- **Changelog wording**: the opening sentence still described the fallback as degrading straight to "symbols-only", missing the L0 binary-metadata correction from an earlier round. Brought it in line.

---

napetrov:
One more finding, verified and fixed in 72be921 — and a good catch: since this PR registered `action.yml`/`cli_options.py`/`dumper.py` as `ast-frontend-resolution`'s `fact_sources` in an earlier round, those primary sources themselves needed to actually carry the complete, corrected contract, not just the doc pages built from them.

- **`--allow-ast-frontend-fallback`/`--ast-frontend` help text and `action.yml`'s `ast-frontend` input description**: both were still missing the castxml-unsupported-version fallback trigger and the `--frontend-context device` exception the canonical page now documents. Updated both to state all three triggers plus the device exception, so `--help-all`/the GitHub Marketplace listing show the same contract as the docs. Regenerated `cli-reference.md` and `github-action-inputs.md`.

---

napetrov:
One more finding, verified and fixed in 1954482 — my own previous-round fix reintroduced a bug already fixed elsewhere: the new `cli_options.py` help text said the device-context shortcut applies "regardless"/"always", missing the `ABICHECK_AST_FRONTEND=castxml` pin exception (same evidence as before: `test_header_ast_parser_env_pinned_castxml_rejects_device_context`). Qualified both help strings to match `action.yml`'s already-correct phrasing. Regenerated `cli-reference.md`.

---

napetrov:
One more finding, verified against source and empirically, fixed in bfbcc62:

- **Hybrid pin also rejects device context**: confirmed in `dumper.py`'s `dump()` (lines 1265-1278) that a `hybrid`-resolved backend — explicit `--ast-frontend hybrid`, or `auto` pinned via `ABICHECK_AST_FRONTEND=hybrid` — is intercepted *before* `_header_ast_parser` even runs, and rejects a non-`host` `--frontend-context` outright (hybrid has no device concept). Confirmed empirically that `_resolve_header_backend('auto')` returns `'hybrid'` under that env var. My "not pinned to castxml" wording implied a hybrid-pinned `auto` would still route to clang — it doesn't. Reworded all three places (`header-backend-capabilities.md`, `github-action.md`, `cli_options.py` help text) to scope the shortcut to an `auto` that actually resolves to plain castxml.

---

napetrov:
One more finding, verified empirically, fixed in 612ca9d:

- **Env-pinned clang also works**: confirmed with a direct call that `_resolve_header_backend('auto')` returns `'clang'` under `ABICHECK_AST_FRONTEND=clang`, and `_resolve_single_ast_backend('auto', 'device')` returns `'clang'` with no rejection — an env-pinned-to-clang `auto` satisfies the device context exactly like an explicit `--ast-frontend clang`. My "only way to satisfy it" phrasing omitted the env-pin path. Added it alongside the explicit flag.

---

napetrov:
Addressed the last outstanding P0 item (P0.1) in fe4e304 — a site-wide "unreleased docs" banner.

Full stable/dev channel separation (separate URLs, a per-channel Action-pin policy) is real infra work and stays out of scope for this PR, but the minimal, self-maintaining mitigation is in:

- `docs/hooks.py`: `on_config` reads `pyproject.toml`'s own `version` at build time (same regex approach `scripts/gen_repo_facts.py` already uses, since `tomllib` needs Python 3.11+ and this repo floors at 3.10) into `config.extra.latest_published_version` — never a hand-copied string that can go stale.
- `docs/overrides/main.html`: mkdocs-material's documented announcement-bar block, shown on every page, linking to the existing "Upgrading to 0.6" page.

Verified the banner actually renders with the correct version by grepping the built `site/` output on several pages, plus `mkdocs build --strict` clean, `check_docs_contract.py` 0 errors, and the docs/mkdocs pytest suite (143 tests) passing.

With this, **all P0 items from the original review are now addressed** in this PR.

---

napetrov:
Two findings, same root cause, fixed together in 854f0a6 — good catch, this would have made the banner itself go stale exactly the way it exists to prevent:

- **Wrong fact owner**: `docs/hooks.py` read `pyproject.toml`'s version directly, which gets bumped ahead of the actual release cut. Switched to `repo_facts.json`'s `latest_release` — the field this repo already maintains decoupled from `project_version` specifically for this purpose.
- **`agent-skills.md`'s hand-typed "0.5.0"**: extended the hook with a narrow `{{ latest_published_version }}` markdown substitution and replaced the hardcoded version with it, so it tracks the same fact-owner value as the announcement bar.

Verified the placeholder actually substitutes correctly in the built HTML with no literal text left over, plus the missing/malformed `repo_facts.json` fallback path.

---

napetrov:
Good P1 catch, fixed in 5a0bdda: `github-action.md`'s `ast-frontend` row had grown, across seven rounds of edge-case fixes in this PR, into a near-complete duplicate of `header-backend-capabilities.md`'s resolution contract — exactly the drift risk `docs/AGENTS.md`'s governing rule (one fact, one canonical owner, everywhere else a short linked summary) exists to prevent, now that that page is this PR's own registered canonical owner. Collapsed the row to a short Action-focused summary + link, and added `summarizes: [ast-frontend-resolution]` to the page's front matter to make the existing `allowed_summaries` registration explicit.

---

napetrov:
`main` advanced past this PR's base (5 PRs merged, including #735 which regenerated `cli-reference.md`/`header-backend-capabilities.md` too), causing a real merge conflict on those two generated files. Merged `main` in and resolved it (f19910e) — both conflicts were the two backends' generated sections drifting apart, auto-mergeable without content loss. Verified:

- `mkdocs build --strict` clean
- `check_docs_contract.py` 0 errors
- Regenerated `cli-reference.md`/`header-backend-capabilities.md` again post-merge — no diff, confirming the merge result matches what the generators would produce
- `check_ai_readiness.py`'s 13 errors are pre-existing on `origin/main` itself (confirmed via a clean worktree), unrelated to this PR
- 199 targeted tests pass

---

napetrov:
Two more findings, fixed in d2b3e9d:

- **Banner told users to check `pip show abicheck`** — the same environment-mismatch class `agent-skills.md`'s banner already correctly avoids. Switched to `abicheck --version`, matching that page's phrasing. Also fixed the template's own comment, which still described the hook as reading `pyproject.toml` (stale since round 15's fix).
- **Changelog fragment** still said `docs/hooks.py` reads `pyproject.toml`'s version, contradicting the actual (round 15) fix. Updated to name `repo_facts.json`'s `latest_release`.

---

napetrov:
One more finding, verified and fixed in 41e1baa, plus the same fix applied proactively:

- **BTF/CTF fallback missing**: confirmed in `abicheck/dumper_debug.py` that ELF debug-info resolution is `DWARF > BTF > CTF` for userspace binaries (kernel binaries prefer BTF), all converted into the same checker-consumed metadata — `README.md` already lists all four formats but `architecture.md`'s L1 section and diagram only named DWARF/PDB. Fixed, and applied the same correction to `getting-started.md`, `cli-usage.md`, and `tool-modes.md`, which inherited the identical simplification from earlier rounds of this PR.

---

napetrov:
One more finding, fixed in 581ead7 — a good one: the banner's own "Upgrading to 0.6" link hardcoded a version-specific destination that would itself go stale once 0.6.0 ships and that guide gets renamed/rewritten for the next boundary. Switched to linking `CHANGELOG.md` on GitHub, which is version-neutral by construction and needs no per-release upkeep. Confirmed the link renders correctly in the built HTML.

---

napetrov:
One more finding, fixed in 49ad47c: `docs/hooks.py` silently degraded to `"unknown"` when `repo_facts.json` was missing/malformed — since `.github/workflows/pages.yml` runs `mkdocs build --strict` directly, that would still succeed and deploy a site-wide `vunknown` banner, exactly the kind of misinformation this mechanism exists to prevent. Added `ReleaseFactError`, raised for a missing file, unreadable/malformed JSON, or an empty `latest_release` field. Verified empirically both directions: removing `repo_facts.json` makes the build exit 1 with a clear traceback, and with the real file present the build still succeeds cleanly.

---

napetrov:
One more finding, verified and fixed in b5997f8: the device-context exception only covered the shared `--ast-frontend` flag/env pin, not `compare`'s per-side `--old-ast-frontend`/`--new-ast-frontend` overrides. Confirmed in `cli_resolve.py`: `_side_compile()` replaces that side's own `CompileContext.frontend` with the literal override value, which `_resolve_single_ast_backend` then treats exactly like the shared flag — `--old-ast-frontend castxml` rejects a device context on the old side alone, independent of the new side or the shared flag. Added a note covering the per-side case.

---

napetrov:
CodeRabbit came out of rate-limit and posted 7 real findings — all verified against source and fixed in a74aa5e:

- **`action.yml` device-context matrix incomplete**: missing the hybrid-pin-rejects and clang-pin-satisfies cases already on `header-backend-capabilities.md`. Added both — and to `cli_options.py`'s own `--ast-frontend` help text (the shared source `compare`/`dump`/`scan` rows in `cli-reference.md` regenerate from), which had the hybrid case but not the clang-pin case.
- **`config-file.md` auto-discovered leniency overclaim**: confirmed `cli_buildsource.embed_build_source()` (used by `--sources`/`--build-info`) has its own separate config resolution that raises a hard `click.UsageError` unconditionally — explicit or auto-discovered alike — unlike `merge_compile_config()`'s `compile:`-block loader. Documented the distinction.
- **"ELF may add DWARF" missing BTF/CTF**: `cli_options.py`, `README.md`, and `architecture.md`'s summary sentence all lacked BTF/CTF, inconsistent with the detailed L1 section from an earlier round. Aligned all three; regenerated `cli-reference.md`/`github-action-inputs.md`.
- **`docs/hooks.py` didn't validate JSON shape**: a JSON list/scalar `repo_facts.json` would raise a raw `AttributeError` instead of `ReleaseFactError`, and a non-string `latest_release` would publish an invalid `v...` banner via `str(value)`. Added both checks plus `tests/test_docs_hooks.py` (9 tests covering the real file and every malformed shape).
- **`tool-modes.md` "Mode Overview" table** (a different row from the one fixed earlier) still said "DWARF/PDB layout"/"else symbols-only". Fixed to match the detailed contract.
- **`getting-started.md`** implied header AST is available for headerless Mach-O. Reworded to state it needs `-H` on any platform.
- **`mkdocs.yml`'s own comment** still said `pyproject.toml` — the parallel comment in `main.html` was fixed in an earlier round, this one was missed.

Verified: `mkdocs build --strict` clean, `check_docs_contract.py` 0 errors, ruff/mypy clean, 9/9 new tests pass.

---

PR #738, opened 2026-08-12
perf: dedup + parallelize S2 preprocessor pre-scan for scan --depth build

napetrov:
## Summary

Fixes the reported `scan --depth build` 4m→20m wall-time jump on a real-world oneDAL build by deduping and parallelizing the S2 preprocessor pre-scan tier.

## Motivation

Investigation traced the jump to `buildsource/preprocessor_scan.py`'s `capture_macros`: L3 build evidence itself was cheap (~1s to ingest; `bazel aquery` ~7s), but L3 unlocks the S2 preprocessor tier, which shelled out to `clang -E -dM` **once per compile unit, serially, with no cap and no dedup**. On the reported build (~2000 TUs) this alone accounted for ~920s of the ~920+37+230s (`preprocessor_scan`/`pattern_scan`/`candidate_snapshot`) that turned a 4-minute scan into 20 minutes — for zero findings (advisory-only by design). There was also no knob to keep L3 but skip S2.

## Changes

- **Dedup by compile context.** `capture_macros`/`capture_header_includes` now probe each distinct `(language, cwd, flag-set)` signature **once** and fan the result out to every compile unit sharing it, instead of one `clang` invocation per TU. Safe for the curated ABI-macro list (`NDEBUG`, `_GLIBCXX_USE_CXX11_ABI`, …), which is driven by command-line/predefined macros, not TU source text — documented as an accepted approximation in the module docstring.
- **Parallel probing.** Probes run concurrently via a thread pool, sized by the new `ABICHECK_PREPROCESSOR_SCAN_JOBS` env var (mirrors the existing `ABICHECK_L4_JOBS` convention). `runs_attempted`/`runs_ok` counter increments are now lock-protected, since they can now be hit concurrently from multiple worker threads (a plain `+= 1` is not atomic).
- **A cap.** `ABICHECK_PREPROCESSOR_SCAN_MAX_PROBES` (default 512) bounds worst-case cost on a build with an unusually large number of distinct compile contexts; truncation is reported in diagnostics/coverage, never silent.
- **A disable knob.** `ABICHECK_PREPROCESSOR_SCAN=0` skips the whole S2 tier while keeping L3 for other tiers — reported as an honest `not_collected` coverage row, same as a missing compile DB or missing `clang`.
- Documents the new knobs in `docs/contribute/performance.md` alongside the existing L4 perf knobs.

## Checklist

- [x] Tests added / updated for new behaviour (7 new tests: dedup, per-signature probing, jobs env override, cap + truncation reporting, disable knob, concurrency race guard)
- [x] Docs updated (`docs/contribute/performance.md`)
- [x] `ruff check`/`ruff format --check` pass locally
- [x] No new `mypy` errors introduced (`mypy abicheck/` clean)

---
_Generated by [Claude Code](https://claude.ai/code/session_01FuPUxGdg5mkXkqmN7CbmBc)_

### DIFF
--- a/abicheck/buildsource/call_graph.py
+++ b/abicheck/buildsource/call_graph.py
@@ -44,7 +44,7 @@ from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field, replace
 from functools import partial
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, TypeVar
 
 from .. import deadline
 from ..build_context import _extract_flags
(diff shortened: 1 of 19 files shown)


napetrov:
Re: [the topics.yaml registration finding](https://github.com/abicheck/abicheck/pull/738#discussion_r) on `docs/contribute/performance.md` — not applying this one, explaining why rather than silently skipping it.

`docs/_meta/topics.yaml`'s registry is scoped to end-user product surfaces (evidence model, verdicts, baseline management, policies, output formats, GitHub Action, …) — every existing topic's `canonical_page` lives under `docs/learn/`, `docs/use/`, or `docs/reference/`. No topic anchors on a `docs/contribute/*` page; that tree is contributor/maintainer-facing internals, not the audience this registry tracks.

`docs/contribute/performance.md` itself already documents an equivalent, pre-existing class of surface — L4's `ABICHECK_L4_JOBS`/`ABICHECK_L4_EXECUTOR`/`ABICHECK_L4_CACHE_DIR`/`ABICHECK_L4_JOB_MEM_GIB` env-var perf knobs — and none of those were ever registered as a topic. The new `ABICHECK_PREPROCESSOR_SCAN*` knobs and `probes_truncated` report field added here are the same category: advisory-only tuning surface for an internal, never-authoritative scan tier, documented alongside its direct precedent on the same page.

Given that established pattern and that this isn't CI-enforced (`check_docs_contract.py` validates the registry's own internal consistency, not "does every new field have an entry"), registering a topic here would be inconsistent with the page's existing content rather than filling a real gap — it would mean either registering this one narrow knob set while leaving its L4 sibling (and the rest of `preprocessor_scan`'s pre-existing, already-unregistered report fields) out, or taking on registering that whole page as an out-of-scope side project for this PR. Leaving as-is.

---

PR #751, opened 2026-08-13
fix: populate BuildEvidence.targets on zero-config Bazel --sources scans

napetrov:
## Problem

`buildsource/build_query.py`'s zero-config `--sources` auto-inference path only ever ran `bazel aquery` for a detected Bazel workspace. `BuildEvidence.targets` is populated *only* by `BazelAdapter`'s cquery-processing method (`_collect_cquery`) — never by its aquery one (`_collect_aquery`, which only ever fills `compile_units`/`link_units`) — so a plain `dump --sources` / `compare --sources` against a Bazel checkout always reported zero resolved targets, no matter what the workspace actually contained.

This is exactly the gap [`abicheck-bazel-lab`](https://github.com/napetrov/abicheck-bazel-lab)'s `build_bazel_evidence_pack.py` script was built to work around for the explicit `--build-info` path (by constructing a `BazelAdapter` directly with both pre-captured cquery and aquery files) — that workaround is no longer necessary for the zero-config path once this lands.

## Fix

- `inferred_bazel_cquery_command()` — a fixed, abicheck-authored `bazel cquery --output=jsonproto deps(//...)`, scoped identically to the existing aquery command (same `deps(//...)` universe), so the two stay consistent.
- `_run_bazel_cquery_supplement()` — runs it as a second, independent subprocess alongside the existing aquery call in `run_inferred_build_query()`.
- Both results are merged through **one** `BazelAdapter(aquery=..., cquery=..., ...).collect()` call, mirroring how an explicit `--build-info` bazel-cquery/-aquery pair already combines both.

The cquery run is a pure supplement, never the primary signal:
- A missing launcher, timeout, non-zero exit, or parse failure only appends a diagnostic to `BuildEvidence.diagnostics` and leaves `targets` empty (the pre-existing behavior) — it never demotes the aquery-derived compile/link-unit ingest's own extractor status from `ok`/`partial` to `failed`.
- No `--include_param_files` on the cquery command (unlike aquery): cquery's jsonproto already carries each target's fields inline, unlike aquery's action argv which can spill to `@params` files.

## Testing

- Two new regression tests in `tests/test_build_query_inference.py`:
  - `test_run_bazel_populates_targets_from_cquery_supplement` — end-to-end, both aquery and cquery succeed, `BuildEvidence.targets` is populated.
  - `test_run_bazel_cquery_supplement_failure_does_not_demote_aquery_status` — cquery fails, aquery still succeeds, extractor status stays `ok`, `targets` stays empty, failure recorded only as a diagnostic.
  - Plus a command-shape test for `inferred_bazel_cquery_command()`.
- All 60 tests in `tests/test_build_query_inference.py` pass, including every pre-existing test (none needed changes — the existing single-response `fake_run` mocks return the same fixture for both the aquery and cquery calls, which the cquery-side parser degrades gracefully on).
- `tests/test_buildsource_inline_nodb.py` and `tests/test_bazel_adapter.py` pass (142 total across the three files).
- `mypy abicheck/` — clean (360 files).
- `ruff check` / `ruff format --check` — clean.
- `scripts/check_ai_readiness.py` — 0 errors.
- Changelog fragment added via `scriv create`.

---
_Generated by [Claude Code](https://claude.ai/code/session_019S6MjuZW7ineG3bboJTGBq)_

## Summary by CodeRabbit

- **New Features**
  - Improved zero-configuration Bazel source scans with supplemental target discovery.
  - Scan results now combine target information with existing compile and link evidence.
  - Added support for Bazelisk when running supplementary discovery.

- **Bug Fixes**
  - Supplementary discovery failures no longer reduce the status of successful primary scan results.
  - Improved timeout handling ensures scans respect the available execution time.

- **Documentation**
  - Added release documentation covering the enhanced Bazel scanning behavior.

### DIFF
--- a/abicheck/buildsource/build_query.py
+++ b/abicheck/buildsource/build_query.py
@@ -814,6 +814,7 @@ def run_inferred_build_query(
             return None
         if not _resolve_query_launcher(cmd, system, which, extractors):
             return None
+        query_start = time.monotonic()
         proc = _run_query_process(cmd, system, sources, timeout, merged, extractors)
         if proc is None:
             return None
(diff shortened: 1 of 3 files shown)


napetrov:
CI red on **Native PE/Mach-O compare workflows (windows-latest)**, but it's failing on infra, not this PR's diff: `choco install mingw` fails with `mingw not installed. The package was not found with the source(s) listed.` — a Chocolatey community-feed availability issue, unrelated to the Python-only `build_query.py`/test change here (no Windows/mingw dependency in this diff). Will re-run once the base is confirmed green, or address separately if it turns out to reproduce on `main`.

---

PR #754, opened 2026-08-14
fix(clang-plugin): refuse unsafe same-major apt fallback for Intel icx/icpx

napetrov:
## Summary

Correct `producer: clang-plugin`'s behavior for Intel's oneAPI DPC++/C++ Compiler (`icpx`/`icx`) based on a real, end-to-end evaluation (Intel oneAPI 2026.1.1, Clang 22 frontend) that found the plugin's existing same-major apt-fallback build path produces an ABI-mismatched, crashing artifact for this compiler specifically — not a merely-degraded one.

## Motivation

Real testing showed:
- A plugin built with RTTI enabled fails to load into `icpx` (`undefined symbol: _ZTIN5clang15PluginASTActionE`).
- A plugin built against libstdc++ crashes (exit 139) once it reads its own arguments, because `icpx`'s frontend stack uses libc++ containers across the same Clang plugin ABI boundary (`ParseArgs`'s `std::vector<std::string>`).
- Linking `libclang-cpp`/`libLLVM` into the plugin as a workaround instead produces a different crash (duplicate LLVM command-line-option registration).

`actions/collect-facts/run.sh` previously fell back to installing a same-major `clang-$major`/`llvm-$major-dev`/`libclang-$major-dev` apt package for *any* compiler reporting a matching `__clang_major__`, Intel's fork included — silently attempting a build now known to produce a broken artifact.

## Changes

- `actions/collect-facts/run.sh`: detect the Intel oneAPI fork via the `__INTEL_LLVM_COMPILER` predefined macro (`_is_intel_llvm_compiler`) and refuse the same-major apt fallback for it specifically. Add a `plugin-artifact`/`plugin-artifact-sha256` input pair to use an already-built, already-certified plugin binary directly — the recommended distribution model (build once per exact toolchain fingerprint, publish the artifact, commit only a digest/lock in client repos). When building against a genuine vendor-bundled Intel LLVM/Clang SDK (`llvm-cmake-prefix`), pass the `-fno-rtti -stdlib=libc++` recipe real testing found necessary, and mark `producer-version` as `+experimental-intel-llvm`.
- `contrib/abicheck-clang-plugin/CMakeLists.txt`: add explicit `ABICHECK_PLUGIN_RTTI` (`auto|on|off`) and `ABICHECK_PLUGIN_STDLIB` (`auto|libc++|libstdc++`) overrides. `auto` is a no-op, so every existing vanilla-Clang build is unaffected.
- Document the findings, the certified-artifact distribution model, and the experimental/not-certified status in `contrib/abicheck-clang-plugin/README.md` and `AGENTS.md`, and in `docs/use/producing-source-facts.md`.
- `tests/test_action_collect_facts.py`: cover `_is_intel_llvm_compiler`, the apt-fallback refusal (and that it doesn't misfire for vanilla Clang), and the `plugin-artifact` path (success, sha256 mismatch, missing file).

Default behavior for `dump --sources` (Full source scan) and the `abicheck-cc` wrapper — the portable, always-supported ICX paths — is unchanged.

## Checklist

- [x] Tests added / updated for new behaviour (`tests/test_action_collect_facts.py`, 92 tests pass locally)
- [x] Changelog fragment — not added: this PR touches no `abicheck/**/*.py` files (shell/CMake/Action config/docs/tests only), so `changelog.d/README.md`'s trigger doesn't apply
- [x] Docs updated (`contrib/abicheck-clang-plugin/README.md`, `AGENTS.md`, `docs/use/producing-source-facts.md`)
- [x] `ruff check`/`ruff format --check` pass on the touched Python test file; `bash -n` and a YAML parse pass on the shell/Action changes
- [x] No new `mypy`/`ruff` errors introduced

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01SiMc6SKcRKwHQZuYb6VG68)_

## Summary by CodeRabbit

* **New Features**
  * Added experimental support for Intel oneAPI `icx` and `icpx` toolchains.
  * Added support for using prebuilt Clang plugins with optional SHA-256 verification.
  * Added configurable RTTI and standard-library compatibility settings for plugin builds.

* **Bug Fixes**
  * Prevented unsafe fallback to potentially incompatible system packages for Intel compilers.

* **Documentation**
  * Added guidance for Intel toolchain setup, compatibility requirements, certified plugin artifacts, and fallback options.

### DIFF
--- a/actions/collect-facts/action.yml
+++ b/actions/collect-facts/action.yml
@@ -81,14 +81,48 @@ inputs:
       Compiler used for the build (informational for replay/wrapper; for
       clang-plugin this selects which clang the plugin is built against and
       must match the clang that later loads it via -fplugin=). Vendor
-      compilers built on top of Clang are supported here too (e.g. Intel's
+      compilers built on top of Clang are recognized here too (e.g. Intel's
       icpx/icx) -- the real LLVM major is read from the compiler's own
       __clang_major__ macro, not parsed out of --version (which for icpx/icx
-      prints a vendor product version, not an LLVM number). See
-      llvm-cmake-prefix below for pointing the plugin build at that vendor's
-      bundled LLVM/Clang dev files.
+      prints a vendor product version, not an LLVM number), and the compiler
+      is separately probed for __INTEL_LLVM_COMPILER to detect the Intel
+      oneAPI DPC++/C++ Compiler fork specifically. See llvm-cmake-prefix and
+      plugin-artifact below -- for that fork, building against a same-major
+      apt/distro Clang development package is refused (real testing found
+      the result an ABI mismatch that crashes rather than merely fails to
+      load; see contrib/abicheck-clang-plugin/README.md's Intel oneAPI
+      section), so one of those two inputs is required instead.
     required: false
     default: 'clang++'
+  plugin-artifact:
+    description: >
+      clang-plugin only: path to an already-built, already-certified
+      libabicheck-facts.* shared object to use directly, skipping the CMake
+      build entirely. This is the recommended path for a vendor fork whose
+      plugin-development SDK isn't available in this job (Intel's icpx/icx
+      chief among them) -- build once against the exact toolchain image,
+      publish the artifact (a GitHub Release asset, an OCI artifact, or a
+      layer in the same pinned image as the compiler), and have the calling
+      workflow resolve/download it before this Action runs. The artifact is
+      still smoke-tested against the resolved `compiler` before use (same
+      check a from-source build gets), so a stale or mismatched artifact
+      fails loudly here rather than crashing or misbehaving inside your real
+      build. Pair with plugin-artifact-sha256 to pin the exact binary a
+      client repository's lock manifest names, rather than trusting whatever
+      the download step happened to fetch.
+    required: false
+    default: ''
+  plugin-artifact-sha256:
+    description: >
+      Expected sha256 digest of plugin-artifact. When set, the artifact's
+      digest is checked before it is smoke-tested or used, and a mismatch
+      fails the step -- this is the "lock manifest" half of the certified-
+      artifact distribution model (see contrib/abicheck-clang-plugin/
+      README.md): a client repository commits this digest (plus the
+      resolved compiler fingerprint), not the binary itself. Ignored when
+      plugin-artifact is not set.
+    required: false
+    default: ''
   install-deps:
     description: >
       Install system dependencies needed by the selected producer (clang/
@@ -201,6 +235,8 @@ runs:
         INPUT_LIBRARY: ${{ inputs.library }}
         INPUT_EXTRACTOR: ${{ inputs.extractor }}
         INPUT_COMPILER: ${{ inputs.compiler }}
+        INPUT_PLUGIN_ARTIFACT: ${{ inputs.plugin-artifact }}
+        INPUT_PLUGIN_ARTIFACT_SHA256: ${{ inputs.plugin-artifact-sha256 }}
         INPUT_INSTALL_DEPS: ${{ inputs.install-deps }}
         INPUT_LLVM_CMAKE_PREFIX: ${{ inputs.llvm-cmake-prefix }}
         ACTION_PATH: ${{ github.action_path }}
--- a/actions/collect-facts/run.sh
+++ b/actions/collect-facts/run.sh
@@ -91,6 +91,8 @@ PUBLIC_ROOTS="${INPUT_PUBLIC_ROOTS:-}"
 LIBRARY="${INPUT_LIBRARY:-}"
 EXTRACTOR="${INPUT_EXTRACTOR:-auto}"
 COMPILER="${INPUT_COMPILER:-clang++}"
+PLUGIN_ARTIFACT="${INPUT_PLUGIN_ARTIFACT:-}"
+PLUGIN_ARTIFACT_SHA256="${INPUT_PLUGIN_ARTIFACT_SHA256:-}"
 INSTALL_DEPS="${INPUT_INSTALL_DEPS:-true}"
 LLVM_CMAKE_PREFIX="${INPUT_LLVM_CMAKE_PREFIX:-}"
 ACTION_PATH="${ACTION_PATH:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)}"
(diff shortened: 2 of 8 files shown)


napetrov:
Pushed a follow-up commit addressing the Codex/CodeRabbit findings:

- Resolved `plugin-artifact` to an absolute path before exporting it (Codex, P1) — a relative path stayed relative through `ABICHECK_PLUGIN_SO`/`ABICHECK_PLUGIN_FLAGS`, so the caller's real build could fail to find it.
- Pinned the plugin's CMake configure to `-DCMAKE_CXX_COMPILER="$compiler_path"` (CodeRabbit, Major) — without it, CMake could silently configure against a different compiler than the one this script resolved.
- Shortened the Intel oneAPI warning in `docs/use/producing-source-facts.md` to link back to `contrib/abicheck-clang-plugin/README.md` as the canonical source instead of duplicating it (CodeRabbit, docs contract).

New tests cover both fixes; the full `tests/test_action_collect_facts.py` suite (94 tests) passes locally.


---

napetrov:
Re the "keep one canonical Intel LLVM explanation" comment on `docs/use/producing-source-facts.md`: a topic already exists for this — `source-evidence-producers` in `docs/_meta/topics.yaml` has `canonical_page: use/producing-source-facts.md` with `contrib/abicheck-clang-plugin` registered as one of its `fact_sources`. So `producing-source-facts.md` is already the intended canonical narrative home, and the README is correctly a supporting fact source rather than a second canonical page — `canonical_page` also has to be a `docs/`-relative path per the registry's own schema, so pointing it at the contrib README isn't an option anyway.

I did shorten the warning to usage-specific guidance and link back to the README for the detailed findings, per the actual ask. `scripts/check_docs_contract.py` passes with 0 errors on this change. I don't think a new/changed topic registration is needed here since the existing one already covers it — resolving.


---

napetrov:
Pushed another follow-up (`2fe5609`) for the two latest Codex findings:

- `contrib/abicheck-clang-plugin/CMakeLists.txt`: `ABICHECK_PLUGIN_RTTI`/`ABICHECK_PLUGIN_STDLIB`'s cache `STRINGS` property is GUI-only, not enforced by CMake — a case mismatch (`-DABICHECK_PLUGIN_RTTI=OFF`) or typo silently matched neither branch. Now normalized via `string(TOLOWER ...)` and rejected with `message(FATAL_ERROR ...)` if still unrecognized.
- `docs/_meta/topics.yaml`: registered `actions/collect-facts/action.yml` as a `fact_source` on the existing `source-evidence-producers` topic, per `docs/AGENTS.md`'s "every new Action input must register a topic" rule.

`tests/test_action_collect_facts.py` (94 tests) and `scripts/check_docs_contract.py` (0 errors) both still pass.


---

napetrov:
The `Validate examples (clang, debug-headers, shard 3/3)` failure is a transient GitHub Actions infra issue, not caused by this PR — it failed while downloading the `actions/checkout` action itself, before any of this PR's code ran or even the repo was checked out:

```
##[error]An action could not be found at the URI 'https://codeload.github.com/actions/checkout/tar.gz/d23441a48e516b6c34aea4fa41551a30e30af803'
##[error]Failed to download archive ... after 1 attempts.
```

No action needed here; will re-check when it (hopefully) recovers on retry.


---

napetrov:
Update: several more jobs (`Clang header fallback regressions`, `C.6 + scan (LLVM 17, host g++)`, and likely others) are failing the same way — `codeload.github.com` returning 500/429 while GitHub Actions tries to download the `actions/checkout` action itself, before any job code runs:

```
##[warning]Failed to download action ... 500 (Internal Server Error) / 429 (Too Many Requests)
##[error]An action could not be found at the URI ...
```

This is a GitHub Actions infrastructure issue affecting action downloads broadly, unrelated to this PR's diff. Will check back once it clears and re-run/verify affected jobs.


---

napetrov:
Re the two latest Codex findings on `actions/collect-facts/run.sh`:

1. **"Verify that the smoke test actually emits facts"** — fixed in `e425a75`. The smoke test now requires a real TU record in its scratch `out=` directory (same `validate_inputs_pack`/`tu_count` check `phase: verify` already runs on the real pack), so a wrong/stale artifact that dlopens but never registers the `abicheck-facts` action now fails at smoke-test time instead of silently passing. Confirmed safe to require unconditionally by reading `AbicheckFactsPlugin.cpp`: `FactsConsumer::HandleTranslationUnit` calls `writeTu()`/`ensureManifest()` for every TU with no `publicCount` gate, so even the trivial smoke source always produces a real record under the genuine plugin.

2. **"Quote artifact paths in the exported plugin flags"** — not fixed in this PR. This is pre-existing behavior (the flat, unquoted `plugin_flags` string construction predates this PR; I only moved it into the shared `_finish_clang_plugin` tail unchanged) and applies identically to `$OUTPUT`, `public-roots=`, and `library=` today, not just the new `plugin-artifact` path — README.md's own documented usage (`-DCMAKE_CXX_FLAGS="$ABICHECK_PLUGIN_FLAGS"`) already relies on the current unquoted convention. A correct general fix needs a decision on representation (shell-quoted string vs. a different mechanism like a response file) that works consistently across CMake/Make/Bazel consumers, which is a bigger, separate change touching every exported flag, not a one-line quote around `plugin_so`. Leaving as a known, pre-existing limitation rather than a partial fix scoped to only the new input.

`tests/test_action_collect_facts.py` (95 tests) passes.


---

napetrov:
Good catch — fixed in `8493a6e`. The smoke-test validator now checks `report.errors` too, not just `tu_count`, mirroring `_verify_pack`'s own check on the real pack. Added a regression test that triggers `report.errors` (duplicate `tu_id`s) alongside a positive `tu_count` to cover exactly this case. `tests/test_action_collect_facts.py` (96 tests) passes.


---

napetrov:
Good find — this was a real bug undermining the whole PR's safety guarantee. Fixed in `cc6036b`.

`_is_intel_llvm_compiler` piped the already-fully-captured `$defines` string through `printf ... | grep -q ...`. Under this script's global `set -o pipefail`, `grep -q`'s early exit on match can SIGPIPE the upstream `printf` on a large macro dump before it finishes writing, making the pipeline report failure even though the match was real — silently taking the "false" branch and defeating the Intel-fork refusal entirely. Since `$defines` is already an in-memory bash string (from the command substitution just above), no pipe was ever needed — now matched directly with a bash glob pattern.

Verified the fix is real, not just plausible: reverted to the old `grep -q` implementation locally and confirmed the new regression test (large padded macro dump, match placed early so `grep -q` exits well before the writer finishes) fails reliably (5/5 runs) against it, then confirmed 5/5 passes against the fix. `tests/test_action_collect_facts.py` (97 tests) passes.


---

napetrov:
`unit-tests (windows-latest, 3.13, false)` failed for real — fixed in `769a0c9`:

`test_relative_plugin_artifact_is_resolved_to_absolute` compared `ABICHECK_PLUGIN_SO` against Python's `str(Path)`, which is backslash-separated on native Windows Python. But `run.sh` resolves a relative `plugin-artifact` via `_native_pwd`, which normalizes to the forward-slash/`cygpath -m` form on Windows Git Bash — the same distinction `_native_abspath()`'s own docstring in this test file already documents for `OUTPUT`. My test just used the wrong helper. Fixed by using `_native_abspath(...)` like every other path assertion in the file. The actual `run.sh` behavior was already correct; this was a test-only bug.

97 tests pass locally (Linux); windows-latest re-run pending.


---

PR #759, opened 2026-08-14
fix: close canonical_finding_id backend-dependence for three type-slot kinds

napetrov:
## Summary

`canonical_finding_id` (introduced by #753) is still backend-dependent for `ATOMIC_QUALIFIER_CHANGED`, `CHAR8T_MIGRATION`, and `BIT_INT_WIDTH_CHANGED`, breaking the cross-backend `finding_id:` suppression contract for these three kinds.

## Motivation

`canonical_finding_id` is meant to let a suppression rule minted from one header backend's report (e.g. CastXML) reliably match the equivalent finding from another (e.g. Clang), even when the two backends spell a type differently (`"char const*"` vs `"char const *"`).

`diff_atomic.py`, `diff_char8t.py`, and `diff_bit_int.py` all pass `iter_type_slot_changes()`'s raw backend type spellings straight through as `old_value`/`new_value`, and each kind's `description_template` embeds `"{old} → {new}"` mid-sentence after a leading `{detail}` clause — the same shape `struct_field_type_changed` already handles. None of the three kinds was in `_TYPE_BEARING_DISCRIMINATOR_KINDS`/`_DESCRIPTION_EMBEDS_VALUES_KINDS` (`finding_identity.py`), so their raw, backend-specific spelling remained part of the CANONICAL-tier primary ID and changed the resulting hash — this was already called out as a known, accepted gap for `ATOMIC_QUALIFIER_CHANGED` specifically in a code comment, but `CHAR8T_MIGRATION` and `BIT_INT_WIDTH_CHANGED` have the identical data flow and weren't mentioned there.

Concretely: `suppression.py`'s `Suppression._matches_finding_id()` does strict equality against a freshly computed `report_canonical_finding_id()`, with no fallback — so a `finding_id:` rule copied from one backend's report silently fails to match the equivalent finding from the other, producing recurring unsuppressed findings after switching frontend or CI profile.

## Changes

- Added `atomic_qualifier_changed`, `char8t_migration`, and `bit_int_width_changed` to `_TYPE_BEARING_DISCRIMINATOR_KINDS` and `_DESCRIPTION_EMBEDS_VALUES_KINDS` in `abicheck/finding_identity.py`, after verifying each kind's `{detail}` is always a fixed direction/transition string (never a per-item identity like a field or parameter name) — the same safety bar the existing allowlist entries already meet, so this doesn't risk collapsing two distinct findings the way an unconditional sweep would.
- Added parameterized regression tests in `tests/test_canonical_finding_id.py` covering all three kinds:
  1. equivalent CastXML/Clang type spellings collapse to one canonical id;
  2. a `finding_id:` suppression rule minted from one backend's report matches the equivalent finding from the other;
  3. semantically different transitions (opposite atomic direction, different char8_t migration direction, different `_BitInt` width) still produce distinct canonical ids.
- Added a `changelog.d/` fragment.

## Checklist

- [x] Docs updated if user-visible behaviour changed — not applicable (internal identity-hashing fix, no user-facing docs to update)
- [x] `ruff check`, `ruff format --check`, `mypy abicheck/` all pass locally

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01KnTc7WeobEKRgXGS91ZxLt)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of finding identifiers across supported analysis backends for atomic qualifier, `char8_t`, and `_BitInt` width changes.
  * Cross-backend suppression now works reliably when equivalent type differences are reported with varying spellings.

* **Tests**
  * Added regression coverage to verify stable identifiers while preserving distinctions between different type transitions.

* **Documentation**
  * Added a changelog entry describing the backend-stability improvements.

### DIFF
--- a/abicheck/finding_identity.py
+++ b/abicheck/finding_identity.py
@@ -1189,6 +1189,54 @@ def _canonicalize_params_list(value: str) -> str:
     )
 
 
+# Matches an `_Atomic`-qualified spelling in either form castxml/clang emit:
+# a bare `_Atomic` sentinel (castxml, see below) or a real wrapped spelling
+# like `_Atomic(struct Foo *)` (clang). Anchored to the whole string (after
+# stripping) rather than reusing diff_atomic.py's own word-boundary search,
+# since this helper's job is "is this ENTIRE type-slot spelling the atomic
+# side of the transition", not "does _Atomic appear anywhere".
+_ATOMIC_SLOT_RE = re.compile(r"^_Atomic\b")
+
+
+def _canonicalize_atomic_slot(value: str) -> str:
+    """Canonicalize one ``diff_atomic.py`` type-slot spelling for the
+    identity discriminator (Codex review, fresh evidence).
+
+    ``dumper_castxml.py`` cannot model C11 ``_Atomic`` at all -- it emits a
+    bare ``Unimplemented`` node with no reference to the wrapped type, so
+    its own ``_type_name()`` spells the qualified side of the transition as
+    the literal, lossy sentinel ``"_Atomic"`` (see that function's own
+    comment). The direct-clang backend, by contrast, retains the real
+    wrapped spelling clang's own AST printer produces (e.g.
+    ``"_Atomic(struct Foo *)"``). Plain :func:`canonicalize_type_name`
+    normalizes whitespace/const-ordering/pointer-sigil spacing, but has no
+    notion of "these two spellings name the same qualified type" -- so
+    without this helper, CastXML's bare sentinel and Clang's concrete
+    spelling would never hash to the same canonical id for the identical
+    qualifier-added/-removed transition, defeating the whole point of
+    adding ``atomic_qualifier_changed`` to
+    :data:`_TYPE_BEARING_DISCRIMINATOR_KINDS`.
+
+    Safe to collapse the qualified side down to the bare ``"_Atomic"``
+    sentinel for identity purposes: ``diff_atomic.py``'s own detection
+    (``_has_atomic``) only ever tests *presence* of the qualifier via a
+    regex search, never compares the wrapped type across old/new -- two
+    genuinely different ``_Atomic``-wrapped inner types on the same slot
+    can never both reach this detector as two distinct findings (the
+    detector only fires on a presence transition, and continues past an
+    unchanged, still-``_Atomic`` slot entirely) -- so no real discriminating
+    information is lost. The *unqualified* side of the transition (the
+    plain, non-``_Atomic`` type both backends spell reliably) is left to
+    ordinary :func:`canonicalize_type_name` normalization, since that side's
+    exact spelling IS backend-comparable and still needs to distinguish,
+    e.g., an unrelated base-type change from a pure qualifier change.
+    """
+    stripped = value.strip()
+    if _ATOMIC_SLOT_RE.match(stripped):
+        return "_Atomic"
+    return canonicalize_type_name(stripped)
+
+
 def _stringify_change_value(value: object) -> str:
     """Deterministic string form of a ``Change.old_value``/``new_value``.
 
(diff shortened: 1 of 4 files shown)


abicheck/finding_identity.py
```diff
@@ -1082,6 +1094,14 @@ def _is_batch_shaped_change(change: Change, kind_value: str) -> bool:
         "typedef_base_changed",
         "template_param_type_changed",
         "template_return_type_changed",
+        # diff_atomic.py/diff_char8t.py/diff_bit_int.py: old_value/new_value
+        # are iter_type_slot_changes()'s raw type-slot spellings, and
+        # `detail` (folded into `description` via change_registry's
+        # `{detail}` placeholder) is always a fixed direction string, never
+        # a per-item identity -- see this set's own comment above.
+        "atomic_qualifier_changed",
```
napetrov, line 1103: Confirmed and fixed in 296eabd: added `_canonicalize_atomic_slot`, which collapses any `_Atomic`-qualified spelling (bare `"_Atomic"` sentinel or a real wrapped spelling) to a fixed sentinel for identity purposes before it reaches the discriminator — safe because `diff_atomic.py`'s own detection only tests qualifier presence, never compares the wrapped inner type. Wired into `_change_discriminator`'s `atomic_qualifier_changed` branch. Fixing this also surfaced a second, independent bug in the description-substitution logic (raw_old_str being a literal substring of raw_new_str corrupted the replacement) — fixed in the same commit. New regression tests: `test_atomic_qualifier_changed_castxml_lossy_sentinel_matches_clang` and `test_atomic_qualifier_changed_description_substitution_handles_containment`.

---

napetrov (commented):
Addressed all three review findings in 296eabd — see inline reply on the P1 finding for details on the CastXML lossy-sentinel fix and the description-substitution bug it surfaced. Both CodeRabbit quick-wins (bogus BitInt width fixture, missing type annotations) are also fixed.

---

abicheck/finding_identity.py
```diff
@@ -1161,6 +1189,102 @@ def _canonicalize_params_list(value: str) -> str:
...
+    transition, and continues past an unchanged, still-``_Atomic`` slot
+    entirely) -- so no discriminating information about the wrapped
+    content is lost.
+
+    **Not** safe to collapse a declarator applied *outside* the wrapper
+    (Codex review, second round, fresh evidence): ``dumper_castxml.py``'s
+    ``PointerType``/``ReferenceType`` handling appends its own sigil as a
+    suffix onto whatever it wraps (``self._type_name(pointee) + "*"``), so
+    "pointer to atomic T" (``_Atomic(T) *`` on Clang) spells as
+    ``"_Atomic*"`` on CastXML -- distinguishably different from "atomic
```
napetrov: Confirmed and fixed in a84b8e9. Verified the mechanism against `dumper_castxml.py`'s own `PointerType`/`ReferenceType` handling (`self._type_name(pointee) + "*"`): "pointer to atomic T" spells as `"_Atomic*"` on CastXML (the pointer wraps the Atomic node and appends its own suffix), while "atomic pointer to T" spells as the bare `"_Atomic"` sentinel (the Atomic node wraps the pointer, so no outer sigil is ever appended) — these are genuinely different qualified types that must not collapse together.

`_canonicalize_atomic_slot` now locates the `_Atomic(...)` wrapper's own matching close paren (tracking nesting depth for a wrapped type that itself contains parens) and canonicalizes only what follows it, discarding just the wrapped inner content CastXML can't recover while preserving the outer declarator it does retain. New regression test `test_atomic_qualifier_changed_preserves_outer_declarator` verifies both shapes stay distinct and each still matches its own CastXML-spelled equivalent.

---

napetrov (commented):
Addressed the second P1 finding in a84b8e9.

---

napetrov:
Addressed both P1 findings from the last review round in 05f42fa (the code they pointed at moved to a new module, `finding_identity_atomic.py`, so I couldn't reply inline):

1. **Compound payload changes** — `canonicalize_atomic_slot` now takes the *other* (opposite) side's raw spelling and reconstructs the qualified side's full spelling to compare against it: the wrapped payload is only collapsed when that comparison shows it's redundant (the common cross-backend-ambiguous case CastXML forces); when it genuinely differs (`int` → `_Atomic(long)`) the payload is preserved, so it stays distinct from a pure `int` → `_Atomic(int)` qualifier-only change. New test: `test_atomic_qualifier_changed_preserves_compound_payload_change`.
2. **cv-qualified atomics** — a leading `const`/`volatile` prefix is now stripped before atomic detection and reattached (normalized) to the result, so `"const _Atomic"` (CastXML) and `"const _Atomic(int)"` (Clang) collapse to the same canonical id. New test: `test_atomic_qualifier_changed_normalizes_cv_qualified_spellings`.

Also split the growing atomic-slot canonicalization logic into `finding_identity_atomic.py` — it had pushed `finding_identity.py` over the 2000-line hard cap twice in a row.

---

PR #767, opened 2026-08-14
fix: P0.4 follow-up — close review gaps left by merge timing on #763

napetrov:
PR #763 (the P0.4 "first-class `analysis_assurance` model" effort) merged into `main` at `891bd9d7` while its 9th review round was still in flight — a merge-timing race, the same situation PR #766 documents for #762. Two review findings from that round arrived after the merge and are addressed here instead, against `main`.

## Finding 1 (P1) — Propagate the requested depth into assurance

`compute_analysis_assurance()` gates `depth_satisfied` off `DiffResult.requested_depth`, but nothing ever populated that field. An explicit `compare --depth source` that never actually reached source evidence (e.g. both sides lack a compile database, so the effective depth stays `headers`) silently read `requested_depth=None`/`depth_satisfied=None` and could still report `status="complete"`, letting `--require-complete-analysis` exit `0` despite the explicitly requested depth never being reached.

Fixed in `abicheck/cli_compare_helpers.py`'s `_report_compare_result`: it now copies its own Click-validated `--depth` string onto `DiffResult.requested_depth` before recomputing `analysis_assurance` — only from an *explicit* flag, never an inferred depth from bare `--sources`/`--build-info` or `.abicheck.yml`'s `source.method`.

## Finding 2 (P1) — Reject partially overlapping graph coverage

Round 8's fix in `abicheck/analysis_assurance.py`'s `_graph_completeness()` flagged asymmetric confirmed-family coverage only when the two sides' confirmed-family sets were fully **disjoint** (`isdisjoint()`). Fresh evidence: a **partial-overlap** case — old confirms `{call_graph, type_graph}`, new confirms only `{call_graph}` — is not disjoint (they share `call_graph`), so it fell through to `"complete"` even though `buildsource/source_graph_findings.py` only trusts a family when **both** sides cover it, meaning `type_graph` was silently skipped for that comparison.

Fixed by comparing the two sides' confirmed-family sets for any inequality (`!=`) rather than only total disjunction — subset, superset, and partial-overlap pairs are all "the two sides disagree on what they confirmed," same as a fully disjoint pair.

## Testing

- New regression tests for both findings, at the unit level (calling `compute_analysis_assurance`/`_graph_completeness` directly) and end-to-end through the real `compare` CLI.
- Companion tests confirm round 8's original disjoint case, and the "both sides confirm the identical set" case, are unaffected.
- New tests split into a sibling file (`tests/test_analysis_assurance_depth_and_graph_overlap.py`) to keep `tests/test_analysis_assurance.py` under the AI-readiness file-size hard cap.
- `ruff check`, `ruff format --check` (new/changed lines), `mypy abicheck/` (0 errors), `pytest tests/test_analysis_assurance*.py`, and the full fast suite all pass; the fast suite's 21 failures are the same pre-existing, environment-only (castxml/toolchain-availability, ADR-verified-commit staleness) failures present on `origin/main` before this change.
- `python scripts/check_ai_readiness.py`: 15 errors, identical to the `origin/main` baseline (no new findings).
- Changelog fragment `changelog.d/20260814_183041_noreply_p0_4_analysis_assurance.md` amended with both fixes.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_017jAiyuYpVCdif4RqZpJ1mP)_

### DIFF
--- a/abicheck/analysis_assurance.py
+++ b/abicheck/analysis_assurance.py
@@ -56,15 +56,37 @@ Nothing here shells out, re-parses a binary, or re-runs an extractor.
   ``compare --depth source`` that never actually reached source evidence
   (both sides lacking a compile database, say) silently read
   ``requested_depth=None``/``depth_satisfied=None`` and could still report
-  ``status="complete"``. Fixed at the one front end that knows the user's
-  real, explicit ``--depth`` request: ``cli_compare_helpers.
-  _report_compare_result`` now copies its own Click-validated ``--depth``
+  ``status="complete"``. Fixed first at the one front end that knows the
+  user's real, explicit ``--depth`` request: ``cli_compare_helpers.
+  _report_compare_result`` copies its own Click-validated ``--depth``
   string onto ``DiffResult.requested_depth`` before recomputing this block
   (never an *inferred* depth from bare ``--sources``/``--build-info`` or
   ``.abicheck.yml``'s ``source.method`` -- only an explicit flag is this
-  comparison's stated request in the same on-the-record way). Any other
-  caller that never sets ``DiffResult.requested_depth`` keeps the prior,
-  unaffected behavior (``None``, depth-completeness untested).
+  comparison's stated request in the same on-the-record way). That fix was
+  CLI-only, though the underlying gap was not (P2 review,
+  ``discussion_r3787839902``): ``service.run_compare_request()`` -- the
+  shared, typed Python-API/MCP entry point ``service_compare_pipeline.
+  classify_compare_pair`` implements -- never gets this treatment, since
+  the CLI's own ``compare`` command resolves its evidence through
+  ``resolve_compare_request`` but classifies through a hand-rolled
+  ``compare_snapshots`` call of its own (its ``--depth`` flag never even
+  reaches ``CompareRequest.depth``; that field only carries a real value
+  for a *direct* typed-API caller). Fixed there too:
+  ``classify_compare_pair`` now applies the identical ``if request.depth
+  is not None`` stamp-and-recompute, so a direct
+  ``service.run_compare_request(CompareRequest(..., depth="headers"))``
+  call gets the same `requested_depth`/`depth_satisfied` receipt the CLI
+  already did, without disturbing a caller that never sets
+  ``CompareRequest.depth`` (unaffected, same as before). The two call
+  sites are not one shared helper because each recomputes from genuinely
+  different inputs the pipeline made available at a different point: the
+  CLI recomputes with a possibly out-of-band ``--old/new-build-info``/
+  ``--old/new-sources`` pack (`_resolve_side_pack`) that never touches
+  ``old``/``new`` directly, while ``classify_compare_pair`` recomputes from
+  each snapshot's own already-embedded ``build_source`` (mirroring
+  ``checker.compare()``'s own call) since a typed request's
+  ``InputSpec.sources``/``build_info`` are embedded before resolution ever
+  returns -- there is no separate out-of-band pack to fold in on that path.
   ``effective_depth`` is always computed here, independent of that field,
   from what each side's snapshot actually carries (mirrors
   ``cli_dump_helpers.evidence_depth_label``'s logic, reimplemented locally
--- a/abicheck/analysis_assurance.py
+++ b/abicheck/analysis_assurance.py
@@ -49,15 +49,28 @@ Nothing here shells out, re-parses a binary, or re-runs an extractor.
 - Requested-vs-effective *depth* reuses ``checker_types.EVIDENCE_DEPTH_VALUES``
   (``binary``/``headers``/``build``/``source``) rather than inventing a
   parallel vocabulary. ``requested_depth`` mirrors
-  ``DiffResult.requested_depth`` -- the G30 report-identity field nothing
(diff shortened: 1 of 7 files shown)


abicheck/analysis_assurance.py
napetrov, line 1: **Finding 1 fix (Propagate the requested depth into assurance)**

`compute_analysis_assurance()` reads `result.requested_depth` here to gate `depth_satisfied`, but nothing populated that field before this PR — `cli_compare_helpers._report_compare_result` never copied the CLI's resolved `--depth` onto `DiffResult.requested_depth`, so an explicit `compare --depth source` that never reached source evidence (e.g. both sides lack a compile database, leaving the effective depth at `headers`) silently read `requested_depth=None`/`depth_satisfied=None` and could still report `status="complete"`.

Fixed at the front end that knows the user's real, explicit request: `cli_compare_helpers._report_compare_result` now copies its own Click-validated `--depth` string onto `DiffResult.requested_depth` before recomputing this block (see `abicheck/cli_compare_helpers.py`) — only from an explicit flag, never an inferred depth from bare `--sources`/`--build-info` or `.abicheck.yml`'s `source.method`. Regression coverage: `tests/test_analysis_assurance_depth_and_graph_overlap.py::TestRequestedDepthPropagation` (unit-level and end-to-end CLI).

---


---

napetrov, line 1: **Finding 2 fix (Reject partially overlapping graph coverage)**

Round 8's `_graph_completeness()` fix only flagged a fully **disjoint** confirmed-family pair (`old_confirmed_families.isdisjoint(new_confirmed_families)`). Fresh evidence: the partially overlapping case — old confirms `{call_graph, type_graph}`, new confirms only `{call_graph}` — is not disjoint (they share `call_graph`), so `graph_completeness` stayed `"complete"` even though `buildsource/source_graph_findings.py` only trusts a family when **both** sides cover it, meaning `type_graph` was silently skipped for the comparison.

Fixed by changing the condition from "sets are disjoint" to "sets are not equal" (`old_confirmed_families != new_confirmed_families`) — any difference between the two sides' confirmed-family sets, whether total disjunction or partial overlap, now reads as asymmetric (`graph_completeness="unknown"`, folding into `status="partial"`), and the resulting note names which family/families are confirmed on only one side. Regression coverage: `tests/test_analysis_assurance_depth_and_graph_overlap.py::TestGraphCompletenessPartialFamilyOverlap`, plus companion tests confirming round 8's original disjoint case and the "identical confirmed sets" case are unaffected.

---


---

abicheck/cli_compare_helpers.py
```diff
@@ -1076,6 +1077,29 @@ def _report_compare_result(
...
+    # `analysis_assurance` below so the requested-vs-effective gate actually
+    # has something to check. Only ever set when the flag was explicitly
+    # given, mirroring the same "explicit override, never inferred"
+    # discipline `_resolve_compare_collect_mode` already applies to `depth`
+    # itself for collect-mode resolution -- an inferred depth (from
+    # `--sources`/`--build-info` alone, or `.abicheck.yml`'s `source.method`)
+    # is not this comparison's stated request in the same on-the-record way
+    # an explicit `--depth` flag is.
+    if depth is not None:
+        result.requested_depth = depth
```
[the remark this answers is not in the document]
napetrov: Fixed in `ac57d27c`.

`service_compare_pipeline.classify_compare_pair` — the shared classification half `service.run_compare_request()` (the typed Python API / MCP `abi_compare` entry point) actually runs — now stamps `DiffResult.requested_depth` from an explicit `CompareRequest.depth` (`if request.depth is not None`, mirroring the CLI's own guard) and recomputes `analysis_assurance` from each snapshot's own embedded `build_source`, since this path has no out-of-band `--old/new-build-info`/`--old/new-sources` pack directory to fold in the way the CLI does — it mirrors `checker.compare()`'s own recomputation instead of `cli_compare_helpers._resolve_side_pack`.

`cli_compare_helpers._report_compare_result`'s own stamp is deliberately kept, not removed: it still needs to recompute after resolving its own out-of-band pack, which `classify_compare_pair` has no visibility into — so the two are genuinely different inputs at different points in the pipeline, not a duplicate to collapse into one helper.

Added a direct typed-API-level regression test class, `TestRequestedDepthPropagationSharedPipeline` (`tests/test_analysis_assurance_depth_and_graph_overlap.py`), covering exactly the repro this finding describes:
- a direct `run_compare_request(CompareRequest(..., depth="headers"))` call where the depth is satisfied,
- the companion no-depth-set case (unaffected, same as before),
- and a direct `classify_compare_pair()` unit-level case where the requested depth is *not* satisfied, confirming `status != "complete"`.

Verification: `pytest tests/test_analysis_assurance*.py` (89 passed), `ruff check`/`ruff format --check`, `mypy abicheck/` (0 errors), the full fast suite (26912 passed, 21 pre-existing environment-only failures — castxml/toolchain-availability and ADR-verified-commit staleness, identical to the `origin/main` baseline), and `python scripts/check_ai_readiness.py` (15 errors, identical to the `origin/main` baseline — no new findings). Patch coverage on the touched lines in `analysis_assurance.py`/`service_compare_pipeline.py` is 95%/92% respectively via `pytest tests/test_analysis_assurance*.py --cov=abicheck.analysis_assurance --cov=abicheck.service_compare_pipeline`.

---

---

PR #770, opened 2026-08-15
refactor(cli)!: hard cleanup — remove superseded and duplicate flags (no deprecation window)

napetrov:
## Summary

Physically remove the CLI surface that #757 superseded, and fold the four duplicate spellings the cleanup list names, instead of keeping any of it as hidden-but-functional aliases. Pre-1.0, and the replacement spellings are days old — a deprecation window here only preserves the maintenance cost (alias-conflict resolvers, duplicate help text, two code paths) without protecting anyone who has had time to depend on the old names.

This is now the **whole** first-delete list, plus the four atomic replacements. Nothing is deferred.

Rebased onto `main` @ 5b29af4 (through #769/#774).

## Motivation

The CLI audit's conclusion was that the biggest source of perceived complexity is *two ways to say one thing*. #757 added the better spellings but kept the old ones alive, which made the surface larger, not smaller — `--help-all` hides them from readers, but the parser, the resolver, and the tests all still carry them.

A flag that no longer does anything is the worst case: `--verify-runtime` had already been hardened into a documented safety no-op, so a caller passing it got silence rather than an error.

## Removed outright, no aliases

- **`compare --verify-runtime`** — already a safety no-op that never executed anything and always reported `attempted=False`. Removed with `abicheck.runtime_probe`, the `consumer_runtime_load_failed` `ChangeKind` nothing could produce anymore (395 kinds now), and the `verify-runtime` inputs on `action.yml`, `actions/check-target`, and the `check-single`/`check-project` reusable workflows.
- **`--gcc-path` / `--gcc-prefix` / `--gcc-option`** on `compare`/`dump`/`scan` — `--compiler` / `--compiler-prefix` / `--compiler-option` are the only spelling now. This also deletes `_merge_compiler_aliases`, whose only correct answer for the repeatable option pair was to reject mixing the two spellings, since Click does not expose the relative argv order of two separately-collected tuples. `CompileContext`'s internal `gcc_*` field names are unchanged; the Action keeps its `gcc-path`/`gcc-prefix`/`gcc-options` input names and forwards them to `--compiler*`.
- **`--contract-evaluation`** on `compare` and `scan --against` — `--contract` already implied it, and gains a fourth value **`auto`** for the one thing the switch alone expressed: evaluate without stating a domain, letting D7's chain below `explicit_cli` decide. Without that value the `legacy_alias` and `project_config` tiers for `contract.mode` would have become unreachable from the CLI.
- **`compare --show-impact`** — `--report-mode impact` was already documented as its exact equivalent.
- **`aggregate --expect` / `--optional` / `--report-prefix`** — the expected-target set is declared only by `--manifest` or `--run-plan` now (or waived with `--discovered-only`).
- **The four per-category `--severity-<category>` flags** (compare/scan, and the Action's `severity-addition` input) — hidden CLI duplicates of `.abicheck.yml`'s own `severity:` block, which is now their one spelling. `--severity-preset` stays as the visible coarse per-run override. `test-action.yml`'s severity lane now names a config fixture instead of the flag.
- **`--strict-suppressions`, `--require-justification`, `--public-symbol`, `--public-symbols-list`, `--show-redundant`, `--collapse-versioned-symbols`** — same shape: each duplicated a `suppression:`/`scope:` config key (`suppression.strict`, `suppression.require_justification`, `scope.public_symbols`, `scope.show_redundant`, `scope.collapse_versioned_symbols`).
- **`dump --public-header` / `--public-header-dir`** — declaration provenance (ADR-015) now comes from `-H/--header` itself: a file entry tags that header public, a directory entry tags everything under it, via the `split_public_header_inputs` partition `compare` has always applied to its own `-H` list. `scan` keeps its own `--public-header-dir`.
- **`dump -p/--build-dir` and `--compile-db`, and `scan --compile-db`** — `--build-info` already takes exactly that operand (a build dir, a `compile_commands.json`, or a pack), so it is now the one flag for it on every command. With `-H` also given, the database it resolves to parameterizes the header parse the way `-p` did; `--compile-db-filter` still scopes it. One usage error goes away with the flag: a database with no headers is now an ordinary L3-only dump rather than "requires -H/--header".
- **`project validate-use-cases --against` / `--against-new`** — see the feature port below.

## Changed — the four atomic replacements

- **`--policy-file` folded into `--policy NAME|PATH`.** A built-in profile name selects that profile; anything else resolves as a policy document (a path, or a packaged built-in like `security`), exactly what `--policy-file` did. Two flags for one question, the second silently winning when both were given, is now one — and the "`--policy` is ignored when `--policy-file` is given" warning has nothing left to warn about. The Action's `policy-file` input is unchanged and still outranks `policy`.
- **`--secondary-format` + `--secondary-output` folded into `--write FORMAT=PATH`.** Half the pair was a usage error in either direction, so they were one option spelled as two; the two half-given checks are gone rather than unreachable.
- **`--old-ast-frontend`/`--new-ast-frontend` folded into a side-aware `--ast-frontend`** on `compare`: `--ast-frontend old=castxml --ast-frontend new=clang`, ADR-040 Lever 1's same `old=`/`new=` convention as `--header`/`--include`/`--version`. A bare value still applies to both; `dump`/`scan` keep the single-valued spelling.
- **`--include-dependencies` renamed `--include-system-declarations`.** It restores the declarations a system/toolchain header contributed to the header AST, which has nothing to do with the `DT_NEEDED` graph `--follow-deps` walks — the old name read as the latter. The snapshot field, the `service.run_dump` parameter, and the cache key keep their internal spelling.

## The one feature port: `compare --use-cases`

`project validate-use-cases --against/--against-new` was a manifest validator that had grown a second snapshot-diffing surface inside it. Deleting the flags alone would have removed use-case impact analysis, so the capability moved to where the comparison already happens:

`abicheck compare OLD NEW --use-cases MANIFEST` resolves each declared use case's entrypoints against **both** sides' source graphs and reports which of the comparison's own findings each use case reaches — as the report's `use_case_impact` block (report schema **2.39**) and a text/markdown section. Each side is explained against its own graph and the two are unioned per symbol, so a symbol added on the NEW side is not read as unattributed just because OLD's graph never had it.

Read-only by construction: an unattributed finding is an absence of proof, not proof the finding is harmless, so it never moves a verdict, a gate, or an exit code. `unattributed_changes` counts the remainder rather than leaving it implied. `project validate-use-cases` still checks a manifest's structure.

## Bug-fix test contract

One commit in this branch is a `fix:` — `e870b74`, three defects this refactor itself introduced mid-flight and then closed. Answering for that commit, plus the rows this diff's own touched areas trigger:

- Bug class: a flag removal that leaves a *neighbouring* declaration holding the removed one's trailing attribute, and a flag-value normalization applied to one command's local variable instead of to the shared resolver input every command rebuilds from. Both are "removed X, but Y still reads where X used to be", not three unrelated typos.
- Publicly observable failure: `abicheck scan --against BASE NEW --contract auto` exited 1 with an uncaught `ValueError` from `coerce_contract_mode`; every compare-mode run of the composite Action forwarded `--used-by false` (the orphaned `default: 'false'` from the deleted `verify-runtime` input landing on the preceding `used-by` input), failing unless a file literally named `false` existed; `scripts/benchmark_comparison.py` exited 64 on both case115 lanes instead of measuring.
- Regression test fails on base: yes — `tests/test_cli_compare_contract_evaluation.py::TestContractFlagResolvers` and the parametrized compare/scan `--contract auto` case fail on the pre-fix tree specifically on the `scan` side. The `compare` side passed, which is exactly why the local-variable normalization looked sufficient.
- Negative control: `--contract public|exports|all` must still reach the resolver as an explicit CLI value with `ParameterSource.COMMANDLINE` intact — pinned by `test_a_named_domain_is_left_alone`, so the fix cannot "work" by demoting every `--contract` value to `DEFAULT`.
- Public-surface test: driven through the real `compare`/`scan` CLIs with `CliRunner`, and through `action/run.sh` executed for real in `tests/test_action_run_sh_*` — not only through `resolve_contract_domain` in isolation.
- Axes covered: both commands that take `--contract` (`compare`, `scan --against`); both resolver entry shapes (explicit kwargs vs. `ctx.params` rebuild); the Action's YAML input surface and its `run.sh` argv assembly; and the `scripts/` consumer.
- General invariant: `--contract auto` never reaches `coerce_contract_mode` on any front end — it is normalized to "no domain stated" *and* demoted from `COMMANDLINE` to `DEFAULT` at one shared function, so recording it as an explicit CLI value cannot re-create the D7 precedence bug the value exists to avoid. Separately, every retired spelling is registered in `scripts/check_docs_contract.py`'s `_RETIRED_SURFACES`, so a stale mention in `docs/` fails a gate rather than surviving as prose.
- Real-dependency test: this diff touches the published report-schema boundary. `use_case_impact` (schema 2.39) is asserted through the real `compare --format json` chokepoint and validated against the published `compare_report.schema.json` by `tests/test_cli_compare_contract_evaluation.py`'s `jsonschema.validate` case, not against a hand-built dict.
- Malicious fixture + side-effect absence: this diff touches the Action/workflow trust boundary. `action/run.sh`'s argv assembly is *executed* (not text-asserted) by `tests/test_action_run_sh_artifact_set.py` and `tests/test_action_run_sh_helpers.py`, including the audit-only path that must omit `--policy`/`--suppress` entirely and the internal `--write` injection that must be skipped when `extra-args` already carries one.
- Verdict, gate and exit code checked independently: yes. The severity tests that moved to a project config assert all three separately in one run — `res.exit_code == 1`, `gate["blocking_categories"] == ["addition"]`, and `Verdict: COMPATIBLE` (`tests/test_cli_scan.py`) — so a compatible verdict with a non-zero gate exit cannot be conflated into a single assertion.
- False-positive removed / real break preserved: the widening overlay moved from `--public-symbol` to `scope.public_symbols`, and `tests/test_surface.py::TestWideningCLI` asserts **both** directions in one test — the same `InternalCache` layout change must stay *filtered* under bare `--scope-public-headers`, and must come *back* into the report once the config names it. A widening that suppressed nothing, or one that stopped filtering at all, fails one half or the other. `dump`'s provenance switch to `-H` is guarded the same way from the other side: `tests/test_provenance_integration.py` still requires every origin to stay `UNKNOWN` with no public-header set given, so deriving the sets from `-H` cannot silently classify everything public.
- Known unsupported cases: `--pack` gate-severity folding is still not extended to `scan` (pre-existing, unchanged here). The one retired pair deliberately *not* registered in the docs gate is `--public-header`/`--public-header-dir`, since `scan --public-header-dir` is still live and the gate matches plain substrings; its executable guard is `tests/test_cli_contract.py`'s per-command option-set snapshot instead.

## Not changed, deliberately

The typed Python API (`CompareRequest`/`ScanRequest`) still requires `contract_evaluation=True` alongside `contract_mode`, and `service._validate_contract_mode` still enforces it — both are documented public-API contracts, and the CLI resolver runs before either is constructed. `ScanRequest.compile_db` likewise stays as an API field.

The **second-PR** items from the cleanup list are still out of scope, as agreed: removing `--exit-code-scheme`, folding `--stat`/`--recommend` into `--report-mode`, `scan --artifact-set` → positional directory dispatch, and moving `--annotate*` to the Action layer.

## Structure

`reporter.py` was at the 2000-line hard cap, and the new `use_case_impact` block pushed it over; its contract/assurance report blocks moved to `reporter_contract_blocks.py`, the same split `reporter_markdown`/`report_summary` already are.

## Verification

- `ruff check`, `mypy abicheck/` (0 errors), `scripts/check_docs_contract.py` (0/0), and `scripts/check_ai_readiness.py` clean apart from the pre-existing ADR `**Verified:**` receipt errors, which fail identically on unmodified `main` in this environment because the clone is shallow and those commits are not reachable from `origin/main`.
- Every derived artifact regenerated: CLI reference, Action reference, detector spec, example case pages + `ground_truth.json` prose, agent-skill trees, skill-eval pack, published JSON schemas.
- Spot-checked the resulting surface: `--write`, `--policy NAME|PATH`, `--ast-frontend old=/new=`, `--include-system-declarations`, `--use-cases` present; `--severity-abi-breaking`, `--strict-suppressions`, `--public-symbol`, `--public-header`, `-p/--build-dir`, `--compile-db`, `--policy-file`, `--secondary-*`, `--old/new-ast-frontend`, `--against-new` all gone.

## Checklist

- [x] Changelog fragment added — `changelog.d/20260814_120000_claude_cli_hard_cleanup.md`
- [ ] `pre-commit run --all-files` passes locally — `verify.py`'s `fmt-check` step fails identically on unmodified `main` here (pinned-ruff mismatch), so it is not a signal from this branch

https://claude.ai/code/session_01GkoWr6uGQA7HBkfrGfDLbp


napetrov:
`unit-tests (windows-latest, 3.13)` is red here, but the failure reproduces on `main` and predates this branch — I'm not pushing a fix for it into this PR, but here's the diagnosis so it isn't lost.

**Same test, same assertion, same numbers on both:**

```
FAILED tests/test_action_compile_context_parity.py::TestCompileContextForwardingParity::
       test_gcc_options_quoted_value_stays_one_token - AssertionError: assert 3 == 2
```

- `main` @ a76d6dd — `cmd = ['--ast-frontend', 'clang', '--gcc-path', ..., '--gcc-prefix', ...]`
- this PR @ 6a91b57 — `cmd = ['--ast-frontend', 'clang', '--compiler', ..., '--compiler-prefix', ...]`

The only difference is this PR's `--gcc-path` → `--compiler` rename, which cannot change `cmd.count("--compiler-option")`. Both `add_flag_shlex_split` in `action/run.sh` and the test itself are byte-identical to `main` on this branch.

`main` has been red on this lane since 85c1bb1 (#757, which introduced the function); the last green push run is 0ca0c40.

**Cause.** `action/run.sh:82`:

```python
for tok in shlex.split(sys.argv[1], posix=(os.name != "nt")):
```

On Windows that resolves to `posix=False`, which produces exactly the three malformed tokens the function exists to prevent:

```
posix=True   -> ['-DMSG=hello world', '-DOK=1']          # 2 --compiler-option pairs
posix=False  -> ['-DMSG="hello', 'world"', '-DOK=1']      # 3 -- the observed failure
```

`os.name` answers "which OS is Python on", but the question is "which quoting convention did the user write this YAML input in" — and that is POSIX on every runner, because `run.sh` is a bash script executed by Git Bash on Windows. Splitting the *same* `gcc-options:` value differently per platform is itself the defect. Every other input in that script already goes through bash's own POSIX word-splitting, including this function's own `_PY_BIN`-unset fallback to `add_flag`.

So the fix looks like dropping the condition:

```diff
-for tok in shlex.split(sys.argv[1], posix=(os.name != "nt")):
+for tok in shlex.split(sys.argv[1], posix=True):
```

with the `import os` above it no longer needed. Worth a Windows-path case (`-IC:\sdk\include`) alongside the existing quoted-value one, since POSIX mode treats backslash as an escape — that is presumably what the `nt` branch was reaching for, and it deserves a real test rather than a platform switch.

Happy to push that here if you'd rather not carry a separate PR — say the word. Every other lane on this PR is green so far.

---

PR #773, opened 2026-08-15
docs: clarify Action include-dir docs for dependency headers (PVXS acceptance)

napetrov:
## Summary

Traces and closes out a real-world PVXS Action acceptance-run failure (`castxml`/`clang` aborting on `pvxs/version.h`) — clarifies the Action's include-dir docs and adds an end-to-end regression test.

## Motivation

A PVXS Action acceptance run reported all 4 self-compare jobs failing inside `abicheck` header parsing on `pvxs/version.h`, with only an opaque `castxml ... returned non-zero exit status 1` reaching the logs.

I reproduced this end-to-end against real EPICS Base 7.0 + PVXS 1.5.2 sources (built both from scratch) through the real `abicheck` CLI. Root cause: `pvxs/version.h` transitively `#include`s EPICS Base's `epicsVersion.h`/`epicsTime.h` — headers belonging to a *separate dependency* — and only PVXS's own `include/` was passed via `header`, with no `-I`/`include` for EPICS Base's include directory.

This is **not a header-parse defect** in the current codebase: both header AST backends (`dumper_castxml_probe.py` for castxml, `dumper_clang_errors.py` for clang) already raise an actionable `SnapshotError`/`HeaderToolchainError` naming the missing include and pointing at `--include-dir`/`-I`, rather than a bare subprocess exit-code message. Confirmed directly:

```
Error: Failed to load --binary .../libpvxs.so: ... clang failed to parse the header(s) (exit 1):
.../pvxs/include/pvxs/client.h:18:10: fatal error: 'epicsTime.h' file not found
Hint: the include 'epicsTime.h' was not found. Add its directory with --include-dir / -I, ...
```

Adding `-I` for EPICS Base's include directories made the self-compare succeed cleanly (`verdict: COMPATIBLE`, `exit_code: 0`, `NO_CHANGE`).

The gap was that nothing in the Action's own docs called out this "public header transitively includes a dependency's header" pattern, so an integrator following just `header`/`public-header-dir` has no signal that a separate `include`/`-I` is required for a dependency like EPICS Base.

## Changes

- `action.yml`: `include`/`old-include`/`new-include` input descriptions now explicitly document the "header includes a dependency's header" pattern, with the concrete PVXS/EPICS Base example that triggered this investigation.
- `docs/reference/github-action-inputs.md`: regenerated via `scripts/gen_action_reference.py` to stay in sync.
- `tests/test_dependency_header_include_diagnostic.py`: new end-to-end regression test (mirrors `test_pvxs_regression.py`'s real-compile pattern) that builds a small library whose public header transitively includes a sibling dependency's header, and checks both halves: (1) without the dependency's `-I`, the CLI surfaces the actionable hint, not an opaque error; (2) with it, the same library parses cleanly and a self-compare reports no ABI change.
- Changelog fragment (`changelog.d/`) documenting the finding, in the style of the existing `pvxs_version_scan` fragments from prior PVXS validation passes.

## Checklist

- [x] Changelog fragment added — docs-only change (no `abicheck/**/*.py` touched), fragment added anyway for traceability
- [x] Docs updated (`action.yml` + regenerated `docs/reference/github-action-inputs.md`)
- [x] `ruff check`/`ruff format --check`/`mypy` pass on the new test file
- [x] `test_action_reference.py` and `check_ai_readiness.py` pass with no new errors

---

---
_Generated by [Claude Code](https://claude.ai/code/session_0187Neicki3kfs5FtW4xe2fX)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved guidance when dependency header include paths are missing.
  * Comparisons can complete successfully after supplying required dependency include directories.

* **Documentation**
  * Clarified that shared include paths support dependency-header roots and may be derived automatically.
  * Documented that side-specific include paths replace, rather than extend, the shared include list.

* **Tests**
  * Added regression coverage for missing dependency includes and successful no-change self-comparisons.


napetrov:
Fixed the real failure from this run: `integration-tests (macos-latest)` failed because my new regression test (`test_missing_dependency_include_dir_gives_actionable_hint`) assumed a hard CLI failure on every platform. On Mach-O, `dump`'s header-based scoping is best-effort — it degrades to a `UserWarning` + export-table fallback (exit 0) instead of a hard error, which is existing, documented behavior (see AGENTS.md's Mach-O scoping note), not a defect. Pushed 2b1b94d, which checks for the actionable hint wherever it actually surfaces (CLI output on hard failure, or the emitted warning on the soft fallback) instead of assuming one outcome.

The other failure, `unit-tests (windows-latest, 3.13, false)` — `test_action_compile_context_parity.py::TestCompileContextForwardingParity::test_gcc_options_quoted_value_stays_one_token` (`assert 3 == 2` on `--compiler-option` count) — is **not caused by this PR**. It's in an unrelated subsystem (gcc-options shell-quoting forwarding in `run.sh`) that this PR never touches, and it's already failing identically on `main` at HEAD (run [#6996](https://github.com/abicheck/abicheck/actions/runs/31856306714), commit `d7c90b83`, same job, same assertion, same command snippet). Leaving it alone as out of scope for this PR.

---

napetrov:
Re the "keep the clang-only regression out of the castxml-gated marker" comment: confirmed the underlying observation directly — without castxml on PATH, both this new test *and* `test_pvxs_regression.py` report `Required tools not found: castxml` and skip, because `tests/conftest.py`'s `_integration_skip_reason()` gates on `castxml`+`gcc`+`g++` unconditionally for any `integration`-marked test on Linux, regardless of which backend an individual test actually needs. So the "runs on clang-only hosts" claim is genuinely inaccurate right now — but this is a pre-existing, repo-wide gate `test_pvxs_regression.py`'s identical docstring claim already relied on before this PR, not something introduced here.

Didn't apply the literal suggestion (dropping the `integration` marker) though: `tests/CLAUDE.md` explicitly says *"Don't change the marker scheme — CI gates depend on it"*, and the marker — not this test's own g++/clang `skipif` guards — is what keeps a real-compile test out of the "fast, ~45s" default lane (same file: *"Mark tests that shell out (gcc, castxml, etc.) with the matching marker so default runs stay fast"*). Removing it would make this test start executing real compiles by default on any dev machine that happens to have g++ installed, which is a bigger behavior change than this PR should be making. Pushed `1f07534` instead, correcting the docstring to state the real, current constraint accurately rather than the false "runs without castxml" claim.

---

napetrov:
Addressed two more findings in `bb2fc47`, both verified against code before writing the fix:

- **Codex** ("restrict replacement semantics to compare mode"): confirmed — `dump` forwards `include`/`new-include` as two separate `-I` flags (`run.sh:665-666`), and `scan` concatenates shared + side-specific paths (`cli_scan.py:1605-1609`, `run.sh:1046-1048`); only `compare`'s `_resolve_per_side_options()` actually overrides. Scoped the replace-vs-append claim per mode. Also caught and fixed a related error in `old-include`'s own description from the prior commit: it's not compare-only — `scan --against` uses it for the baseline side too (`run.sh:1044-1047`), additively there, unlike compare's override.
- **CodeRabbit** (document the Mach-O warn+fallback path): confirmed and already independently known — this PR's own regression test had to account for exactly this split (`service_header_scoped.py`'s `_try_header_scoped_dump`, which covers PE too, not just Mach-O — checked its docstring). Documented both the ELF hard-fail and the PE/Mach-O soft-fallback outcomes.

---

PR #776, opened 2026-08-15
fix: close two evidence-crediting holes in the gates from #769

napetrov:
## Summary

Two review findings that arrived on #769 after it was cut, and merged before they could land there. Both are the same shape: a gate crediting evidence that cannot exist.

## Motivation

**A test-shaped basename outside a `tests` tree counted as a regression test.** `testpaths = ["tests"]` and the canonical command is `pytest tests/`, so `foo/bar_test.py` is never collected by any lane — crediting it let a `fix:` PR satisfy the structural half of the bug-fix contract with a file nothing runs. The single real file the fallback matched makes the point better than a hypothetical: `agent-evals/tasks/*/hidden_tests/test_*.py`, an eval fixture the suite deliberately does not execute.

**An unresolvable base was a warning, and the run still exited 0.** With `--diff-scoped` on a deletion-only change, the removed function is absent from `touched`, so a survivor inside the function the branch gutted goes unreported — green for a measurement that never happened, which is exactly the shape the mutation gate exists to remove.

## Changes

- `is_test_path()` — dropped the outside-`tests/` basename fallback. Every directory named `tests` in this repository is executed by some lane (the root suite, the clang plugin's, the layout tool's), so the directory rule needs no allowlist. One test reads `testpaths` out of `pyproject.toml` rather than restating the configuration claim the rule rests on.
- `check_mutation_score.py` — an unattributable removal now fails the run **when it could hide something**: the removals are in a module that carries a surviving mutant. It still only warns when they are not, because a removal in a module with no survivor has nothing to hide, and failing there would block branches on an absence of evidence that cannot matter.
- Keyed on *pure* deletion hunks (`pure_deletion_paths()`), not removals in general: a modification removes and adds in the same hunk and its new side is already attributed, so keying on removals alone failed closed on ordinary edits.

## Verification

Each fix falsified by reverting it:

| Fix | Reverted to | Result |
|---|---|---|
| basename fallback removed | previous fallback | 3 path tests fail |
| fail-closed on at-risk removals | warn-and-exit-0 | survivor-at-risk test fails |
| pure-deletion keying | all removals | ordinary-modification control fails |

Two negative controls each: a module with no survivor (warns, exits 0) and an ordinary modification (no failure). 260 tests pass across the two gate suites; `ruff check`, `ruff format --check` and `check_ai_readiness.py` clean.

## Bug-fix test contract

- Bug class: a gate accepting evidence that cannot exist — an uncollected test file, and an unmeasurable diff reported as measured
- Publicly observable failure: `check_bugfix_test_contract.py` exits 0 for a `fix:` that changed only `foo/bar_test.py`; `check_mutation_score.py --diff-scoped` exits 0 on a deletion-only diff whose base is unreadable while a survivor sits in the deleted function
- Regression test fails on base: yes — `test_a_test_basename_outside_a_test_tree_is_not_collected` and `test_an_unreadable_base_fails_closed_when_a_survivor_could_be_hidden` both fail against the previous predicates
- Negative control: a module with no surviving mutant still only warns, and an ordinary modification is not treated as an unattributable removal — so neither fix can pass by failing on everything
- Public-surface test: yes — both go through `gate.main()` with real argv, not the internal predicates alone
- Axes covered: pure deletion, modification, pure addition; survivor in the same module vs. a different one; test basename inside a `tests` tree, outside one, and in the agent-eval fixture tree
- General invariant: a path is regression evidence only if some lane collects it, and a diff-scoped run exits 0 only when every removal that could carry a survivor was attributed to a function
- Known unsupported cases: a removal spanning several functions inside one mixed hunk is attributed through its new side only; the base-side half is consulted only for hunks with no new side at all

## Checklist

- [x] Changelog fragment added (`scriv create`) if this touches `abicheck/**/*.py` — n/a, no `abicheck/` source changed
- [x] Docs updated if user-visible behaviour changed — the rules' own docstrings carry the reasoning

https://claude.ai/code/session_01Rhf5cz7rTCDBtyFrT6MxG4

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rhf5cz7rTCDBtyFrT6MxG4)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test detection to recognize only files collected from valid test directories.
  * Strengthened mutation checks to fail when pure deletions could hide surviving issues and the comparison baseline is unavailable.
  * Preserved non-fatal warnings when no relevant hidden issue is found.

* **Tests**
  * Expanded coverage for test-path validation, deletion detection, and mutation-score enforcement.
  * Added checks for repository test collection settings and common edge cases.

### DIFF
--- a/scripts/check_bugfix_test_contract.py
+++ b/scripts/check_bugfix_test_contract.py
@@ -152,10 +152,7 @@ _TRUST_BOUNDARY_PREFIXES = (".github/workflows/", ".github/actions/")
 
 #: Directory component that marks a test tree.
 _TEST_DIR = "tests"
-#: Basename shapes that mark a test module.
-_TEST_BASENAME_PREFIX = "test_"
-_TEST_BASENAME_SUFFIX = "_test.py"
-_TEST_BASENAMES = frozenset({"conftest.py"})
+
 #: Prose suffixes. A file under `tests/` with one of these is documentation,
 #: not an executable test or a fixture — editing `tests/CLAUDE.md` alongside a
 #: shipped-code fix must not satisfy "you changed a test" (Codex review).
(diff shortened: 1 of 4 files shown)


napetrov:
CI red on `unit-tests (windows-latest, 3.13)`, failing on the base branch too — not from this PR.

The single failure is `tests/test_action_compile_context_parity.py::TestCompileContextForwardingParity::test_gcc_options_quoted_value_stays_one_token`, asserting `assert 3 == 2` on `cmd.count("--compiler-option")`. The same test fails identically on `main` at `150a710cb` ([run 31855022527](https://github.com/abicheck/abicheck/actions/runs/31855022527), Windows lane, `1 failed, 26552 passed`), so it predates this branch. This PR touches only `scripts/check_bugfix_test_contract.py`, `scripts/check_mutation_score.py` and their tests — nothing near `action/run.sh` or compile-context forwarding.

It looks like Git Bash on Windows evaluating the gcc-options/compiler-option forwarding differently from Linux (the extra token is `_FULL_ENV`'s own `INPUT_COMPILER_OPTION`, which the Linux run does not accumulate alongside `INPUT_GCC_OPTIONS`), but diagnosing that belongs with the Action's shell layer rather than here. Leaving it for a dedicated fix rather than folding an unrelated `action/run.sh` change into this one; every other lane on this PR is green.

---

PR #778, opened 2026-08-15
docs: fix contradicting and overclaimed technical statements in the ABI/API track

napetrov:
## Summary

Fixes the P0 truthfulness items from the ABI/API documentation review: claims in the learning track that contradicted either a sibling canonical page or the implementation. Review rounds then surfaced one place where the *code* made the same kind of claim, so this also corrects a user-visible finding message.

## Motivation

The review's finding was that the docs' main problem has shifted from structural gaps to **self-contradiction**. Two pages disagreed about where triviality evidence comes from; a tutorial page asserted a stronger unwinding outcome than the canonical deep dive that owns the topic; L0 detection was described as proving more than it can. Each item was verified against the code or the specification rather than taken at face value — and one of the review's own claims did not survive that check (see "Not addressed").

## Changes

- **`noexcept` (Part 4 §5)** asserted one unconditional outcome — landing pads omitted, `std::terminate()` fires, every destructor skipped. That contradicts `exception-unwinding-abi.md`, which correctly states the outcome depends on the compiled caller, toolchain, and control flow. Part 4 now separates the four independent facts (linkage / type system / behavioral contract / scanner behavior). The C++17 note also pointed at the wrong mangling component: `noexcept` encodes as `Do`, not the `E` that terminates every function type (verified: `_Z4takePDoFvvE` vs `_Z10take_plainPFvvE`).
- **Triviality evidence** was stated two incompatible ways. Part 4 claimed `value_abi_trait_changed` reads "the DWARF trivially-copyable flag"; `class-layout-abi.md` claimed the traits come only from castxml. Both wrong — DWARF has no such attribute (`dwarf_advanced.py` *infers* non-triviality-for-calls from DIE structure), and `dumper_castxml.py` deliberately leaves both traits `None`; only the direct-clang AST path supplies them. Also corrected: `data_size_bits` needs the optional layout-tool pass, `base_offsets` is not populated by plain direct-clang, and `vptr_offset_bits` *is* measured from DWARF rather than derived as `0` everywhere.
- **Binary-only (L0) `_ZTV`/`_ZTI` detection** was described as observing a virtual-method or base-class change from `.dynsym` alone, reorders included. A pure reorder preserves the size; so does a base swap keeping the same `type_info` runtime class and base count (which constrains `_ZTI` only, not `_ZTV`). A vtable entry is also not a source-level virtual — a virtual destructor takes two (verified against GCC: `_ZTV1A` 24 B → 40 B). Coverage is per retained symbol, not per polymorphic class.
- **MSVC/PE**: dropped the claim that multiple vptrs are absent from Itanium — it defines secondary virtual tables and a vptr per polymorphic base (verified against the spec); the real divergences are vbtables and the RTTI access path. Calling conventions scoped to Windows x86 (`__vectorcall` remains distinct on x64). CRT rule now separates heap identity from CRT compatibility, and Part 5's owning table was corrected to match.
- **`NO_CHANGE` → patch release** removed. A verdict constrains a version bump but cannot select one.
- **Part 8's accuracy ladder** (~1/3, 75%, 90%, 95%) disagreed with the one measured table. It now links to that table, and describes the climb as non-decreasing rather than rising at every rung (`headers` and `build` measure identically).
- **Part 1** equated "defined" with "exported", described stored symbol names as demangled, presented the name as universally the only lookup key, and said static linking ends the compatibility question — each corrected.
- **A sweep for twins** of the above, since five review rounds each found a corrected claim still standing elsewhere: `examples/case15_noexcept_change/README.md` carried the deleted `std::terminate` sentence verbatim, and `sycl-test-abi-coverage.md` carried the reorder overclaim twice.
- **Code fix:** `vtable_slot_count_changed` reported the `_ZTV` group size as a virtual-slot count — see the contract below.

### Not addressed (and why)

The review's P0.1 said CI is red because of CRLF handling in `test_readme_exit_code_table_is_in_sync` / `test_cli_doc_exit_code_table_is_in_sync`. **CI is genuinely red on `main`, but not for that reason** — neither test exists in this repo, and no docs-sync gate is failing. The real Windows failures are 6 tests in `test_action_run_sh_py_safe_path.py` and `test_action_run_sh_severity_summary.py` (`WindowsPath('/tmp/...')`, `/c/Users/...` vs `C:\Users\...`, a `mktemp` failure path exiting 0), identical on this branch and on `main`. Windows path/bash-shim portability in the Action's shell layer, unrelated to these changes and not verifiable from Linux.

The review's P1–P5 items (navigation restructure, hub trimming, splitting Learn pages from scanner-model reference, growing Verification & Assurance, the platform-model layer) are structural rather than truthfulness work and are deliberately not in this PR.

## Bug-fix test contract

- Bug class: a user-visible diagnostic asserting more than its evidence supports — `vtable_slot_count_changed` converted the whole `_ZTV` symbol size into a virtual-slot count, but that symbol spans the primary table plus any vcall/vbase offsets and secondary tables.
- Publicly observable failure: `case174_secondary_vtable_group_changed` reported `Vtable for 'Derived' changed size: 40 -> 80 bytes (~3 -> ~8 virtual slots)` for a `Derived` that declares no virtuals of its own — its 80 bytes are the primary table plus a secondary table for a base that became polymorphic.
- Regression test fails on base: yes. `test_description_does_not_claim_a_virtual_slot_count` asserts `"virtual slots" not in change.description`; on the base revision the detail string is exactly `~3 → ~8 virtual slots`, so the assertion fails there and passes only with the fix.
- Negative control: single-inheritance cases must keep reporting the same magnitudes — `test_virtual_added_grows_vtable` and `test_virtual_removed_shrinks_vtable` still assert the unchanged byte values, and `case142`'s `~4 -> ~5` is unchanged in magnitude (only relabelled). The fix must not suppress or alter any finding, and `test_stable_vtable_is_silent` / `test_added_only_vtable_not_reported` pin that the emitted set is unchanged.
- Public-surface test: the assertion is on `Change.description` returned from `_diff_elf_layout()`, the detector entry point whose output reaches the report and CLI — not on the `_vtable_group_entries()` helper's arithmetic, which is unchanged and was never the bug.
- Axes covered: single inheritance (case142/case09 shape) and the secondary-vtable-group shape (case174's real 40→80 byte sizes). Golden tests were run explicitly since they sit outside the fast lane and could have pinned the old string; `agent-evals/skills/skill-eval-pack.json` embeds it too and is regenerated.
- General invariant: the finding's description never claims a count of virtual slots. What L0 establishes is a change in the emitted vtable-group size; the byte sizes it reports were already exact and remain so, and the `ChangeKind`, verdict, and exit code are untouched.

## Checklist

- [x] `mkdocs build --strict` clean (no dangling links/anchors), `check_docs_contract.py` 0 errors / 0 warnings, `check_ai_readiness.py` 0 errors
- [x] No new `mypy` or `ruff` errors introduced; 28042 unit tests pass, plus the golden lane

https://claude.ai/code/session_019DAP1jt2uDPo6KvLDznTji


napetrov:
Addressed the review findings in `45fabc0` and `327ceb0`. Ten were valid and are fixed — most usefully the `vptr_offset_bits` provenance (DWARF genuinely reads a measured `DW_AT_data_member_location`, so calling it derived-as-`0` for all three producers was wrong), and the internal contradiction I had left behind where the "intrinsically hard" bullet still told castxml-only readers the opposite of the new producer matrix.

**One finding I'm declining, with reasoning** — the proposal to make `_vtable_slots()` parse only the primary vtable component, or rename `vtable_slot_count_changed`:

The underlying observation is correct and is exactly why this PR touched the file: `_ZTV` is a whole vtable group, so `size/ptr - 2` over-reports for multiple or virtual inheritance. But neither remedy works here.

- **Parsing the primary component is not possible at L0.** Decomposing a group into primary + secondary tables requires knowing the inheritance shape — which is precisely the DWARF/header evidence this detector exists to operate without. From a symbol size alone there is no boundary to find. `_ZTI`'s runtime class gives a coarse hint, but it distinguishes "one non-virtual base" from "multiple/virtual", not the per-component word counts.
- **Renaming the `ChangeKind` is a public-surface break.** `ChangeKind` values appear in report JSON/SARIF, policy files, and suppression rules, so a rename is a breaking API change requiring an ADR (per `AGENTS.md`'s "Adding a new ChangeKind" and the public-API note in `abicheck/CLAUDE.md`) — not something to fold into a documentation PR.

So the correct response for this PR is the one it already takes: state plainly, in both the page and the module docstring, that the kind is named for its commonest cause and that the *detected fact* is "the emitted group changed size" — with the `struct D : A` → `struct D : virtual A` case called out by name. If the numeric slot delta in the finding's message should also be softened, or an ADR opened for a rename, that's worth doing as its own change against `diff_elf_layout.py` with a multiple-inheritance regression test; I'd rather not reshape a detector's output inside a docs-truthfulness PR.

The RUF002 Unicode-minus flags are fixed too (they weren't failing this repo's pinned ruff config, but the ASCII spelling is free).

---

napetrov:
**On the vtable slot-count finding — I was wrong to scope this out earlier, and it's now fixed in `f785b26`.**

I declined the first version of this because the two remedies offered (parse the primary component; rename the `ChangeKind`) are respectively impossible at L0 and a public-surface break. Both of those still hold. But I treated that as disposing of the whole finding, when the actual defect sits in a third place I named as a follow-up and should simply have done: the **finding's message text**.

The catalog already contained the proof. `case174_secondary_vtable_group_changed` printed:

```
vtable_slot_count_changed: Vtable for 'Derived' changed size: 40 -> 80 bytes (~3 -> ~8 virtual slots)
```

`Derived` declares no virtuals of its own — the 80 bytes are the primary table plus a secondary table for a base that became polymorphic. So this PR had reached a state where the docstring said "the detected fact is *the emitted group changed size*, not *the slot count changed*" while the tool three functions below printed a slot count. That is the same defect the PR exists to remove, just with the two halves in code rather than in two docs.

Now reports `~N -> ~M vtable-group entries`. Byte sizes were already exact and are unchanged, as are the `ChangeKind`, verdict, and exit code — message text only. `_vtable_slots` → `_vtable_group_entries`, with case174 named in its docstring; regression test asserts the description never claims virtual slots, using case174's real sizes; six example READMEs and the regenerated pages updated; `skill-eval-pack.json` regenerated (its vtable-change scenario embeds the string).

---

**On CI:** the `unit-tests (windows-latest, 3.13)` failure on this branch is **pre-existing on `main`** — the same 6 tests, same assertions, in `test_action_run_sh_py_safe_path.py` and `test_action_run_sh_severity_summary.py` (`WindowsPath('/tmp/...')`, `/c/Users/...` vs `C:\Users\...`, and a `mktemp` failure path exiting 0). Compare run [31881108376](https://github.com/abicheck/abicheck/actions/runs/31881108376) on `5b52989` against [31885124480](https://github.com/abicheck/abicheck/actions/runs/31885124480) here. Windows path/bash-shim portability in the Action's shell layer, unrelated to these changes, and not something I can verify a fix for from Linux. Worth its own issue.

Note also that an earlier review comment described that failure as CRLF handling in `test_readme_exit_code_table_is_in_sync` / `test_cli_doc_exit_code_table_is_in_sync` — neither test exists in this repo, and no docs-sync gate is failing.

---

PR #780, opened 2026-08-15
fix: P0.4 analysis-assurance parity for scan, and Action gate for both

napetrov:
## Summary

`scan --against` now supports `--require-complete-analysis` (P0.4), matching `compare`, and the composite GitHub Action now maps an assurance-gated exit 1 to a dedicated `ANALYSIS_INCOMPLETE` verdict on both commands instead of mislabeling it.

## Motivation

A fresh end-to-end audit of `abicheck/abicheck@5b52989` from `napetrov/abicheck-bazel-lab` found that P0.4's analysis-assurance CLI contribution existed on `compare` but the composite Action never gave a caller a correctly-labeled way to see it fail: `scan --against` didn't accept `--require-complete-analysis` at all (a hard usage error), and `action/run.sh`'s exit-1 disambiguation had no notion of this axis, so an assurance-gated run read as `SEVERITY_ERROR` (a severity-policy failure it is not) or the catch-all `ERROR` (an operational failure it is not) — on either command.

## Changes

- `scan --against --require-complete-analysis`: floors the exit code to 1 on incomplete `analysis_assurance.status`, folded with the same `max` discipline `compare`'s flag and the contract-coverage axis already use (never lowers a 2/4/5/6). `scan --format json`'s summary now always carries `analysis_assurance` (nested under `diff`), regardless of the flag — mirroring `compare`'s report shape at the root. `scan` also now echoes `assurance_floor_diagnostic` to stderr, exactly like `compare`'s `_exit_with_severity_or_verdict` already does — the composite Action's detection reads only stderr for this axis (see below), and this was missing from the first push (Codex review). Rejected as a usage error without `--against`, alongside the rest of the baseline-only flags (`_COMPARISON_ONLY_FIELD_PREDICATES`/`_COMPARISON_ONLY_FLAGS`).
- `ScanRequest` deliberately does **not** gain a typed-API field for this — `CompareRequest` doesn't have one for `compare`'s own identically-scoped flag either, so this keeps the two commands' CLI-only treatment consistent rather than introducing a one-sided asymmetry.
- `action/run.sh`: new `_assurance_gated()` helper (stderr-diagnostic-based, since — unlike the coverage ledger — the JSON `analysis_assurance` block is always present regardless of whether the flag was passed, so it isn't self-describing about *this run's* gating). Both `compare` and `scan` map an assurance-gated exit 1 to a new `ANALYSIS_INCOMPLETE` verdict, disambiguated from a coincident severity or contract-coverage gate on the same exit 1, with an unconditional (`fail-on-*`-independent) `FINAL_EXIT=1` and a dedicated job-summary block.
- `docs/reference/exit-codes.md` / `docs/reference/cli-reference.md` (regenerated) updated to describe the new `scan` flag and the Action's mapping.

## Bug-fix test contract

- Bug class: missing CLI surface (`scan --against` had no `--require-complete-analysis` at all) + Action exit-code mislabeling (an axis the Action had no notion of fell through to an unrelated verdict) + a missing stderr diagnostic on the `scan` path that the Action's own detection depends on.
- Publicly observable failure: `abicheck scan lib.so --against baseline.json --require-complete-analysis` was a hard usage error regardless of how incomplete the evidence was; and, once wired, an assurance-gated Action step reported `SEVERITY_ERROR`/`ERROR` instead of the correct `ANALYSIS_INCOMPLETE`, which a caller's own tooling could misinterpret as an unrelated severity-policy or operational failure.
- Regression test fails on base: yes — `tests/test_scan_analysis_assurance.py`'s tests fail on base with a Click usage error (no such option), and `test_the_floor_diagnostic_is_echoed_to_stderr` fails on the pre-fix revision (no diagnostic emitted, the gap Codex caught); `tests/test_action_analysis_assurance_verdict.py`'s tests fail on base asserting `outputs["verdict"] == "ANALYSIS_INCOMPLETE"` (base produces `SEVERITY_ERROR`/`ERROR`).
- Negative control: `test_flag_omitted_never_changes_the_exit_code`/`test_complete_run_is_unaffected_by_the_flag` (scan CLI) and `test_a_run_without_the_flag_is_unaffected` (Action, both modes) assert the pre-existing behavior is unchanged when the flag/diagnostic is absent.
- Public-surface test: yes — all new tests drive the real `scan` Click command and the real `action/run.sh` script (via a stub `abicheck` on `PATH`), not internal helpers directly.
- Axes covered: flag-alone, flag-coincident-with-a-severity-gate, flag-coincident-with-a-contract-coverage-gate, and the pre-existing no-flag baseline — on both `compare` and `scan`.
- General invariant: the analysis-assurance axis is `max`-folded and orthogonal, mirrored end-to-end across `compare`/`scan` CLI, the composite Action, and (unconditional gating, never disabled by `fail-on-*`) — verified with real ELF-export-backed breaking/compatible/incomplete-evidence snapshot pairs, not toy fixtures.
- Verdict, gate and exit code checked independently: yes — CLI tests assert `res.exit_code`; Action tests independently assert `outputs["verdict"]`, `outputs["exit-code"]`, `outputs["_summary"]`, and `outputs["_exit"]` (the actual process exit code) separately.
- Known unsupported cases: the directory/package (release) `compare` fan-out still rejects `--require-complete-analysis` (P0.6, unchanged, already documented); a gate pack (`--pack`) does not fold this axis for `scan` any more than it already didn't for `compare`.
- Malicious fixture + side-effect absence: `tests/test_action_analysis_assurance_verdict.py::TestHostileReportContentCannotExecute` drives the real `action/run.sh` against a report/stderr crafted with shell command-substitution payloads (`$(touch marker)`) embedded in `analysis_assurance.notes` and in the stderr diagnostic text itself, and asserts the marker file never materializes (no code execution) while the verdict computation still reaches the correct answer; a separate test proves a forged stderr line that merely *mentions* the axis without the real diagnostic's exact shape is not mistaken for it (no spoofed gate).

## Checklist

- [x] Changelog fragment added (`changelog.d/20260815_130000_claude_scan_assurance_gate.md`)
- [x] Docs updated (`docs/reference/exit-codes.md`, regenerated `docs/reference/cli-reference.md`)
- [x] Full fast suite (28k tests via `pytest -n 4`), `mypy`, `ruff check`, `check_ai_readiness.py`, `check_docs_contract.py` all clean locally

## Known gap left for follow-up

Not attempted in this PR (see `abicheck/AGENTS.md`'s own "Known gaps" convention and `napetrov/abicheck-bazel-lab`'s `UPSTREAM_TO_ABICHECK.md` P0.3): the deeper dump-vs-scan extraction-contract divergence that currently makes a fresh `scan --against` a fresh `dump` baseline return `NOT_COMPARABLE` for the same project, because `dump`'s primary snapshot doesn't apply the L3 build context it embeds to its own L2 header parse the way `scan`'s live-binary path does. That is a materially larger, cross-cutting architecture change and is being investigated/tracked separately.

---

## Summary by CodeRabbit

* **New Features**
  * Added `--require-complete-analysis` for baseline comparisons with `scan --against`.
  * Added the equivalent Composite Action input and `ANALYSIS_INCOMPLETE` verdict.
  * Reports now include analysis-assurance status, incomplete targets, and exit contributions.

* **Bug Fixes**
  * Incomplete analysis now affects status when explicitly required.
  * Added validation for unsupported audit, release, directory, and package comparisons.

* **Documentation**
  * Updated CLI, exit-code, Action input, report schema, and changelog documentation.
  * Documented a known limitation affecting certain native ELF dump comparisons.


napetrov:
Fixed in 2015ec9: `_run_baseline_compare` now echoes `assurance_floor_diagnostic` to stderr on the `scan` path too (it previously only folded the exit code), matching `compare`'s own `_exit_with_severity_or_verdict`. Added `test_the_floor_diagnostic_is_echoed_to_stderr` in `tests/test_scan_analysis_assurance.py`, which drives the real `scan` command end-to-end and fails without this fix — the gap wasn't caught the first time because the Action-level tests used a hand-crafted stderr in the stub rather than the real CLI's own output. Also added `TestHostileReportContentCannotExecute` to `tests/test_action_analysis_assurance_verdict.py` for the bugfix-test-contract's malicious-fixture requirement this action/-touching diff triggers.

---
_Generated by [Claude Code](https://claude.ai/code/session_0148YM1JdtrGzWfUxXCBekzb)_

---

napetrov:
The Windows unit-tests failure (`unit-tests (windows-latest, 3.13, false)`) is pre-existing, not caused by this PR — none of the 6 failing tests are in files this diff touches (`test_action_run_sh_py_safe_path.py`, `test_action_run_sh_severity_summary.py`; zero diff against either), and they're MSYS path-translation/tmp-cleanup-timing issues (`/c/Users/...` vs `C:\Users\...`, a leaked tmp dir). The base branch's own most recent CI run at this PR's exact base SHA (`13be967`) shows the identical "Run tests (Windows — parallel, no coverage)" step already failing. Will re-check when the base branch recovers, per the "base branch recovered" re-run guidance rather than trying to fix unrelated Windows flakiness here.

Also pushed 3 more fixes addressing review feedback:
- Bumped `SCAN_SCHEMA_VERSION` 1.15 → 1.16 with a history entry for the new `diff.analysis_assurance` key (Codex P1).
- Rewrote `_assurance_gated()` to check `$CMD` (the real invocation argv, not attacker-influenceable) for the literal `--require-complete-analysis` token before consulting the report/stderr — the JSON `analysis_assurance.status` is now the authoritative signal once that's confirmed, with the stderr diagnostic only as a last-resort fallback when there's no readable JSON report (Codex P1, real spoofing risk: an unanchored stderr grep alone could be forged by any attacker-influenced diagnostic text to fail an otherwise-clean, flag-less run). Added `test_the_real_diagnostic_text_cannot_spoof_the_gate_without_the_flag_in_cmd` proving the exact scenario Codex described no longer spoofs.

Deliberately not doing (Codex P2, `ScanRequest` typed-API parity): `service_scan.py` is already at the 2000-line AGENTS.md hard cap, so adding the field needs either a real split of that file or trimming elsewhere first — out of scope for a same-PR reactive patch. Filed as a follow-up rather than force-fit here; `compare`'s own `CompareRequest` has the identical CLI-only asymmetry today, so this isn't a new class of gap.

---
_Generated by [Claude Code](https://claude.ai/code/session_0148YM1JdtrGzWfUxXCBekzb)_

---

napetrov:
Fixed in 1b05a8f:

- **P1, real bug**: `scan --against --depth source --sources <tree-with-no-compile-db>` silently reported `analysis_assurance.status="complete"` under `--require-complete-analysis`, because `_run_baseline_compare` never stamped `DiffResult.requested_depth` before `analysis_assurance` was computed — `compare`'s own CLI path already had this exact fix (`cli_compare_helpers._report_compare_result`), `scan`'s didn't. Now mirrors it: stamps `requested_depth` (only when genuinely pinned, not auto-resolved) and recomputes `analysis_assurance` before the exit fold. New end-to-end test reproduces the exact scenario and fails without the fix.
- **P2**: `action.yml`'s `verdict` output now declares `ANALYSIS_INCOMPLETE`, and `docs/reference/github-action-inputs.md` is regenerated.
- Two CodeRabbit doc nits (hard-wrapped inline code spans; scoped the new AGENTS.md gap entry to the verified ELF path only).

---
_Generated by [Claude Code](https://claude.ai/code/session_0148YM1JdtrGzWfUxXCBekzb)_

---

napetrov:
Investigated the `unit-tests (windows-latest, 3.13, false)` failure at `8f28fb5`. All 6 failures are in `tests/test_action_run_sh_py_safe_path.py` and `tests/test_action_run_sh_severity_summary.py`, exercising `action/run.sh`'s `$_PY_SAFE_DIR`/`$_PY_BIN` resolution and `_report_query`'s path-anchoring logic (lines ~23-24 and the `_py_safe_dir`/`_py_bin` functions near the top of the file) — none of which this PR's diff touches. This PR's `run.sh` changes are all additive, scoped to the new `require-complete-analysis`/`_assurance_gated` gating logic further down the file (around lines 1150, 1429, 1700-2050).

`test_action_run_sh_py_safe_path.py`'s own module docstring documents that this exact area (`$_PY_SAFE_DIR`'s Windows behavior) has a known history of `windows-latest`-runner-specific failures that couldn't be fully root-caused remotely (PR #774) — consistent with what's failing here: MSYS/Git-Bash path-translation mismatches (`WindowsPath(...).is_absolute()` false, `/c/Users/...` vs `C:\Users\...` anchoring) and a `mktemp` fallback assertion, not anything related to `analysis_assurance`/`aggregate` logic this PR adds.

Given the failure is confined to a pre-existing, previously-flagged fragile area this PR doesn't touch, I'm not attempting a fix here — it would be an unrelated, unverified change to `run.sh` internals under review pressure for a different PR. Will keep watching in case it reproduces on a later push or looks like something this PR's changes actually caused.


---

PR #860, opened 2026-08-25
fix: per-target evidence projection and consumer_compile consumer-context extraction

napetrov:
### Motivation

- Project cells were forwarding a single workflow-global evidence pack to every target instead of selecting each target's `targets[].evidence.path` from `build-output.json`, causing incorrect or missing source evidence for per-target checks.
- `consumer_compile` described a consumer compilation context but was not executed; the pipeline parsed candidate headers under the producer context only, preventing correct same-binary/different-client-compiler evaluations.

### Description

- Resolve per-matrix-cell evidence: `check-project.yml` now reads the downloaded `build-output/build-output.json`, looks up the current cell's target (`baseline_target` or `name`), runs `validate_build_output()` (the same check `project validate-build` runs) before trusting any declared evidence, validates confinement/existence of `targets[].evidence.path`, and emits a cell-scoped `evidence-pack` output instead of forwarding one global pack.
- Forward resolved evidence: cells forward `steps.candidate.outputs.evidence-pack || inputs.evidence-pack-path` into `check-target` so each target consumes only its declared pack, independently of `evidence-producer`, and tested against the same resolved producer value the `evidence-producer` field itself forwards.
- Reject a `projection: inferred` evidence pack outright until real attribution-based TU filtering is wired end to end (no dump/compare entry point threads `attribution`/`expected_target_id` through `ingest_inputs_pack()` today), so a build-wide inferred pack can't silently leak one target's evidence into another's check.
- Reject an embedded newline/carriage-return in the untrusted `evidence_producer.kind` value before it reaches the line-oriented `$GITHUB_OUTPUT` file.
- Consumer-context extraction: `actions/check-target/action.yml` gains inputs `consumer-ast-frontend`, `consumer-gcc-path`, `consumer-gcc-options`, and `consumer-compile-active`, and the composite action runs a dedicated `Extract candidate consumer context` dump (materializes `check-target-consumer.abi.json`) whenever a profile's `consumer_compile:` overlay is declared — including when the overlay's own derived fields (frontend/gcc-path/gcc-options) all happen to resolve empty (e.g. a `binding:`-only overlay with no `--toolchain-bindings` resolution), instead of silently falling back to the producer's own compiler options. The later `Run analysis` step consumes that snapshot instead of reparsing candidate headers under the producer context.
- Operational errors: `actions/check-target/run.sh` treats a failing consumer-context extraction as a typed operational error (emits an `operational-error` envelope), so failures surface correctly to `aggregate`.
- Documentation: reconciled `docs/reference/project-targets-schema.md`, `docs/integration/scenarios/multi-platform.md`, and the G34 plan doc so all three agree the candidate-side consumer extraction is applied end to end, with the still-open baseline-side gap (`publish-baseline.yml`/`update-main-baseline.yml` never apply a `consumer_compile:` overlay) documented consistently in each.
- Tests: `tests/test_reusable_workflows_project_evidence.py` drives the real `Resolve candidate binary/binaries` `run:` script through `bash -c` against a real `$GITHUB_OUTPUT` file and hand-built `build-output.json` fixtures (not just YAML-shape assertions) for every hardening case below.

Files changed (high level): `.github/workflows/check-project.yml`, `actions/check-target/action.yml`, `actions/check-target/run.sh`, `abicheck/buildsource/run_plan.py`, docs under `docs/reference/…` and `docs/integration/scenarios/…`, `docs/contribute/plans/g34-producer-consumer-compiler-profile-separation.md`, tests added/updated, and a `changelog.d/` fragment.

### Testing

- `pytest -q tests/test_reusable_workflows_project_evidence.py tests/test_action_check_target.py` and the broader `run_plan`/`project_targets`/`check_target`/`build_output` suite (688 tests) — all pass.
- `ruff check` / `ruff format --check` on touched files pass.
- YAML parsed successfully for the modified workflow/action files; `mkdocs build --strict` and `scripts/check_docs_contract.py` pass on the touched docs.
- `mypy` clean on touched Python modules.

## Bug-fix test contract

- Bug class: Trusting an untrusted per-target build-output.json evidence claim (or a workflow-global compiler-context fallback) without structural validation, or forwarding untrusted content unescaped into the line-oriented `$GITHUB_OUTPUT` file — a family of "accept without verifying" and "silently drop a declared override" defects in the project-cell evidence/consumer-context resolution pipeline.
- Publicly observable failure: A `check-project.yml` matrix cell would (a) silently compare a target against another target's declared pack, a shared/mismatched pack, or a build-wide inferred pack, hiding or fabricating findings; (b) let a crafted build-output.json inject an extra `$GITHUB_OUTPUT` record via an embedded newline in `evidence_producer.kind` (e.g. overriding `new-library=`), potentially comparing a baseline against itself while the resolver still exits 0; (c) silently skip the declared consumer-context extraction and parse candidate headers under the producer's own compiler flags instead of the schema-documented isolated client view; (d) silently fall back to sources-based evidence for a target whose evidence claim was itself malformed (empty path, no projection).
- Regression test fails on base: Yes. Every new/updated resolver test in tests/test_reusable_workflows_project_evidence.py executes the real run: block via bash -c against a real $GITHUB_OUTPUT file, and each was confirmed to fail against the pre-fix script -- e.g. the newline-injection test exits 0 with an injected new-library= record before the fix, and the shared-pack/manifest-mismatch/inferred-projection/empty-evidence tests all silently succeed (wrongly forward evidence-pack) before validate_build_output() was wired in.
- Negative control: test_resolver_forwards_a_validated_targets_evidence_and_producer is the happy-path control -- a genuinely valid (per validate_build_output()) build-output.json still resolves its evidence-pack/evidence-producer normally, so the hardening does not reject well-formed input; test_resolver_selects_only_the_current_targets_evidence confirms a target's own valid evidence is still forwarded and not swallowed by the new checks.
- Public-surface test: Every hardening test extracts the real run: step text from check-project.yml and executes it via a real bash -c subprocess against a real $GITHUB_OUTPUT file and a real build-output.json fixture on disk -- the same entry point GitHub Actions itself invokes -- not an internal Python helper.
- Axes covered: projection: declared and projection: inferred evidence claims; an evidence claim with a non-empty vs. empty/missing path; both LF and CR newline injection; target vs. bundle matrix-cell kinds (bundle cells stay excluded from consumer-context activation via the pre-existing matrix.kind != 'bundle' gate); the resolved-vs-legacy evidence-producer disagreement; an overlay whose derived fields are all empty vs. non-empty.
- General invariant: Any per-target evidence.path this resolver forwards as evidence-pack has passed validate_build_output() and is never a projection: inferred pack forwarded unfiltered; any value read from the untrusted build-output.json that reaches $GITHUB_OUTPUT is free of embedded newline/CR; the separate consumer-context dump activates whenever a profile's consumer_compile: overlay is declared, independent of whether its derived fields happen to be non-empty.
- Malicious fixture + side-effect absence: Yes. The resolver tests build hand-crafted, untrusted build-output.json fixtures (embedded newline/CR in evidence_producer.kind, a pack shared across two projection: declared targets, a pack whose manifest.library names a different target, a projection: inferred pack, an evidence: {} claim with no path, and the pre-existing absolute/../-escaping evidence.path case) and drive the real workflow step against them via bash -c plus a real $GITHUB_OUTPUT file, asserting both the rejection (non-zero exit, no evidence-pack output) and, for the escape cases, that a sentinel file placed outside the confined roots is left unchanged.

## Checklist

- [x] Changelog fragment added (`scriv create`) — see `changelog.d/consumer-project-evidence.fixed.md`
- [x] Docs updated (`docs/reference/project-targets-schema.md`, `docs/integration/scenarios/multi-platform.md`, `docs/contribute/plans/g34-producer-consumer-compiler-profile-separation.md`)
- [x] `ruff check` / `ruff format --check` / `mypy` pass locally on touched files

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a8df9b082b0832bbc30c34d21358f10)


napetrov (commented):
Fixed in b545646 (stacked on fbef76f).

**"Reject output delimiters in the projected producer kind" (P1)** — `evidence_producer.kind` is now validated for embedded `\n`/`\r` before being written to `$GITHUB_OUTPUT`, matching the existing pattern for resolved candidate/evidence paths. Regression coverage: `test_resolver_rejects_newline_in_evidence_producer_kind` (both LF and CR), which drives the real `run:` script through `bash -c` against a real `$GITHUB_OUTPUT` file and confirms the injected `new-library=` line never lands.

**"Validate target evidence before treating it as scoped" (P2)** — `validate_build_output()` (the same check `project validate-build` runs) is now called before trusting `targets[].evidence`, failing closed on any validation error (shared pack across targets, `manifest.library` mismatch, etc.), including catching the `FileNotFoundError`/`ValueError` it raises outright for a build-output.json that isn't even schema-valid, so that case gets a clean `::error::` instead of an uncaught traceback. Regression coverage: `test_resolver_rejects_evidence_shared_across_targets`, `test_resolver_rejects_evidence_whose_manifest_names_another_target`, and a happy-path control (`test_resolver_forwards_a_validated_targets_evidence_and_producer`) using a build-output.json that genuinely passes validation.

**"Scope inferred packs before forwarding them" (P1)** — confirmed no dump/compare entry point (nor any real caller of `ingest_inputs_pack()`) threads `attribution`/`expected_target_id` through today, so filtering a `projection: inferred` pack by target isn't actually wireable end to end without a real CLI/service feature addition. Rather than attempt that under review pressure, the resolver now rejects `projection: inferred` evidence outright (a `projection: declared` per-target pack is unaffected). Regression coverage: `test_resolver_rejects_inferred_projection_evidence`, using a genuinely valid (per `validate_build_output()`) inferred pack + attribution file, confirming the rejection is this new check and not an unrelated validation failure.

**"Test the resolved producer when routing build info" (P2)** — the build-info routing condition now tests `steps.candidate.outputs.evidence-producer || inputs.evidence-producer`, the same resolved value the `evidence-producer` field two lines below already forwards, instead of the raw legacy `inputs.evidence-producer` input. Regression coverage: `test_build_info_routing_tests_the_resolved_producer_not_the_legacy_input`.

All 16 tests in `tests/test_reusable_workflows_project_evidence.py` pass, along with the broader `run_plan`/`project_targets`/`check_target`/`build_output` suites (686 tests). `ruff check`/`ruff format --check`/`mypy` clean on touched files.

---

napetrov (commented):
Checking the remaining threads against the current head (b545646):

- **"Route projected evidence into the build-info input" (P1)** — fixed; `Run check-target`'s `build-info` now forwards `steps.candidate.outputs.evidence-pack` independently of `evidence-producer`, falling back to `inputs.build-info` (`.github/workflows/check-project.yml:1099`, and now also uses the resolved producer per the fix above).
- **"Preserve new-side replacement semantics in the consumer dump" (P2)** — fixed; `header`/`include` in "Extract candidate consumer context" now select `new-header`/`new-include` when given, falling back to the shared value, matching `compare`'s own new-side precedent (verified against `_resolve_per_side_options()` directly).
- **"Preserve the global frontend fallback for consumer dumps" (P2)** — fixed; `consumer-ast-frontend`/`consumer-gcc-path`/`consumer-gcc-options` now fall back to `inputs.ast-frontend`/`gcc-path`/`gcc-options`, gated on `matrix.consumer_compile_active` so a profile with no `consumer_compile:` overlay is unaffected.
- **"Fall back to global compiler inputs for consumer dumps" (P2)** — same fix as above; both `gcc-path` and `gcc-options` fall back correctly.
- **"Forward the compiler prefix to consumer extraction" (P2)** — already forwarded: `actions/check-target/action.yml`'s "Extract candidate consumer context" step passes `gcc-prefix: ${{ inputs.gcc-prefix }}`.
- **"Reconcile the remaining consumer-compile documentation" (P2)** — `docs/integration/scenarios/multi-platform.md`'s "Current support / Not yet supported" table and the G34 plan doc's Phase 0 acceptance criteria now agree: the candidate-side consumer extraction is documented as applied, and the baseline-side gap (next item) is documented as open in both places.
- **"Generate baseline snapshots under the consumer context" (P1)** — this is the one item here that's a genuine, still-open gap rather than something to patch reactively. `publish-baseline.yml`/`update-main-baseline.yml` read only `build-output.json` and never consult `run-plan.json`, so they have no per-profile `consumer_compile:` field to apply. Closing it needs `actions/baseline` and `derive_baseline_libraries()` to gain the same per-library compile-context resolution `check-target` already has, sourced from a real per-target run-plan read neither baseline workflow performs today. Documented as "Still open" in both `multi-platform.md` and the G34 plan doc (Phase 0's acceptance criteria) rather than left silently unaddressed — a `consumer_compile:` check against a real baseline channel resolves `NOT_COMPARABLE`/`ProfileMismatchError` today (the comparability gate correctly refusing a genuine profile mismatch), not a silently wrong verdict, so this is a coverage gap rather than a correctness bug. `baseline-channel: none` (audit-only) is unaffected. Leaving this thread open since it names a real follow-up rather than something this PR should attempt.

---

napetrov (commented):
Fixed in 3dee35a. Added a new `consumer-compile-active` input to `check-target` (forwarded from `check-project.yml` as the overlay's own presence marker, mirroring `run_plan.py`'s `consumer_compile_active` field independently of what its three derived fields resolved to), and OR'd it into the "Extract candidate consumer context" step's activation condition — so a `consumer_compile: {binding: clang20}`-shaped overlay with no `--toolchain-bindings` resolution and no workflow-global fallback still triggers the separate candidate dump instead of silently falling through to the producer's own compiler options.

Also fixed a stale self-contradiction this investigation turned up in `docs/reference/project-targets-schema.md`'s own `consumer_compile:` section (it still said extraction/merge was "not yet wired" three paragraphs above the section documenting it as applied end to end).

Regression coverage: `test_consumer_dump_activates_from_the_overlay_marker_alone`. Full related suite (`tests/test_reusable_workflows_project_evidence.py` + `tests/test_action_check_target.py`, 74 tests) and the broader `run_plan`/`project_targets`/`check_target`/`build_output` suite (687 tests) pass; `ruff`/`docs-contract`/`mkdocs build --strict` clean.

---

napetrov (commented):
Fixed in ac6d3a5. The resolver now gates `validate_build_output()` (and the projection: inferred rejection) on the target declaring an `evidence` **object** at all, not on that object happening to carry a non-empty `path` — so `evidence: {}` (or any malformed claim with an empty/missing path) is validated and fails closed instead of silently falling through to the "no evidence for this target" branch.

Regression coverage: `test_resolver_validates_an_evidence_claim_with_an_empty_path` (an `evidence: {}` target, confirmed to fail on `evidence.projection` before this fix would have let it pass with `evidence-pack=` empty). Full suite (`tests/test_reusable_workflows_project_evidence.py`, 18 tests; broader `run_plan`/`project_targets`/`check_target`/`build_output` suite, 688 tests) passes; `ruff`/YAML syntax clean.

---

PR #881, opened 2026-08-26
refactor(model): make abicheck.model a package and split fact dataclasses from their parsers

napetrov:
## Summary

ADR-061 Phase 5, first slice: `abicheck.model` becomes a package, and the binary/debug fact dataclasses move into it from the parsers that used to own both halves.

## Motivation

ADR-061's Phase 4 section records `service.py` as blocked, with the reason stated concretely: nine `*_metadata.py` modules conflate a *model dataclass* with its *parser*. `AbiSnapshot` has typed fields of `ElfMetadata`/`PeMetadata`/`MachoMetadata`/`DwarfMetadata`/`AdvancedDwarfMetadata`, so classifying those modules `extract` — which is what their parsers are — makes `model -> extract` and `storage -> extract`, both forbidden. The ADR names the fix as Phase 5's "parsers and catalogs" scope.

Creating the `model` package **forced** that split rather than merely permitting it, which is worth stating because it determined the shape of this PR: `abicheck/model.py` and `abicheck/model/` cannot coexist (Python resolves one import name, and `check_architecture.py` has a `module-package-collision` finding saying so), and the flat `model.py` imported `elf_metadata` at runtime for `SymbolBinding`. The dependency that blocked Phase 4 was also the dependency that blocked opening the package Phase 5 needs.

## Changes

- **`abicheck/model.py` (1208 lines) is now `abicheck/model/`**, split by responsibility into `vocabulary`, `declarations`, `entities`, `extraction_contract`, `snapshot` and `stdlib_surface`. None is above ADR-061's 800-line ceiling, so the flat module's `debt.yaml` entry is **retired**, not lowered. `__init__.py` is a re-export list with `__all__`; every name the flat module exported still resolves from `abicheck.model`.
- **Eleven modules kept their parsing and handed their dataclasses to `model/*_facts.py`**: `elf_metadata`, `pe_metadata`, `macho_metadata`, `dwarf_metadata`, `dwarf_advanced`, `sycl_metadata`, `symvers_metadata`, `python_api`, `python_ext`, `numpy_capi`, `build_mode`. Each imports and re-exports its own types, so `from abicheck.elf_metadata import ElfMetadata` and every sibling spelling are unchanged. `elf_metadata.py` 1366 → 1193; `dwarf_advanced.py` 1300 → 1249; both baselines lowered.
- **Internal `abicheck/` callers switched to the model owner** (ADR-061 migration rule 3) — 15 files. The parser re-export is compatibility surface for external callers and tests, not the internal path.
- **`AbiSnapshot.index()` now builds its three lookup maps through one primitive.** It carried three hand-written copies of "index by key, first wins, warn about what was dropped"; that rule is now `model/first_wins_index.py`. Behaviour is unchanged, including the wording and the `×N` counts of the duplicate-symbol warnings.
- **`abicheck/model/AGENTS.md`** gives the package its scoped routing contract, including the two rules that are easy to get wrong here (a new `AbiSnapshot` field is a storage event; never add `frozen=True` to a fact class carrying `@cached_property`).
- **Contract and docs**: `modules.yaml` drops `abicheck/model.py` from `model`'s `legacy_paths` and classifies `name_classification.py`, `buildsource/model.py` and `buildsource/pack.py`; ADR-061's status, Phase 5 section and index row record what landed; root `AGENTS.md` routes a new fact's field to `model/` rather than to the parser.

### One residual, recorded rather than papered over

`AbiSnapshot.build_source` is typed `BuildSourcePack`, whose class is five data fields plus `load`/`write`/`content_hash`/`verify_integrity` — persistence, which ADR-061 assigns to `storage`, and `model` may not import `storage`. `buildsource/pack.py` is therefore classified `model` by *target owner* with the persistence split noted in the ADR as still owed. Splitting it means separating a dataclass from its own load/write methods without changing what `BuildSourcePack(...)` constructs for callers this repository cannot see, which is its own slice.

## Checklist

- [ ] `pre-commit run --all-files` passes locally — `fmt-check` fails repo-wide for the pre-existing reason `AGENTS.md` documents (the tree has never been `ruff format`ed; 568 files). The files this PR adds or touches are format-clean.

## Verification

- `python scripts/check_architecture.py` — 0 errors
- `mypy abicheck/` — clean, 459 source files
- `ruff check abicheck/ tests/` — clean; `ruff format --check` clean on every file this PR adds
- `pytest -m golden` — 10 passed, output unchanged
- New tests: `tests/test_first_wins_index.py` (13, the primitive's contract as Hypothesis invariants), `tests/test_model_package_surface.py` (42, the flat-module export surface, each parser's re-export identity, that no parser still *declares* a moved type, and that no `model` module imports an outer package)

---
_Generated by [Claude Code](https://claude.ai/code/session_014vbBGMtNJL74iamFsQDT7s)_


napetrov:
Pushed `f2d118b`, fixing three things a full local unit run turned up. Two were mine; one was not, and that one is worth stating explicitly.

**Mine, fixed:**

- `test_docs_contract_has_no_ownership_errors_on_real_repo` — `docs/_meta/topics.yaml`'s `baseline-lifecycle` topic listed `abicheck/model.py` as a fact source, and that check resolves entries against the real tree. Repointed at `model/snapshot.py`, along with the two front-matter copies of the same list and the prose reference in `docs/reference/snapshot-format.md`.
- `test_duplicate_mangled.py` (2 cases) — they patched `logging.getLogger("abicheck.model")` by exact name. `AbiSnapshot` now lives in `abicheck.model.snapshot`, so its logger is that submodule's. Worth being precise about the blast radius: the name stays *under* the `abicheck.model` hierarchy, so logging configuration keyed on the package is unaffected — only an exact-name patch or an exact-name record filter sees the difference, which is what ADR-061 means by moving internal tests to the implementation owner.

**Not this PR's:** `test_generated_reference_is_in_sync_with_service_all` fails on the base commit (`c599daf`) too — verified by checking out the base and running it there. `ff03fd3` (#872, the same-binary warning work) changed the `collect_metadata` and `sniff_text_format` docstrings without regenerating `docs/reference/python-api-reference.md`. No fix PR exists for it, but the fix is the regeneration command the test itself names, so I ported it rather than waiting: the diff is exactly those two docstrings and touches nothing in this PR's scope.

Gates after the push: `check_architecture.py` 0 errors, `mypy abicheck/` clean (459 files), `ruff check` clean, and the four affected test modules plus the two new ones green (310 passed). The full unit suite is re-running locally.

---

napetrov:
## `mutmut (detector core)` is failing, and it is not this PR's

**What failed:** not a mutant — mutmut's **clean test** run, the unmutated baseline pass it does before generating anything. The lane aborts there:

```
mutation-score: mutmut run tail:
  -     'layout',
  - )
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
Failed to run clean test

ERROR: `mutmut run` exited 1 — the run aborted, so any results on disk are from
an earlier run, not this one. Not reading them.
```

So the lane never measured a single mutant, and its verdict says nothing about this PR's mutation resistance either way. (The guard behaving correctly, incidentally — a failed measurement failing loudly rather than reporting zero survivors is exactly what `--require-baseline` and this error path exist for.)

**Why it is not this PR's:**

- The failure tail is **byte-identical** on three other branches: [`storage-format-plan`](https://github.com/abicheck/abicheck/actions/runs/33001585865) (#879), [`mutation-testing-optimization`](https://github.com/abicheck/abicheck/actions/runs/33003044294) (#877), and [`same-binary-coverage-warning`](https://github.com/abicheck/abicheck/actions/runs/33005410707) (#872) — same two diff lines, same `Failed to run clean test`.
- The lane last succeeded at **17:34 UTC** and has failed on **every** run since **18:08 UTC**, on branches that predate this one. This PR's branch was opened at 19:57 UTC.
- #872 and #879 both merged into `main` with this lane red, so it is red on the base branch now.
- This PR's own diff touches `only_mutate` modules (`diff_platform.py`, `diff_symbols_renames.py`, `suppression.py`, …) in **import lines only** — and a clean-test failure is upstream of mutation anyway.

**No fix exists to port.** No PR is open against it, so there is nothing I can bring into this branch that would no-op once merged.

**A lead for whoever picks it up**, stated as far as I verified it and no further:

The assertion repr tail (`'layout',` then a closing `)`) matches `TestMissingFamilies` in `tests/unit/storage/test_availability.py` — both cases there compare a tuple ending in `"layout"`. That file passes locally (121 passed), passes with a fresh Hypothesis database, and passes in every `unit-tests` lane on this PR (all four green). So the difference is mutmut's own environment: it runs pytest with `mutants/` as the working directory over a copied tree, and something about that copy changes the result.

I checked and **ruled out** the obvious candidate: `tests/unit/storage/adr062_scope.py` resolves `pathlib.Path("abicheck/storage")` relative to the cwd, which does exist inside `mutants/`. So that is not the mechanism, and I have not identified what is.

Everything else on this PR is green — 79 of 80 checks, including all four `unit-tests` lanes, `lint-and-types`, `ai-readiness`, both codecov checks, and the integration/parity lanes. I have re-run this job once; if it reproduces, it needs its own fix on `main` rather than anything on this branch.

---

PR #886, opened 2026-08-27
docs(plans): add G41-G45 plans and update G38/G39 for upstream review

napetrov:
## Summary

Turns an external upstream-only review (base commit `327df7b5616bcfaea8c330aad418b796c17f3970`, PRs #860/#883 merged) into tracked plans and one direct documentation fix.

## Motivation

The review confirmed two previously-tracked declarative-project blockers are now closed (per-target evidence routing, candidate-side `consumer_compile` extraction), but found 15 narrower, still-open items across correctness, functionality, and hardening. This PR creates/updates the plan documents needed to track all of them, cross-referenced against work this repo has already shipped (G30/G33/G34/G38/G39, the root `AGENTS.md`'s known-gap entries) so nothing duplicates what's already done, and fixes the one item (a stale doc claim) that was quick enough to correct directly instead of merely planning.

## Changes

- **New plan G41** — baseline/candidate consumer-compile context parity, per-target header/compile-context projection, a declarative assurance contract, and `DumpRequest` execution convergence. The review's P0 items 1-4, treated as one coordinated correctness initiative ("old and new sides extracted from the same resolved target, headers, compiler context, evidence context, and assurance contract").
- **New plan G42** — explicit check identities, named deployment environments, and environment-aware system-provider resolution (review items 5, 6, 12).
- **New plan G43** — safe projection via real TU-to-target evidence attribution, extending G30/ADR-053's already-implemented attribution core (review item 9).
- **New plan G44** — a pybind11/nanobind binding-ABI provider (review item 11).
- **New plan G45** — header-only project targets and a validated `build-output.json` producer helper (review items 13, 14).
- **G38 update** — new Phase 14 (decouple the three diff-derived bundle detectors from public-surface-scoping starvation of bundle-internal findings) and Phase 15 (the still-open `check-project.yml`/Action/CLI wiring for the already-shipped `BundleFacts`/`bundle_variants` Python-API layer) — review items 7, 8.
- **G39 update** — new Phase 5 (pack-level evidence-producer receipt, a prerequisite for a fail-closed declarative-project consumer to trust per-finding provenance tags) — review item 10.
- **`docs/use/multi-binary.md` fix** — corrected the now-stale claim that a custom `--policy` document never reaches bundle verdicts. Verified against current source: PR #883 added `compare_bundle()`'s `policy_file` forwarding (`BundleDiffResult.bundle_verdict` now calls `policy_file.compute_verdict(...)` when supplied), and `cli_compare_release_helpers.py`'s release fan-out already threads it through for both the bundle result and the build-config matrix result.
- **`docs/contribute/plans/index.md`** — new rows for G41-G45, refreshed status notes for G38/G39.

This is a docs-only change — no `abicheck/**/*.py` touched, so no changelog fragment is needed, and the bug-fix test contract section doesn't apply (not a `fix:`/`perf:`/`security:` PR).

## Checklist

- [x] Docs updated (this PR *is* the docs update)
- [x] Verified against current source before editing (`compare_bundle`/`BundleDiffResult.bundle_verdict`, `check-project.yml`'s per-target evidence routing, `RunPlanCheck`'s `consumer_compile_*` fields, `project_targets.py`'s `kind: library` validation, `python_ext.py`'s pybind11/nanobind recognition)
- [x] `python scripts/check_docs_contract.py` — 0 errors (2 pre-existing, unrelated warnings)
- [x] `python scripts/check_ai_readiness.py` — 0 errors (pre-existing warnings only)
- [ ] `mkdocs build --strict` — not run in this environment (mkdocs/docs extras not installed); links were checked manually and via `check_docs_contract.py`, which validates path existence

---
_Generated by [Claude Code](https://claude.ai/code/session_01FCjY2KUXq1dmePKayJ9XoA)_


napetrov (commented):
Thanks — verified each of these against current source and pushed fixes in 55ea38c:

- **`docs/use/multi-binary.md` policy-file overclaim**: confirmed. `_collect_bundle_result()`/`_run_bundle_analysis()` (the `compare-release` fan-out) only pass the bare `policy` string to `analyze_bundle()`; only the stored-`BundleFacts` Python-API driver (`bundle_facts.compare_bundle_from_facts`, `bundle_side_input.compare_bundle_sides`) threads a resolved `PolicyFile` through. Rewrote the doc to say so precisely, and tracked the actual fix as new G38 Phase 16.
- **G44 canonical package routing**: rewrote "Files & surfaces" to route the new binding-ABI extractor/model/matcher through `extract/`/`model/`/`compare/`/`workflows/`/`policy/` per ADR-061, with only a minimal recognition hook left in the legacy `python_ext.py`/`scan_engine.py` entry points.
- **Front matter on G41/G42/G44/G45**: added (`doc_type: contributor`, `level: expert`, `lifecycle: active`, `generated: false`), matching the pattern already used by G34/G39.
- **G43 duplicates already-shipped attribution work**: confirmed by reading `link_attribution.attribute_sources_to_targets()`, `build_output._inferred_evidence_projection_issues()`, and `inputs_pack.ingest_inputs_pack()` — the attribution model and its validation already exist. Rewrote G43 from scratch to scope it to the actual remaining gap (the `check-project.yml`/CLI wiring ADR-053 itself defers), revised effort L → M.
- **G43 multi-target attribution**: confirmed `_filter_tus_by_attribution()` already keeps a TU for every target its attribution set names. Corrected the rewritten G43 to state this is valid shared evidence, not an ambiguous/incomplete case — only an attribution-absent source is incomplete.
- **G45 build-output validation gap**: confirmed `BuildOutputTarget` has no kind discriminator and `_binary_issues()` unconditionally requires a binary. Added the required `build_output.py` model/validator changes to G45's design and files sections.

`check_docs_contract.py`/`check_ai_readiness.py` both clean after the fixes.

---

napetrov (commented):
Second round, all addressed and pushed in d9ec195. Verified against `architecture/modules.yaml` before applying — `extract/` and `policy/` are declared target layer paths there (`legacy_paths` point at today's flat modules) even though the physical directories don't exist yet, so routing new work to them is correct, not premature:

- **G41 Phase 1 → `storage/`**: routed the new baseline-manifest schema/serialization/selection-key logic to `abicheck/storage/` (ADR-061's owner for snapshot/baseline schemas), keeping `buildsource/baseline_publish.py`/`baseline_set.py` as orchestration-only callers.
- **G38 Phase 14 → `compare/`/`workflows/`/`frontends/`**: routed the new unscoped bundle-detector matching to `compare/`, coordination to `workflows/`, and CLI plumbing to `frontends/`, instead of growing `bundle.py`/`bundle_side_input.py` (both in `legacy_root_modules`) or `cli_compare_release.py` (a `frozen_root_families["cli_"]` entry).
- **G42 provider resolver → `extract/`/`model/`/`compare/`/`policy/`/`workflows/`**: routed accordingly instead of `bundle.py`.
- **G45 emit helper → `storage/`/`frontends/`**: routed the schema/write path and CLI adapter accordingly, instead of a new `buildsource/build_output_emit.py` + growing `cli_project.py` directly.
- **G45 binary-less check execution**: confirmed by reading both files — `actions/check-target/action.yml` declares `new-library: required: true`, and `run.sh`'s compare-mode dispatch does `CMD+=("${INPUT_NEW_LIBRARY:?new-library is required}")`, a hard failure with no operand. Added the `action.yml`/`run.sh` changes needed for a binary-less check to the plan's scope and effort estimate.
- **Stale Phase 15 → 16 link**: fixed (the policy-forwarding gap I documented last round is Phase 16, not 15).

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Third round, all verified and pushed in 76b571e:

- **G44 → `SurfaceProvider`**: confirmed both `docs/contribute/scipy-scientific-python-roadmap.md`'s "Recommended architecture" section and `docs/contribute/python-build-ecosystem-positioning.md`'s "Proposed provider shape" section already specify this exact `identify`/`collect`/`compare`/`coverage` contract and name `BindingAbiProvider` as an intended implementation. Rewrote G44's comparison-semantics section to adopt it, including the ADR-032/ADR-034 evaluation prerequisite both docs call for.
- **G45 CLI gap**: you're right that my earlier "confirm it already works" framing was wrong. Read `frontends/cli/commands/dump.py`/`compare.py` directly: `dump_source_only()`'s own code comment confirms `-H`/`--header` is silently discarded (no L2 pass at all) for a binary-less dump, and `compare.py` has no header-only operand shape. This is genuinely new implementation work, not confirmation — bumped G45's effort from M to L.
- **G43 three-layer Action chain**: confirmed by reading the `uses:` steps — `check-project.yml` → `actions/check-target` → repository-root `action.yml`/`run.sh`. Added the two missing intermediate layers to the design and file map.
- **G42 environment schema**: confirmed `environment_matrix.py`'s `_KNOWN_TOP_LEVEL_KEYS` has no sysroot/provider field and silently ignores unknown keys. Added the required schema extension to the plan.
- **G41 → `workflows/`**: routed Phase 2's per-target header/compile-context generation and Phase 3's aggregate failure-class distinction to `workflows/`, per ADR-061 naming `project`/`aggregate` coordination as its explicit responsibility, instead of `buildsource/run_plan.py`/`cli_aggregate.py` directly.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Fourth round, all verified and pushed in 959fa6f:

- **G42 check-id parsing**: confirmed a real bug — `workflows/aggregate/contracts.py`'s `_CHECK_ID_RE` is end-anchored on the depth segment (`@(?P<depth>binary|headers|build|source)$`), so a plain trailing suffix would make `parse_check_id()` return `None` and silently drop the check from grouping. Redesigned as a backward-compatible `~<explicit_id>` regex extension and added `contracts.py` to the plan's required file surface.
- **G42 environment aggregation → `workflows/aggregate/`**: confirmed that package already owns `finding_matrix`; routed there instead of `cli_aggregate.py`.
- **G45 stale precedent**: removed the "mirrors an existing `--depth headers` path" phrasing — you're right that the sibling finding a few threads earlier already established no such path exists; reworded so the Action relaxation and the new CLI capability are explicitly one dependent pair.
- **G39 Phase 5 → `model/`/`storage/`**: routed the receipt schema/serialization there, matching the same fix already applied to G41/G45, with `buildsource/build_output.py` kept to orchestration.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Fifth round, all verified and pushed in 3892da1:

- **G42 → `extract/`/`model/`**: confirmed `environment_matrix.py` is itself a `legacy_root_modules` no-growth entry; routed the new provider/sysroot parser/model there instead of extending it directly.
- **G45 wrong file attribution**: you're right — the `new-library is required` enforcement lives in the repository-root `action/run.sh`, not `actions/check-target/run.sh` (which only forwards inputs into the nested root Action). Confirmed by reading both directly and split the bullet into three, correctly attributed.
- **G41 Phase 1 `actions/baseline`**: confirmed by reading `actions/baseline/run.sh` directly — it builds every dump command from only `libraries`/`build-info`/`depth`, no compiler/frontend input exists on `action.yml` at all, so workflow-level `consumer_compile_*` resolution is inert without new Action inputs. Added to the plan's required file surface.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Sixth round, all verified and pushed in da9764a:

- **G45 contradictory depth rule**: fixed — restated as one rule: only `depth: binary`/`symbols` need a compiled artifact and are rejected; `headers`/`build`/`source` are all valid for a header-only target (`depth: build` meaning ordinary L3 build evidence, no binary required).
- **G43 `inputs_pack.py`**: confirmed a real gap — `_filter_tus_by_attribution()`'s "attributed elsewhere" and "unresolvable" drop reasons collapse into one count kept out of `diagnostics`, so `status` stays `"ok"` even when ownership was genuinely unresolvable. That's exactly what G41 Phase 3's assurance gate needs to catch and can't today. Corrected the "no new logic expected" claim and scoped the actual fix needed.
- **G45 `actions/baseline`**: confirmed by reading `run.sh` directly — it hard-rejects any library entry without `artifact` and always runs `abicheck dump "$artifact"`, so neither baseline workflow can publish a header-only target's baseline today. Added to the file inventory, distinct from G41 Phase 1's separate `consumer_compile_*` gap on the same Action.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Seventh round, all verified and pushed in 14fb62c:

- **G42 mutation bug**: confirmed — `diff_versioning.apply_runtime_floor_contract()` mutates `Change.effective_verdict` in place and skips findings already carrying a modulation, so reusing the same `Change` objects across environments would permanently stamp the first environment's verdict. Redesigned to require a pristine `Change` list per environment.
- **G45 storage/extract boundary**: confirmed via `abicheck/storage/AGENTS.md` — compiler/tool-identity discovery is extraction-layer probing and can't live in `storage/`. Moved it to `extract/`, coordinated by `workflows/`.
- **G41 baseline key**: added requested depth and evidence-producer identity as explicit key components — `profile_fingerprint`/`scope_fingerprint` cover compile context and headers/TUs, neither of the other two axes.
- **G39 Phase 5 artifact conflation**: confirmed by reading `build_output.py`/`inputs_pack.py` directly — G43's `attribution_path` and this phase's receipt are three separate artifacts, not two sections of one manifest. Rewrote to keep them separate, with the receipt's real home being `InputsManifest` (`abicheck_inputs/manifest.json`), not `build-output.json`.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Eighth round, both verified and pushed in 44999f9:

- **G38 Phase 14 sibling-import gate**: confirmed — `_detect_provider_changed()` (`abicheck/bundle.py`) emits `bundle_provider_changed` unconditionally on any cross-library symbol-name move, with no sibling-`DT_NEEDED`-import check today, unlike its two diff-derived siblings. Applying the sibling-import reachability gate to all three would have silently dropped this finding's existing protection for an external consumer linked to the old provider. Restricted the gate to `bundle_intra_dep_signature_changed`/`bundle_intra_type_changed`; kept `bundle_provider_changed`'s unconditional promotion, and added a second acceptance test pinning a provider move with zero sibling consumers still promotes.
- **G43 unresolved-TU coverage signal**: named the full propagation path instead of leaving it as an unhomed "structured signal" — `_filter_tus_by_attribution()` returns the two drop reasons separately, `IngestedInputs` gains a new `unresolved_attribution_tu_count` field (diagnostics/status untouched), threaded into `contract_coverage_ledger.py`'s existing unsuppressible ledger shape, consumed by G41 Phase 3's assurance contract the same way as every other ledger failure.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Ninth round, both verified and pushed in 63d8f22:

- **G42 check-id validator chain**: confirmed — `build_check_id()` (`abicheck/buildsource/check_report.py`) calls `checker_types.validate_check_id()` unconditionally, and `CHECK_ID_PATTERN` there carries the identical end-anchored regex, with `schemas/compare_report.schema.json`'s `check_id` pattern as a third independent copy. Updating only `contracts.py`'s parser (as the previous round left it) means `build_check_id()` itself would raise before the parser ever runs. Extended the plan to require all three — the Python validator, the builder, and the JSON schema pattern (with version bump and `docs/reference/schemas/v1` re-sync) — in lockstep, plus a full construct→validate→parse acceptance test.
- **G45 header-only model routing**: confirmed the `BuildOutputTarget` discriminator/schema work was routed directly at `abicheck/buildsource/build_output.py` (a `legacy_root_modules` no-growth entry), inconsistent with the identical `model/`/`storage/`/`workflows/` split this same plan already applies to the separate emit-build helper a few paragraphs later. Applied the same routing consistently.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Tenth round, all three verified and pushed in adece4b:

- **G45 header-only workflow step**: confirmed — `check-project.yml`'s "Resolve candidate binary/binaries" step unconditionally calls `resolve(cell.get('binary_pattern', ''), ...)` for every non-bundle cell, and an empty/absent pattern makes `resolve()` return `None`, which is treated as a hard `sys.exit(1)` failure before `actions/check-target` (or the root Action's relaxation) ever runs. Added the missing header-only branch to this step as a required file surface.
- **G43 baseline-side attribution**: confirmed — `publish-baseline.yml`/`update-main-baseline.yml` pass one workflow-global `build-info` straight through `actions/baseline`, which forwards it unconditionally to every library's dump. No per-library attribution/target-id input exists on the baseline side. Wiring only the candidate chain leaves the baseline unscoped whenever the old release also used a shared/inferred pack. Added the baseline-side wiring as a required, previously-missing half.
- **G42 job-boundary extract-once**: confirmed — `check-project.yml`'s matrix is one `RunPlanCheck` → one job → one full dump/compare. Two checks differing only by `environment_id` become two independent jobs with two independent extractions today; the plan's "extract/diff exactly once" constraint only covered the in-process mutation-safety fix, not this job-orchestration boundary. Redesigned `RunPlanCheck`'s environment field as a grouped list, added the required run-plan grouping pass and a multi-environment fan-out mode inside one job, and revised the phase's effort from M to L.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Eleventh round, both verified and pushed in fd866cf (both on G42):

- **Grouping key**: confirmed grouping solely by (target, profile, channel, requested_depth) would also collapse checks differing only in `analysis_evidence`/`analysis_policy`/`analysis_assurance_requirement`/explicit `id` — exactly the plan's own replay-vs-clang-plugin acceptance-test pair would wrongly merge. Widened the grouping key to the full non-environment tuple.
- **Per-environment report identity**: confirmed `abicheck/workflows/aggregate/execute.py`'s `collect_reports()` indexes loaded reports by the in-report `target_id` field, not artifact filename, and raises on duplicates. A grouped check's N environment reports would all carry the identical `check_id` unless environment is baked into the id — the previous round's "unique artifact names" claim was wrong. Added a second, distinct check-id qualifier segment (composable with the existing optional `~explicit_id` tail), mandatory for multi-environment groups, plus the required expansion of `ExpectedTargets`' manifest projection to one entry per environment per grouped check.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Twelfth round, all five verified and pushed in a3cdaa5:

- **G43 assurance routing**: confirmed `contract_coverage_ledger.py`'s ledger returns `()` outright with no `--contract` evaluation, while G41 Phase 3's `require_complete_analysis` is a general per-check floor with no such precondition. Rerouted to `analysis_assurance.compute_analysis_assurance()`, the actual general mechanism.
- **G39 shared-pack identity**: confirmed a singular "target id" receipt field conflicts with G43's own shared/inferred-pack scenario. Redefined as projection-aware: single target id or a shared/build-wide marker paired with G43's attribution digest.
- **G41 baseline fan-out**: confirmed `actions/baseline` produces exactly one snapshot per library name regardless of how many depth/evidence-producer contexts the widened selection key distinguishes. Added the required publication fan-out as new orchestration logic, plus a regression test.
- **G38 type-detector reachability**: confirmed `_detect_intra_type_changed()` has no import-resolution check at all — its own docstring documents a name-embedding symbol-table match, unlike its sibling's real import-graph check. The previous round wrongly merged both under one gate. Corrected each detector to keep its own existing mechanism, plus a third acceptance test for the name-embedding-without-import-edge case.
- **G44 binding discriminator**: confirmed `detect_python_extension()`'s own docstring states its recognition is deliberately uniform across Cython/pybind11/nanobind/hand-written C. No framework-specific discriminator exists anywhere in the codebase. Specified `python_ext` as only the broad prefilter, with a new, separately-validated discriminator as real new logic.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Thirteenth round, all three verified and pushed in a500c3a:

- **G43 side-specific attribution**: confirmed a single global attribution path/target-id flag can't correctly scope a two-sided `compare`/`scan` invocation where both `old=`/`new=` use pack-shaped build info — attribution is release-specific. Made the CLI/typed-API flag side-aware, reusing this codebase's existing `old=`/`new=` scoping convention. The workflow/Action chain stays single-sided by construction and is unaffected.
- **G42 gate fields**: confirmed `RunPlanCheck` also carries `required`/`gate_mode`/`allow_new_target` — per-check gating fields, not extraction-relevant ones — which the grouping key didn't cover. Widening the key would defeat grouping for the common "same extraction, different gating per environment" case. Moved these three fields into a new per-environment `EnvironmentEvaluation` value instead.
- **G45 depth spelling**: confirmed `symbols` isn't a canonical depth (`binary|headers|build|source`) and is explicitly rejected by the raw `--depth` CLI parameter. Corrected both mentions to state only `binary` is invalid for a header-only target.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Fourteenth round, both verified and pushed in 145d02d:

- **G43 persistence**: confirmed `BuildSourcePack.to_embedded_dict()`/`from_embedded_dict()` enumerate exactly four persisted keys — a plain in-memory field is silently dropped on snapshot embed/reload. If only the old baseline side had an unmapped TU, reloading it would read the count as zero and could report complete assurance for exactly the case this signal exists to catch. Moved the field's real home to `BuildSourceManifest` (already unconditionally serialized, already version-bumpable), plus a stored-baseline round-trip regression test.
- **G42 validator inventory**: confirmed the `~explicit_id` extension and the later `!environment_id` extension were described in two separate places with no single composed pattern shown — so following the doc literally would leave `build_check_id()`/`CHECK_ID_PATTERN`/the JSON schema still rejecting a real environment-qualified report. Added one canonical, fully-composed `_CHECK_ID_RE` definition and pointed every extension site at it.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Fifteenth round, all three verified and pushed in a6c2fd4:

- **G43 canonical attribution identity**: confirmed `attribute_sources_to_targets()` records identities as `target://<id>`/`output://<basename>`, never a bare target id, and `_filter_tus_by_attribution()` does exact set membership. Forwarding the plain project target id would silently drop the whole pack's contribution to the target. Fixed to resolve and forward the same canonical identity `_inferred_evidence_projection_issues()` already computes.
- **G42 no-environment representation**: confirmed the "Environments" acceptance test needs an unconfigured result alongside two named-environment results from one extraction, but the descriptor had no way to represent "no environment" without a separate ungrouped job or a fabricated id. Made `environment_id` explicitly nullable with defined digest/check-id behavior for that entry.
- **G42 digest content binding**: confirmed a digest computed only from resolved matrix/providers YAML says nothing about what a live sysroot probe actually found — two runners can share identical YAML and an identical digest while resolving against physically different libraries. Required the digest to incorporate the actually-probed provider facts, failing closed on an unreachable sysroot rather than falling back to config-only.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Sixteenth round, all three verified and pushed in 0e18c26:

- **G43 accepted identity set**: confirmed `_inferred_evidence_projection_issues()` validates against a *set* of accepted identities (target-graph and link-unit channels can cover different subsets of one target's own TUs), while the previous round forwarded only one resolved identity string to `_filter_tus_by_attribution()`'s exact-membership check. Widened the signature to accept the full identity set — a real, narrow change — and updated every forwarding site.
- **G39 undefined digest**: confirmed no code anywhere computes an "attribution digest" — the previous round referenced one as already existing. Defined it properly: canonical normalization, a reused hashing primitive, a persisted field, and producer/consumer wiring.
- **G42 digest timing**: confirmed the `plan` job (always `ubuntu-latest`, writes `run-plan.json` before any check job runs) cannot probe a sysroot that exists only on the check job's own runner. Split into two digests: `RunPlanCheck.environment_digest` stays config-only (plan-time), and a new `probed_environment_digest` is computed by the check job itself post-probe and stamped onto its own report.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Seventeenth round, both verified and pushed in 1fc4185:

- **G44 cross-release comparison**: confirmed `compare(old, new, policy)` was defined as matching two sibling modules within one release, discarding the old-release/new-release dimension the shared `SurfaceProvider` contract represents — a pre-existing mismatch unchanged across releases would be re-emitted as new every time. Split into a per-release module-pair graph (mirroring G38's bundle-internal shape) plus a genuine cross-release diff of that graph, with a second acceptance test for the cross-release axis.
- **G42 explicit id in grouping key**: confirmed explicit `id` was left in the extraction key — the identical mistake already corrected for `required`/`gate_mode`/`allow_new_target` — meaning two environment checks given distinct ids could never collapse into one `RunPlanCheck`. Moved it onto `EnvironmentEvaluation` alongside the other per-environment fields.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Eighteenth round, both verified and pushed in d54df35:

- **G42 unqualified no-environment report**: confirmed the "mandatory qualifier whenever a group has more than one environment" rule directly conflicted with the `None` entry, which must stay unqualified in every group shape — the acceptance test's own mixed group needed exactly this. Restated so only named-environment entries get `!<environment_id>`, with `None` always producing the plain unqualified `target_id`, and fixed the `ExpectedTargets` expansion to match.
- **G44 pair creation/deletion**: confirmed the module-pair graph diff only covered an already-existing pair flipping agree/disagree, missing pair creation entirely — a new release adding a module that shares a type with an existing one under incompatible internals produces a disagreeing pair with no old counterpart, and the stated rule never fires for it. Added the full transition table and a fourth acceptance-test variant.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Nineteenth round, verified and pushed in 8a57881:

- **G42 environment-dependent checks re-run**: confirmed `checker._env_matrix_contract_changes()` has two halves — `apply_runtime_floor_contract()` reclassifies an existing finding in place (the mutation case already fixed), but `check_platform_baseline_floor()`/musllinux/the wheel-deployment checks are standalone, reading raw evidence directly and producing brand-new findings per environment's own floors. Resetting/copying `Change` objects has nothing to reset for a finding these functions haven't produced yet — an unchanged binary always exceeding its manylinux floor would be reported for only the first environment and silently omitted from the rest. Fixed by requiring the fan-out to rerun the complete step per environment against the raw, already-extracted snapshot, preserving extract-once for the extraction stage itself.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Twentieth round, both verified and pushed in 647f1a1:

- **G39 per-projection receipt fields**: confirmed `compile-context fingerprint`/`public-header-root digest` were still singular pack-wide fields even after identity was made projection-aware — a shared pack has no reason its consuming targets share one compiler context or header set. Made both (and TU inventory) per-canonical-projection fields.
- **G38 Phase 15 remaining scope**: confirmed `--bundle-facts-out` is a `compare-release`-only producer with no path from separate `check-project.yml` per-member matrix cells, and `BundleFacts` stores only per-library snapshots — no `DiffResult`/assurance field at all. A later bundle dispatch has no candidate snapshots, baseline facts, or per-member results to build from. Rewrote the five-step accounting to make this real assembly work explicit, revised effort from M to L.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Twenty-first round, both verified and pushed in dffe60e:

- **G42 singleton check_id collision**: confirmed `run_plan.py`'s existing duplicate-check_id guard rejects the plan when two checks generate the same id — and `analysis_evidence`/`analysis_policy`/`analysis_assurance_requirement` were never part of that generated string. Two checks sharing target/profile/channel/depth but differing only in evidence method (this plan's own acceptance test) correctly don't group, but both generate the identical check_id and get rejected before either runs. Specified two alternative fixes (explicit-id requirement, or auto-folding the differing axes into the id) and flagged the existing guard's message as needing an update either way.
- **G44 cross-release module identity**: confirmed the pair transition table never defines how module identity is matched across releases — a filename-keyed match would misread a simple rename as pair deletion + unrelated pair creation, wrongly gating an unchanged relationship as a new break. Required a stable, content-based identity match before the table applies, modeled on this codebase's existing `binary_fingerprint.py` rename-detection precedent, plus a fifth acceptance-test variant.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Twenty-second round, verified and pushed in f7ec722:

- **G42 downstream re-scoring**: confirmed `checker.compare()` deliberately runs `_env_matrix_contract_changes()` before `_apply_soname_policy()`/`_compute_verdict_for()`, with its own comment explaining why — a floor-decided finding drives the SONAME advisory, and the internal-node demotion reads what this stage already stamped. The previous round's fix reran only `_env_matrix_contract_changes()` per environment while leaving these downstream stages shared from one pass, so only one environment's outcome would get correctly attributed. Fixed by requiring the complete downstream chain (soname policy, verdict computation, contract decision recording) to rerun per environment, with only the genuinely upstream stages staying shared.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Twenty-third round, both verified and pushed in 54c869b:

- **G45 discriminator persistence**: confirmed the header-only target-kind discriminator was scoped only to the pre-dump `build-output.json` schema, with nothing persisted on `AbiSnapshot`/the baseline manifest — so a serialized-and-reloaded header-only baseline is indistinguishable from a failed binary-evidence capture. Added the discriminator as a real `AbiSnapshot` field persisted through the baseline storage envelope.
- **G44 rename-matching key**: confirmed the previous round's fix — using the module's own resolved internals/registration identity as the matching key — is itself flawed: that identity is the compatibility axis being *compared*, not an independent module identity, so it can both collapse distinct compatible siblings and fail to match a module against its own prior self when its internals genuinely changed (exactly the case this detector exists to catch). Replaced with an honest "not yet solved" acknowledgment naming real candidate signals rather than asserting one.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Twenty-fourth round, verified and pushed in e53f587:

- **G38 comparison outcomes out of BundleFacts**: confirmed the previous round's parenthetical — that `BundleFacts` might need extending to carry per-member comparison outcomes — is architecturally wrong. `BundleFacts` is a reusable, one-release facts artifact by its own docstring; a `DiffResult`/assurance result is inherently specific to one old/new pairing plus candidate and policy, which is exactly why `compare_bundle_sides()` already receives `per_library_results` separately rather than reading it off `BundleFacts`. Corrected: member snapshots publish as `BundleFacts` (facts only), per-member reports/results transport to the dispatch job as their own separate artifacts.

`check_docs_contract.py`/`check_ai_readiness.py` both clean.

---

napetrov (commented):
Addressed round 25's review findings (commit 1a83a15):

1. **G38 (`g38-bundle-facts-model-and-multibuild-comparability.md`)**: the Phase 15 design's "transport the ordinary per-target reports" claim didn't hold up — verified `DiffResult` has neither `to_dict()` nor `from_dict()`, and `compare_bundle_from_facts()` only ever receives a live, in-process `list[DiffResult]` today, with no loader that could rehydrate one from a transported report. Replaced with two concrete alternatives for this phase to choose between: recompute each member's `DiffResult` fresh at dispatch time from the assembled snapshot pair, or add a real serializer to `DiffResult` itself as its own scoped follow-up.

2. **G44 (`g44-binding-abi-provider.md`)**: the per-release module-pair compatibility graph was keyed by module pair alone, which silently loses a regression when a pair shares more than one candidate native type (a fresh incompatibility on type B is masked if the pair already disagrees on type A). Reworked to key and diff by `(module pair, shared native type)` instead, with a sixth acceptance-test variant covering exactly this scenario.

Both verified against source before fixing. `scripts/check_docs_contract.py` and `scripts/check_ai_readiness.py` both clean (0 errors; only pre-existing, unrelated warnings).

---

napetrov (commented):
Addressed round 26's review findings (commit 3fbfa7f), both on `g42-check-identity-environments-and-provider-resolution.md`:

1. **Preserve pre-environment verdict modulations**: the design's "reset `effective_verdict`/`modulation_reason`/`modulation_rule` to `None`" alternative was wrong — verified `diff_stdlib_impl.py` constructs some findings with `effective_verdict=BREAKING` already set, and `diff_templates.py`'s `demote_lambda_closure_unexported_findings()` explicitly skips any already-modulated change before demoting others, relying on `apply_runtime_floor_contract()`'s identical skip check to leave those authoritative decisions alone. Blanket-resetting to `None` erases that signal. Fixed by requiring a snapshot of the true pristine pre-environment state instead of a blanket reset.

2. **Isolate contract stages across environment branches**: verified `ContractEvaluationStage.classify()` appends to `self.changes` rather than replacing it, and `build_context()` persists that same accumulated list — so reusing one shared stage instance across environment branches would contaminate every branch after the first with earlier branches' findings. Fixed by requiring a fresh `ContractEvaluationStage` (empty `self.changes`) per environment branch, while keeping the expensive, environment-independent evidence collection (`build_contract_stage()`) shared.

Both verified against source (`diff_stdlib_impl.py`, `diff_templates.py`, `diff_versioning.py`, `contract_pipeline.py`) before fixing. `scripts/check_docs_contract.py` and `scripts/check_ai_readiness.py` both clean (0 errors; only pre-existing, unrelated warnings).

---

PR #920, opened 2026-08-28
docs(pr-c): dump CLI's L4 extractor default silently diverges from compare/typed pipeline

napetrov:
## Summary

Part 2 of 3 in finalizing PR C (typed `dump`/`scan` convergence) of `docs/contribute/plans/cli-cleanup-phase-two.md`. Investigates item 2 (migrating `dump_cmd`'s real execution onto `execute_dump_request`) directly, now that real castxml is available in this session's environment — one of the two previously-cited blockers.

## What I found

Confirmed the flat (L0-L2) snapshot payload produced by the real `dump` CLI and by `resolve_dump_request`/`execute_dump_request` already agree byte-for-byte. But I found a **new, previously-undocumented divergence**: `dump`'s own CLI forwards `--ast-frontend`'s raw, unresolved `"auto"` string straight through to `buildsource.inline._make_source_extractor` (which treats anything but literally `"castxml"` as clang), while `compare`'s implicit-dump operand and the typed pipeline both resolve `"auto"` through `service_compare_evidence.effective_frontend` to castxml. So a plain `dump --depth source` with no explicit `--ast-frontend` silently resolves its L4 source-ABI replay to **clang** — disagreeing with every other resolver.

Traced and reproduced directly (not guessed): a debug spy on the real CLI call confirmed `header_backend="auto"` reaches `perform_elf_dump` unresolved, and a `fact_set.producer` comparison confirmed `"abicheck-cc-clang-extractor"` (CLI) vs `"castxml-source"` (typed pipeline) for the identical input.

This does **not** trip the `profile_fingerprint`/`scope_fingerprint` comparability gate — it silently produces different L4 facts rather than a loud `NOT_COMPARABLE`, live only when the two extractors actually disagree about a declaration (which is exactly what item 3's own castxml phantom-implicit-member bug does).

## Why I didn't fix it

Both "obvious" fixes are each their own real behavior change: matching `dump`'s CLI to castxml would import item 3's still-open bug into every plain `dump` invocation; matching `compare`/typed to clang would contradict this codebase's established castxml-canonical-default architecture. This changes item 2's own risk calculus — "byte-identical to today's output" isn't the right acceptance bar for L4 facts specifically, since today's `dump`-CLI output is itself one side of a pre-existing disagreement, not a stable target to match. A correct migration plan needs to resolve this divergence first, which is its own dedicated, verified decision — consistent with this exact code area's established "known gaps over risky reactive patches" convention (this plan doc alone documents dozens of prior instances of exactly this judgment call).

The actual code migration (`perform_elf_dump`/`handle_non_elf_dump` onto `execute_dump_request`) was not attempted in this pass.

## Changes

- Docs-only: adds this finding to the plan doc's item 2 section, with the full reproduction and reasoning.

## Verification

- `python scripts/check_docs_contract.py` — 0 errors, 2 pre-existing warnings (unchanged).
- `python scripts/check_ai_readiness.py` — 0 errors, warning count unchanged.
- No code touched.

---

https://claude.ai/code/session_01L93nkWGspPxR5vNjJpihML

---
_Generated by [Claude Code](https://claude.ai/code/session_01L93nkWGspPxR5vNjJpihML)_

## Summary by CodeRabbit

* **Documentation**
  * Added a dated note documenting differences in default source-ABI extraction behavior between CLI and typed dump workflows.
  * Clarified how environment settings and configuration affect frontend selection.
  * Documented compatibility handling for extractor differences and identified considerations for future CLI migration.

### DIFF
--- a/docs/contribute/plans/cli-cleanup-phase-two.md
+++ b/docs/contribute/plans/cli-cleanup-phase-two.md
@@ -2462,6 +2462,91 @@ pipelines a fourth time.
   > characterized instead of merely "unverifiable" or "needs its own
   > dedicated pass" in the abstract.
 
+  > **A new, previously-undocumented divergence found while investigating
+  > item 2's migration risk directly (2026-08-28): `dump`'s own CLI L4
+  > extractor default disagrees with `compare`'s implicit-dump operand and
+  > the typed `execute_dump_request` pipeline — a real bug, unrelated to
+  > whether the migration happens, and it changes the migration's risk
+  > profile rather than only its verification burden.** Traced end to end,
+  > not guessed: `perform_elf_dump` (`cli_dump_helpers.py`) receives
+  > `header_backend` from the CLI's own resolved value and forwards it
+  > **unresolved** — `extractor=header_backend`, still the literal string
+  > `"auto"` when `--ast-frontend` is not given — all the way down through
+  > `_write_snapshot_output` → `embed_build_source` → `collect_inline_pack`
+  > to `buildsource.inline._make_source_extractor`, which (per its own,
+  > already-documented behavior) treats anything but the literal string
+  > `"castxml"` as clang. So a plain `dump --depth source` with no explicit
+  > `--ast-frontend` resolves its L4 source-ABI replay to **clang**.
+  > Confirmed with a debug spy on the real CLI invocation (`header_backend`
+  > captured as the literal `"auto"` at the `perform_elf_dump` call site).
+  >
+  > `compare`'s implicit-dump operand and `execute_dump_request` both reach
+  > L4 replay through the *same* shared primitive,
+  > `workflows.artifact.execute.embed_side_build_source`, whose
+  > `source_extractor: str | None = None` parameter's own docstring already
+  > states the contract: `None` keeps that function's own
+  > `service_compare_evidence.effective_frontend` resolution, which resolves
+  > `"auto"` to **castxml** unconditionally (the identical resolution this
+  > plan's item 2 notes above already established for the L2 header-AST
+  > pass). Neither caller passes an override, so both correctly resolve to
+  > castxml. Confirmed directly: a `dump`-CLI-written baseline's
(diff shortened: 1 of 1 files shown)


napetrov (commented):
All three fixed in 1131da6, each confirmed by reading the code directly:

- **Qualify auto resolution by the environment override**: confirmed — `_make_source_extractor` never consults `ABICHECK_AST_FRONTEND`, but `effective_frontend`/`_resolve_header_backend` does, so with it set to `clang` both sides land on clang and the divergence vanishes. Qualified the claim.
- **Distinguish un-gated metadata from unrecorded metadata**: confirmed, real self-contradiction (the paragraph itself already showed `producer` recording the choice two sentences earlier). Corrected: the gap is that the recorded identity isn't folded into the comparability gate, not that it's unrecorded.
- **Reconcile the claimed canonical L4 default**: confirmed — `buildsource.source_extractors.resolver.AUTO_PREFERENCE = (CLANG, CASTXML)` is real and clang-first, though it has no production caller today (only its own test module calls it). Narrowed the claim to what actually constrains the direction: the typed pipeline's own deliberate L4-to-L2 coupling, not a codebase-wide convention.

Also updated the note's premise now that item 3 (the castxml phantom-implicit-member bug this note cited) is fixed as of PR #930, landed in the same session.

---

napetrov (commented):
Fixed in 101bcec — confirmed by reading `diff_source_abi()`/`check_fact_compatibility()` directly: the latter explicitly handles a `producer`/`producer_version`/`compiler_version` mismatch as one of its own named invalidating conditions, selectively suppressing only the categories a producer mismatch actually invalidates while keeping compiler-neutral facts comparable. Folding `producer` into `profile_fingerprint`/`scope_fingerprint` would replace that deliberate, already-built mechanism with a blunt all-or-nothing `NOT_COMPARABLE`. Corrected the note to say so, and to state that any real fix needs to reconcile with this existing mechanism rather than bypass it.

---

Issue #838, opened 2026-08-23
dump --sources TREE -H hdr silently ignores -H; no headers-only dump exists


napetrov:
## Summary

`abicheck dump` refuses to run without a binary, and if you satisfy it with `--sources` (the documented alternative) the `-H` headers are dropped without a word: the snapshot has 0 functions / 0 enums and depth `binary`, and the command exits 0.

This forces a "probe .so" workaround for header-only libraries and for projects whose CI cannot afford a full product build per PR.

## Root cause

A source-only `dump` (no `SO_PATH`) dispatches to `cli_buildsource.dump_source_only()`, whose own docstring says it "writes a binary-less snapshot carrying only the embedded build/source facts" — it embeds L3/L4/L5 facts only, has no L2 header-AST pass, and (confirmed by reading the dispatch call in `cli.py`) never even *receives* `headers` as an argument. There is no code path where `-H` could have any effect here.

## Reproducer (executed, confirmed)

```bash
rm -rf /tmp/d3 && mkdir -p /tmp/d3/srctree && cd /tmp/d3
cat > e.hpp <<'EOF'
typedef enum { a = 0, b, c } my_tag_t;
EOF
cp e.hpp srctree/
abicheck dump -H e.hpp -I . --version t                                       # (A)
abicheck dump --sources srctree -H e.hpp -I . --version t -o src_snap.json    # (B)
python -c "import json;d=json.load(open('src_snap.json'));print(len(d.get('enums',[])),'enums')"
```

**Observed**
- (A) `Error: dump requires a binary (SO_PATH), or --sources/--build-info for a source-only snapshot.` — headers alone cannot be dumped at all.
- (B) exit 0, `Resolved evidence depth: binary`, `0 enums`, ~1.8 KB snapshot. The only warning concerns a missing `compile_commands.json`; nothing says `-H` was ignored.

**Expected**: at minimum (B) must not silently discard `-H` — either parse the headers, or fail with a message naming the ignored flag.

## Where

`abicheck/cli.py`'s `dump_cmd` — the `so_path is None` dispatch to `cli_buildsource.dump_source_only()` (which never receives `headers`) — and `abicheck/cli_buildsource.py:dump_source_only`'s own "no artifact and no --sources/--build-info" rejection, which has no equivalent check for "headers given but this path can't use them."

## Suggested fix

[...]

Issue #834, opened 2026-08-23
Public-surface scoping classifies every transitively-included header as private-header, silently hiding real breaking findings


napetrov:
## Summary

With public-surface scoping on (the default), a declaration is `public_header` only if its own header is one of the `-H` roots (or an explicitly passed `--public-header-dir`). A header reached *transitively* from the `-H` umbrella header — even one sitting inside the single `-I` include root handed to the same dump — is classified `private-header`. If such a type is also not directly referenced in an exported function's signature, its findings move to the out-of-surface ledger and the verdict silently becomes `NO_CHANGE`/exit 0.

Not hypothetical: a canary — one added `int` member to `oneapi::dpl::discard_iterator` in `include/oneapi/dpl/pstl/iterator_impl.h`, reached via the public `<oneapi/dpl/iterator>` — produced, on the *same snapshot pair*:

```
default scoping          : verdict NO_CHANGE, 0 changes, exit 0
--no-scope-public-headers: verdict BREAKING, exit 4
```

`surface_scope.out_of_surface_changes` states the reason verbatim: `"reason": "private-header"`. A CI gate watching the verdict alone sees nothing at all.

## Root cause

A header-AST dump (`dump -H api.h -I include`) only ever parses declarations reachable by `#include` from its own `-H` root(s) — there is no other way for a declaration to end up in the snapshot at all. So *every* header that produced a declaration in the snapshot was, by construction, reached transitively from a `-H` root. But `provenance.classify_origin`/`_matches_public` only ever matches the literal `-H` files/dirs (or an explicit `--public-header-dir`) — the dump's own `-I` roots never become "public directories" for classification purposes — so a header one `#include` deeper than the umbrella header classifies `PRIVATE_HEADER` even though it is not, in any meaningful sense, private.

## Minimal reproducer (executed, confirmed — using the clang frontend since castxml wasn't available in this sandbox; the classification bug is frontend-independent)

```bash
rm -rf /tmp/d4 && mkdir -p /tmp/d4/old/detail /tmp/d4/new/detail && cd /tmp/d4
for s in old new; do
  cat > $s/detail/impl.hpp <<EOF
#pragma once
namespace demo {
class Widget {
public:
    int id() const { return id_; }
private:
    int id_ = 0;
$( [ $s = new ] && echo "    int canary_ = 0;   // deliberate layout break" )
};
}
EOF
  cat > $s/lib.hpp <<'EOF'
#pragma once
#include "detail/impl.hpp"
[...]

PR #925, opened 2026-08-28
refactor(architecture): classify 19 more flat modules (ADR-061)

napetrov:
## Summary

Classifies 14 of my assigned batch of 29 flat `abicheck/*.py` modules into
their ADR-061 responsibility layers (`architecture/modules.yaml`
`legacy_paths`). Classification only — no files moved, no behavior change.
15 modules are deliberately left unclassified; see "Skipped" below.

**Updated after Codex review**: 5 modules I originally classified as
`model` (`config_paths.py`, `contract_coverage_ledger.py`,
`contract_gating.py`, `deadline.py`, `demangle.py`) were correctly flagged
as real responsibility mismatches — each was only forced into `model` by
the checker's import-direction mechanics (their bare "which layer can
every importer legally reach" intersection reduces to `{model}`), not
because `model` is where any of them actually belong. None of the forcing
imports are typing-only, so no `TYPE_CHECKING` fix applies per the
original task's fallback rule — all five are now left **unclassified**,
moved from the "Changes" table into "Skipped" below with the real
conflicting importers named.

## Motivation

Part of the incremental ADR-061 flat-module classification effort
(parallel batches; this is batch 2). `python scripts/check_architecture.py`
reports `Architecture: 0 error(s)` after this change.

## Changes

Classified into `model` (imports nothing outside model, and matches its
real responsibility as vocabulary/shape or a detector-registration leaf,
not merely satisfying the checker):

| Module | Layer | Why |
|---|---|---|
| `contract_relevance_types.py` | model | Forced: `checker_types.py` (model) imports it at runtime (`model` same-layer is the only option satisfying that importer). Zero internal imports — a real fit, not just a technically-valid one (ADR-049 reserved vocabulary, sibling of the already-model-classified `checker_types.py`). |
| `contract_evidence.py` | model | Sibling shape/vocabulary module to `contract_relevance_types.py` (ADR-049 Phase 4 "block" dataclasses, explicitly documented as pure shape with nothing wired to detection/policy logic yet); only imports model-classified modules. |
| `detectors.py` | model | Forced: imported by `checker_types.py` (model) and `confidence.py` (compare) — `model` is the only common candidate. Zero internal imports; itself only individual detection *rule* data, not policy logic. |

Classified into `policy` ("decide relevance/classification/gating"):

| Module | Layer | Why |
|---|---|---|
[...]

PR #932, opened 2026-08-28
test: registry-completeness and evidence-completeness invariants (Phase 9)

napetrov:
## Summary

Implements Phase 9 ("registry-completeness and silent-degradation invariants") of `docs/contribute/plans/bug-class-regression-testing.md` — as a single, bounded slice rather than the plan's full XL-effort scope.

## Motivation

Phase 9 generalizes two incidents already in `tests/regressions/manifest.py`:

- `registry.kind_completeness` (#753 → #759): a hand-maintained, kind-keyed list can silently go stale after a `ChangeKind` rename/removal, with nothing failing to say so.
- `evidence.silent_degradation_to_clean_verdict` (incident #883 among others): missing/dropped evidence handling must be checked independently, "not all derived from one production helper" — and the *existing* regression tests for #883 violate exactly that principle (they mock the very function whose real effect needs proving).

## Changes

- `tests/test_report_classifications_unit.py`: `report_classifications.py`'s seven hand-maintained kind-keyed frozensets (`REMOVED_KINDS`, `ADDED_KINDS`, `BINARY_ONLY_KINDS`, `ENVIRONMENT_DRIFT_KINDS`, `CHANGED_BREAKING_KINDS`, `HIGH_SEVERITY_KINDS`, `MEDIUM_SEVERITY_KINDS`) are now reverse-completeness checked (every member must name a real, live `ChangeKind`), with a mutation check proving the assertion actually fails against a deliberately corrupted set rather than passing vacuously. Also pins `HIGH_SEVERITY_KINDS`/`MEDIUM_SEVERITY_KINDS` disjointness, since `severity()`'s if/elif chain would silently resolve an overlap to `"High"`.
- `tests/test_bundle_side_input.py`: adds `test_policy_file_override_genuinely_demotes_a_real_verdict`, which runs the real, unmocked `service.compare_snapshots` (only `service.resolve_input` is mocked, for the NEW-side I/O boundary) and asserts a real `PolicyFile` override genuinely demotes a real `FUNC_VISIBILITY_CHANGED` finding's verdict — both per-library and in the aggregate `BundleDiffResult.verdict` — contrasted against the unpatched `BREAKING` baseline. Verified locally that reverting `bundle_side_input.py`'s `policy_file` forwarding (the #883 fix) makes this new test fail.
- `tests/regressions/manifest.py`: records the new seed tests on both bug classes and narrows (never deletes) their `known_gaps` to state what's still open.

Deliberately **out of scope** for this PR, per the plan's own listed incidents and this codebase's "known gaps over risky reactive patches" convention:
- Incidents #834/#860 — would need heavy CastXML/header-AST fixtures disproportionate to one bounded PR.
- Incident #838 — already guarded by its own targeted CLI warning.
- Forward-completeness for `REMOVED_KINDS`/`ADDED_KINDS` — both are small, deliberately curated subsets (6/56 and 8/40 of matching-suffix kinds today) for one report section, not an attempt to enumerate every kind sharing that suffix; asserting forward-exhaustiveness there would invent a business rule this module's own authors never stated.

## Checklist

- [ ] Changelog fragment added (`scriv create`) — not needed, diff touches only `tests/`
- [ ] Docs updated if user-visible behaviour changed — no user-visible behaviour changed
- [x] `pre-commit run --all-files` passes locally (`ruff check`, `mypy abicheck/`, `python scripts/check_architecture.py --base origin/main` all clean/pre-existing-only; `ruff format` flags pre-existing, unrelated formatting drift in both touched files per the repo's own documented known gap)

https://claude.ai/code/session_018USuGdPgyVaPK7z6iB2cKd

---
_Generated by [Claude Code](https://claude.ai/code/session_018USuGdPgyVaPK7z6iB2cKd)_

### DIFF
--- a/tests/regressions/manifest.py
+++ b/tests/regressions/manifest.py
@@ -516,7 +516,7 @@ BUG_CLASSES: tuple[BugClass, ...] = (
             "every kind has a mapping, and every mapping key names an "
             "existing kind."
         ),
-        fixed_by=(753, 759),
+        fixed_by=(753, 759, 932),
         seed_tests=(
             "tests/test_canonical_finding_id_completeness.py",
             "tests/test_report_classifications_unit.py",
@@ -558,7 +558,7 @@ BUG_CLASSES: tuple[BugClass, ...] = (
             "result must agree, checked independently, not all derived "
             "from one production helper."
         ),
-        fixed_by=(834, 838, 860, 883),
+        fixed_by=(834, 838, 860, 883, 932),
         seed_tests=(
             "tests/test_fact_conservation_properties.py",
             "tests/test_bundle_side_input.py",
--- a/tests/regressions/manifest.py
+++ b/tests/regressions/manifest.py
@@ -517,24 +517,32 @@ BUG_CLASSES: tuple[BugClass, ...] = (
             "existing kind."
         ),
         fixed_by=(753, 759),
-        seed_tests=("tests/test_canonical_finding_id_completeness.py",),
+        seed_tests=(
+            "tests/test_canonical_finding_id_completeness.py",
+            "tests/test_report_classifications_unit.py",
+        ),
         known_gaps=(
             KnownGap(
                 description=(
-                    "Only canonical_finding_id's classification is "
-                    "exhaustiveness-checked today; the same totality "
-                    "property is not yet generalized to every other "
-                    "kind-keyed registry named in Phase 9 of the plan "
-                    "(evidence kinds, providers, report renderers)."
+                    "canonical_finding_id's classification and "
+                    "report_classifications.py's seven hand-maintained "
+                    "kind-keyed frozensets are now reverse-completeness "
+                    "checked (every member names a live ChangeKind, with a "
+                    "mutation check proving the assertion actually fails on "
+                    "a corrupted set) — but the same totality property is "
+                    "not yet generalized to every other kind-keyed registry "
+                    "named in Phase 9 of the plan (evidence kinds, "
+                    "providers, other report renderers)."
                 ),
                 reference="docs/contribute/plans/bug-class-regression-testing.md#phase-9",
             ),
             KnownGap(
                 description=(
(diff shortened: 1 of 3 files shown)


napetrov:
The `ai-readiness` check (its `architecture` step) is failing on this PR's head:

```
ERROR: [dependency-direction] abicheck/contract_evidence.py:86: model -> policy is forbidden; allowed: (none)
ERROR: [dependency-direction] abicheck/contract_evidence.py:86: model -> policy is forbidden; allowed: (none)
ERROR: [dependency-cycle] observed responsibility cycle: compare -> model -> policy -> compare
Architecture: 3 error(s)
```

This is **not caused by this PR** — the diff here only touches `tests/` (no `abicheck/**/*.py`), and the identical three errors reproduce on `main`'s own current tip (verified locally by checking out `origin/main` at `8d462a9` and running `python scripts/check_architecture.py` directly — same file, same line, same error text). It also failed the same way on `main`'s own CI run for the just-merged PR #925 (run [33178647744](https://github.com/abicheck/abicheck/actions/runs/33178647744), job `ai-readiness`).

Root cause (for whoever picks this up): `abicheck/contract_evidence.py` is classified under `architecture/modules.yaml`'s `model` layer (`may_import: []`), but line 86 imports `CompatibilityEvaluationConfig` from `compatibility_evaluation_config.py`, which is classified under the `policy` layer. This looks like a classification gap left by the recent ADR-061 classification batch PRs (#924/#925) rather than a new import — `contract_evidence.py` either needs reclassifying out of `model`, or the `CompatibilityEvaluationConfig` type it depends on needs to move into `model` alongside it.

I don't have an existing fix to port for this, and reclassifying either module is an ADR-061 architectural call outside this PR's scope (registry/evidence-completeness test coverage) — not a re-run candidate either, since it reproduces deterministically off `main`'s own tip rather than being a flake. Flagging it here so it's visible on this PR's checks; happy to pick it up separately if that's preferred.

---

PR #974, opened 2026-09-01
refactor(architecture): move namespace-move batch-rename detection into compare/ (ADR-061)

napetrov:
## Summary

Fixes 2 of the 6 `architecture/debt.yaml` `debt-no-growth` violations that `scripts/check_architecture.py` currently reports on `main`, as a genuine ADR-061 responsibility move rather than a line-count trim.

## Motivation

`python scripts/check_architecture.py` currently fails on `main` (before this PR) with 6 `debt-no-growth` errors — files that grew past their recorded ADR-061 no-growth baseline. Two of them (`abicheck/diff_symbols_renames.py`, `tests/test_batch_rename_namespace_move.py`) were fixable by finishing the responsibility move `debt.yaml` itself already targets, rather than needing a design decision. The other four (`cli_buildsource_helpers.py`, `serialization.py`, `test_mutation_score_gate.py`, `test_service_unit.py`) are smaller/differently-shaped and left for a follow-up.

## Changes

- **`abicheck/compare/namespace_move.py` (new)**: `find_namespace_move_groups`/`emit_namespace_move_batches` and their private helpers (`_scope_components`, `_qualified_key_scope_components`, `_declaring_entity`) — the whole namespace-move-substitution subsystem, used nowhere else in `diff_symbols_renames.py` — moved here, matching that file's own ADR-061 `compare` classification and the sibling `compare/rename_ambiguity.py` split. `diff_symbols_renames.py` (now 807 lines, was 1284, baseline 1187) re-exports the two entry points so `diff_symbols.py`'s existing public-surface re-export block is untouched.
- **`abicheck/model/synthetic_key.py` (new)**: the synthetic ctor/dtor snapshot-key vocabulary (`SYNTHETIC_CTOR_KEY_PREFIX`, `is_synthetic_ctor_key`, `is_synthetic_dtor_key`) that `namespace_move.py` needs was previously reachable only through `dumper_castxml.py` (`extract`), which `compare` may not import per ADR-061 D1 (`architecture/modules.yaml`). Relocated to `model` — the same "shared vocabulary both `extract` and `compare` need" pattern `model/mangled_name.py` already established — with `extract/headers/castxml/names.py` and `dumper_castxml.py` re-exporting it so every existing import path (including `finding_identity_ctor_dtor.py`) keeps working unchanged.
- **Test split**: `tests/test_batch_rename_namespace_move.py` (1788 lines, baseline 1460) is split by contract axis into `tests/unit/compare/test_namespace_move.py` (namespace-move grouping/emission + the small prefix-rename/destructor-noise regression it shared the file with) and `tests/unit/compare/test_scope_component_parsing.py` (`qualified_name_scope_components`/`strip_trailing_top_level_parameter_list` edge cases, which exercise `diff_cxx_rules` directly) — matching `debt.yaml`'s own recorded target for this entry ("tests/unit/compare alongside namespace/rename comparison migration"). Both new files land under the 1200-line test hard cap, so the `debt.yaml` entry is retired rather than retargeted.
- **`architecture/debt.yaml`**: removed the retired test entry; lowered `diff_symbols_renames.py`'s baseline to its new size.

No behavior change — this is a pure internal move.

## Verification

- `python scripts/check_architecture.py`: 6 errors → 4 (the two fixed here are gone; the remaining four are pre-existing and unrelated).
- `mypy abicheck/`: clean (554 files, 0 errors) — also caught and fixed two real relative-import bugs introduced during the move (`.checker_policy`/`.diff_cxx_rules`/etc. needed to be `..` from inside the new `abicheck/compare/` package) and one real dependency-direction violation (`compare` importing `extract`-classified `dumper_castxml`, fixed by the `model/synthetic_key.py` relocation) — both would otherwise have been silent `ModuleNotFoundError`s / architecture violations.
- `ruff check` / `ruff format --check`: clean on every touched file.
- `pytest`: 736 tests covering every touched code path (the new `tests/unit/compare/` files, `test_namespace_move_cross_position_ambiguity.py`, `test_rename_ambiguity_properties.py`, and the broader `rename`/`namespace_move`/`diff_symbols`/`synthetic`/`scope_component` keyword slice) — all pass, unchanged.

This PR touches only internal module organization with no user-facing behavior change, so no changelog fragment is included (see `changelog.d/README.md`'s "internal refactor" carve-out) — requesting the `skip-changelog` label.

https://claude.ai/code/session_01JJ6j4xpSSWwZFsyZDp3UzW

---
_Generated by [Claude Code](https://claude.ai/code/session_01JJ6j4xpSSWwZFsyZDp3UzW)_

## Summary by CodeRabbit

- **New Features**
  - Detects and reports grouped C++ namespace moves and symbol renames with improved accuracy.
  - Adds shared handling for synthetic constructor and destructor symbols.

- **Bug Fixes**
  - Failed external extraction no longer leaves stale normalized outputs; raw artifacts are preserved.
  - Malformed snapshots now produce clear, classified errors instead of uncaught crashes.
  - Invalid build metadata is rejected rather than silently ignored.

- **Maintenance**
  - Policy override handling and CLI internals were reorganized without changing expected behavior.


napetrov (commented):
Pushed b58253d addressing all 4 open P1 findings:

- `cli_buildsource.py`: restored the `_purge_external_outputs` back-compat alias.
- `storage/snapshot_load_normalization.py`: `extraction_contract_from_dict` now drops a malformed `profile_fields`/`scope_fields` key/value pair instead of coercing it with `str()`, per storage `AGENTS.md` invariant 6. Added `tests/unit/storage/test_snapshot_load_normalization.py` (12 cases, 6 confirmed to fail pre-fix via `git stash`).
- `frontends/cli/options/params.py`: `builtin_policy_names` discovery now routes through `workflows.policy_file` (which re-exports it alongside `builtin_policy_path`) instead of importing `abicheck.policies` directly.
- `frontends/AGENTS.md`: updated the stale note that `cli_params.py` was still flat — it documents the completed move to `cli/options/params.py` now.

`mypy`/`ruff`/`check_architecture.py` all clean; targeted tests pass.

---

napetrov (commented):
Pushed ec116e7. `extraction_contract_from_dict` now reuses `storage/guards.py`'s `mapping`/`decision_key`/`identity_text` primitives to raise `TypeError` for a *present* but wrong-shaped `profile_fields`/`scope_fields` (container-level or per-pair) instead of filtering out the bad entries — matching the package's existing `dependency_scope` precedent of failing a malformed load loudly. Only a genuinely absent field (missing key, or explicit `null`) still degrades to `{}`. Updated the round-trip test that pinned the old degrade behavior, and rewrote the dedicated unit test class to assert rejection (`tests/unit/storage/test_snapshot_load_normalization.py`). `mypy`/`ruff`/`check_architecture.py` clean; 488 tests across every consumer of this code path pass.

---

napetrov (commented):
Pushed dc032c2. Moved `abicheck/policies/__init__.py` from `model`'s `legacy_paths` to `policy`'s in `architecture/modules.yaml`, preserving the `frontends -> workflows -> policy` route — both current importers (`policy_file.py`, `workflows/policy_file.py`) already sit in layers permitted to import `policy`. Also fixed the now-stale `frontends/AGENTS.md` reference that still called it `model`-classified. `check_architecture.py`/`mypy`/`ruff` clean.

---

napetrov (commented):
Pushed c694f09. `_purge_external_outputs` now resolves through the same lazy `__getattr__` (PEP 562) shim this module already uses for the `cli_graph` re-exports, via `importlib.import_module("abicheck.cli_buildsource_helpers").purge_external_outputs` — so the compatibility path always reflects whatever the real owner currently is (including a test's `monkeypatch.setattr` on it), not a snapshot frozen at this module's own import time. `mypy`/`ruff`/`check_architecture.py` clean; `tests/test_cli_moved_surface.py` and the broader `cli_buildsource`/`purge` slice (89 tests) pass.

---

napetrov (commented):
Pushed 97c67f3, extending this PR's ADR-061 scope: finalizes Phase 4's own recorded gap ("`DiffResult._effective_kind_sets`/`_effective_verdict_for_change` still run real policy-resolution logic as model-owned methods — a recorded, unaudited gap").

**Analysis**: `_effective_verdict_for_change` was already a pure delegating call to `reclassify.effective_verdict_for_change`. `_effective_kind_sets` was not — after calling `checker_policy.policy_kind_sets()`, it locally implemented `PolicyFile.overrides` application (build a verdict→set-index table, discard/re-add per override) entirely inline inside `checker_types.py`'s own method body, real algorithmic logic living in a `model`-classified file.

**Fix**: extracted that loop verbatim into a new `checker_policy.apply_policy_file_overrides()` function; `_effective_kind_sets` is now a single delegating call. Deliberately does *not* remove the `model -> checker_policy`/`reclassify` import edges (both methods still have to call something, and ~40 call sites across `report/`/`policy/`/`workflows`-owned `cli_*` modules already call `result._effective_kind_sets()`/`result._effective_verdict_for_change(...)` directly) and does *not* reclassify either module as `policy` (`model`'s own `may_import` is `[]` by design — widening it would legalize any `model` module reaching into policy, not just this one audited edge). Both stay `public_root_surfaces` leaves, the same sanctioned shape `reclassify.py`/`contract_gating.py` already have.

`self.policy`/`self.policy_file` mutated after construction (a pattern several existing tests rely on) still take effect on the next call — both methods still read fresh each time, no caching introduced.

Updated ADR-061's own Phase 4 section and the ADR index row to record the closure.

**Verification**: `mypy`/`ruff`/`check_architecture.py` clean; targeted slice covering every consumer of the two methods (`test_policy_file.py`, `test_report_integrity.py`, `test_reporter.py`, `test_sarif.py`, `test_sprint9_html.py`, `test_junit_report.py`, `test_reclassify.py`, `test_config_review.py`, `test_annotations.py`, `test_frozen_namespace.py`, `test_report_filtering.py`, `test_vtable_severity.py`, `test_bundle.py`, `test_report_model.py`, `test_checker.py` — 1300+ tests) all pass unchanged; a full fast-suite run is in progress.

---

napetrov (commented):
Pushed 2370264, addressing the P2 finding: `workflows/input_resolution.py`'s JSON snapshot loader and `compat/cli.py`'s `_parse_compat_descriptors` both now catch `TypeError` alongside `ValueError`/`KeyError`/`UnicodeDecodeError`/`OSError`, wrapping it into the same clean `SnapshotError`/classified compat exit every other malformed-snapshot failure gets — matching `bundle_facts.py`'s own established precedent (it already catches `TypeError` around its `snapshot_from_dict()` call). Added a regression test at each boundary, verified to fail on the pre-fix code via `git stash`. Also fixed a `CLI_CONTRACT_ALLOWLIST` line-number entry in `check_ai_readiness.py` that this same file's edit shifted.

I see two more P1 findings just landed on `snapshot_load_normalization.py` (`build_mode_from_dict`'s container/`libcpp_abi_version` coercion) — picking those up next.

---

napetrov (commented):
Pushed 55fc1b8, addressing both `build_mode_from_dict` P1 findings:

- A present-but-non-mapping `build_mode` (or its nested `provenance`) now raises `TypeError` via `storage/guards.mapping` instead of silently returning `None` — indistinguishable before from a legacy snapshot that genuinely predates the field.
- `libcpp_abi_version` no longer coerces a numeric string (`"1"`) to `int` — raises `TypeError` via `storage/guards.strict_int` for anything but a real `int`; only a genuinely absent value (missing key/explicit `null`) still reads as `None`.

Updated the one existing test that pinned the old degrade-silently contract (`test_schema_compat.py`, renamed `..._rejects_the_load`) — its real point (a non-dict `provenance` must not raise an opaque `AttributeError`, the original CodeRabbit fix) still holds via the new typed `TypeError`. Added a dedicated 13-case unit test class (`TestBuildModeFieldsNeverCoerce`) — `build_mode_from_dict` had zero direct dict-parsing test coverage before this; 9 of the 13 cases confirmed to fail pre-fix via `git stash`.

`mypy`/`ruff`/`check_architecture.py` clean; 439 tests across every consumer pass.

---

napetrov (commented):
Pushed 99b5d89 + c452334.

**99b5d89** addresses this P2 finding: `frontends/cli/commands/compare_bundle_facts.py`'s `except (SnapshotError, ValueError, OSError)` around `compare_release_against_bundle_facts` now also catches `TypeError`, matching every other malformed-`OLD_FACTS`-shape case that boundary already handles. Regression test verified to fail on the pre-fix code — but only once I strengthened its assertions past `(exit_code, output)` alone: `CliRunner` reports `exit_code=1` with *empty* output for any uncaught exception, indistinguishable at that level from a deliberate `click.ClickException`. `result.exception`'s own type is what actually distinguishes them, so the test asserts on that directly.

**c452334** is unrelated to Codex's findings — while running the full AI-readiness gate to verify these fixes I found 3 `adr-status-sync` errors (stale `Verified:` receipts on ADR-037/044/046 naming an unreachable commit) already present on `main` itself (confirmed via a worktree check against `origin/main`). Ported the fix into this PR per the drive-to-green convention rather than waiting on it separately.

---

napetrov (commented):
Pushed f237bd4. `purge_external_outputs()` now returns `bool` (`True` only when every declared output and the `normalized/<name>/` subtree were confirmed absent afterward — an already-missing file still counts as removed, only a real `OSError` counts as failure). A new `_purge_and_record()` wraps all three call sites in `cli_buildsource_helpers.py`: on a purge failure it forces `record.status = "failed"` regardless of the prior status and appends a diagnostic to both the run-wide ledger and the extractor's own row.

That status change isn't cosmetic — it plugs directly into the pre-existing `_enforce_strict_mode` D9 gate (already covered by `tests/test_build_source_extractor.py`'s `test_cli_strict_mode_fails_on_*` tests), so a purge failure now aborts a `--collection-mode strict` run the same way any other extractor failure already does. Under the default permissive mode it's recorded and surfaced rather than silently ignored, matching this whole subsystem's existing permissive/strict split rather than special-casing purge as always-fatal.

Added `tests/test_pack_shape_purge.py` (7 cases), verified to fail on the pre-fix code via `git stash`. `mypy`/`ruff`/`check_architecture.py` clean.

---

abicheck/cli_buildsource_helpers.py
```diff
@@ -799,6 +800,29 @@ def _run_adapters(
...
+    ``status = "failed"`` alone does not stop a surviving file from being
+    hashed into the published pack identity, since permissive mode (the
+    default) lets collection continue past a failed extractor by design
+    (Codex review). A purge failure isn't a missing-evidence gap permissive
+    mode tolerates; it risks publishing corrupt evidence as genuine, so it
+    always raises, in every collection mode.
+    """
+    name = getattr(manifest, "name", "<unknown>")
+    if not purge_external_outputs(pack_root, manifest):
+        record.status = "failed"
```
napetrov, line 819: Confirmed — `record.status = "failed"` alone doesn't stop `pack_io.write()`'s `_artifact_digests()` from walking `normalized/` unconditionally, and the default permissive collection mode continues past a failed extractor by design, so the leftover file would still be hashed. Fixed in 46e4563 by having `_purge_and_record` raise `click.ClickException` unconditionally on a genuine purge failure, in every collection mode (not just strict) — this is a data-integrity risk, not a missing-evidence gap permissive mode is meant to tolerate. New regression test: `tests/test_pack_shape_purge.py::TestPurgeAndRecordEscalation::test_purge_failure_raises_unconditionally_not_only_in_strict_mode`.

---

napetrov:
`codecov/patch` is failing (75.79% vs. the 90% target). This is not a blocking check: the repo's own `codecov.yml` documents patch/project coverage statuses as advisory unless separately marked as a required branch-protection check, and this PR's checks show no `codecov/*` context among the `(required)`-tagged ones (`test-action`, `docs-pr`).

I verified the code from my most recent commit (46e4563 — `_purge_and_record`'s unconditional abort, `purge_external_outputs`'s return-value contract) is well covered: `abicheck/buildsource/pack_shape.py` is at 93% branch+line coverage under just the new/adjacent tests, and the new `_purge_and_record`/raise path is fully exercised by `tests/test_pack_shape_purge.py`. The lower patch-coverage number reflects the PR's full multi-commit diff (14 commits across this session) rather than this last change specifically, and the authoritative gate for this repo is the 95% floor on `scripts/verify.py --profile pr`'s `unit-pr` step, not the third-party Codecov patch check. Not pursuing further changes for this signal.

---

napetrov:
`mutmut (detector core)` failed after ~3h50m with exit code 143: "The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled." (job [99737188776](https://github.com/abicheck/abicheck/actions/runs/33469073373/job/99737188776)). This is a runner-loss infrastructure failure, not a real mutation survivor or code defect — the `--diff-scoped` step was still executing (no results were ever produced) when the runner was killed, well short of the job's own 240-minute `timeout-minutes`. Not this PR's failure to fix.

Re-ran the failed job (the one permitted re-run for a runner-loss case). Watching for the result.

---

napetrov:
`mutmut (detector core)`'s re-run (attempt 2, job [99788254271](https://github.com/abicheck/abicheck/actions/runs/33469073373/job/99788254271)) also did not complete — this time it ran the full 240-minute budget and was cancelled by the job's own `timeout-minutes: 240`, not a runner-loss signal. Root-caused this one from the `resolve` job's actual output rather than guessing:

- `MATCHED=true` (this PR touches `abicheck/checker_policy.py`/`abicheck/serialization.py`, both in the mutated-module glob list) and `MATCHED_TESTS=true` (this PR modifies `tests/test_serialization_roundtrip.py`, which matches the `mutated_tests` glob) → `require_baseline=true` for this run.
- `require_baseline=true` means `check_mutation_score.py --run --diff-scoped --scope-run-to-diff --require-baseline` needs a committed `mutation-baseline.json` to compare drift against — and there is no `mutation-baseline.json` anywhere in this repository (verified against `origin/main` and the working tree). Per `AGENTS.md`'s own mutation-testing section: *"`mutation-baseline.json` is not recorded yet; until a dispatch run writes it, the drift lanes fail closed rather than passing vacuously."* This is a pre-existing, documented, repo-wide gap this PR did not create — it would fail the same way on `main` today if triggered there.
- Separately, `--scope-run-to-diff`'s own documented behavior (`tests/CLAUDE.md`) is to *refuse to scope at all whenever any `tests/` path is also touched* — and this PR touches ~20 test files across its 14 commits. So the diff-scoped run's test-execution phase ran effectively unscoped, which is almost certainly why it took longer than the "~2 hours for the full baseline" this workflow's own comment cites, and never even reached the missing-baseline failure before being cancelled at the timeout.

Net: this check was never going to pass for this PR regardless of how long it ran, for reasons unrelated to any defect in this PR's code — it needs a maintainer-triggered `workflow_dispatch` (`write_baseline: true`) to record the missing baseline first, which the workflow's own comments describe as *"deliberately manual: accepting the current survivor set is a review decision"* — not something I can or should trigger unilaterally while driving this PR. I already used the one permitted re-run on the earlier runner-loss failure; re-running again would just reproduce this identical outcome, so I'm not spending a second one on it.

---

PR #985, opened 2026-09-01
fix: str-subclass-enum downcast in closure-marker walk, anonymous-tag vocabulary gap

napetrov:
## Summary

Fixes two regressions surfaced by real-world corpora (oneDNN, oneTBB) reported against this repo.

## Motivation

1. `abicheck compare` crashed with `AttributeError: 'str' object has no attribute 'value'` in `diff_types_abicc_parity.py`'s `removed_const_overload` detector, on **any** L2 snapshot containing a closure/anonymous-type marker anywhere in it ("any marker-bearing L2 dump still dies").
2. A checkout-dependent absolute path could leak into an L5 source-graph node identity (and a stripped flat closure-marker spelling), producing a spurious `declaration_renamed` between two builds of an otherwise-unedited declaration.

## Changes

- **Root cause of (1):** `Param.kind` is a `ParamKind(str, Enum)`, which satisfies `isinstance(value, str)`. `qualified_name_segments._collect_strings`/`_walk_rewrite_strings` — the generic tree-walk `storage.snapshot_load_normalization` runs unconditionally on every snapshot load — treated it as ordinary rewritable free text and handed it to `name_classification.strip_anonymous_type_location`, whose `re.sub` call returns a genuinely new, plain `str` object even on zero substitutions (unlike this module's own identity-preserving `apply_anonymous_type_ordinals` fast path). Since `ParamKind`'s own values (`"value"`, `"pointer"`, ...) never contain a marker, the rewrite was always a content no-op but never a *type* no-op — silently downcasting every `Param.kind` reachable from a marker-bearing snapshot, only surfacing later as the reported `AttributeError`. Both walk primitives now exclude any `str`-subclass `Enum` member, closing the downcast for any such field, not just `ParamKind`.
- **Root cause of (2):** `_BARE_ANON_TYPE_LOCATION_RE` (`model/graph_identity.py`) and `_ANON_TYPE_LOCATION_PATH_ONLY_RE` (`name_classification.py`) recognized the `lambda`/`unnamed <kind>` anonymous-tag vocabulary but not the `anonymous <kind>` spelling a real corpus was observed to carry (e.g. `"anonymous union at /.../concurrent_hash_map.h:649:9"`). Both regexes — plus `qualified_name_segments`'s ordinal-renumbering marker prefix and its marker-detection guard, the same vocabulary reused for stable-ordinal line-drift resilience — now also recognize `anonymous <kind>`.

`abicheck/qualified_name_segments.py` sits at ADR-061's fresh-file 800-line cap with zero headroom (a `new-file-size` architecture-gate error, not a style nit), so the fix there is written as inline `isinstance(...) and not isinstance(..., Enum)` checks rather than a new helper function, to land at exactly 800 lines. `abicheck/name_classification.py` is separately debt-baselined at exactly 1258 lines (`architecture/debt.yaml`, `no_growth` rule) and lands at exactly that too. Both gates pass; the only remaining `check_architecture.py` error (`cli_buildsource_helpers.py`) is pre-existing and untouched by this PR (verified via `git stash`).

## Bug-fix test contract

- Bug class: (1) A generic tree-walking string-rewrite primitive's overly broad `isinstance(value, str)` check silently downcasts a `str`-subclass `Enum` field to a plain `str`, because `re.sub` (unlike a hand-written identity-preserving fast path) always returns a new plain-`str` object even on zero substitutions. (2) `identity.environment_taint` (`tests/regressions/manifest.py`) — a checkout-dependent absolute path survives into canonical identity because a location-marker regex's anonymous-tag vocabulary alternation was incomplete.
- Publicly observable failure: (1) `abicheck compare a.json b.json` crashes with `AttributeError: 'str' object has no attribute 'value'` instead of producing a report, for any snapshot pair where either side has ≥1 closure/anonymous-type marker anywhere and ≥1 function with parameters (i.e. almost any real-world L2 header-AST snapshot of a non-trivial C++ library). (2) `compare`/a source-graph diff reports a spurious `risk declaration_renamed` for a completely unedited declaration, purely because it was built from a different checkout path/absolute directory.
- Regression test fails on base: Yes — `tests/test_param_kind_enum_identity.py` (all 7 cases) and the new cases in `tests/test_source_graph_directory_taint.py`/`tests/test_anon_type_location_properties.py` all fail on `main` (verified locally via `git stash` on the fix commit, re-running the tests, then restoring).
- Negative control: (1) `test_walk_rewrite_strings_still_rewrites_a_real_marker_elsewhere` asserts an ordinary string field's real marker is still normalized — the fix doesn't turn the whole rewrite into a no-op, only the enum-typed field is left alone. (2) `test_type_node_id_strips_checkout_directory_from_bare_anonymous_marker` asserts a genuinely different header, and a genuinely different anonymous-tag kind at the same coordinates, both still stay distinct (no over-merging).
- Public-surface test: `test_const_overload_detector_does_not_crash_on_marker_bearing_snapshot` and `test_snapshot_from_dict_round_trip_preserves_param_kind_enum` go through `checker.compare()` and `serialization.snapshot_from_dict()` (the real snapshot-load boundary `abicheck compare` uses), not just the internal walk helper directly.
- Axes covered: Both directions of the walk (`_collect_strings` and `_walk_rewrite_strings`), dict keys as well as dataclass fields, the load-time normalization path (`storage.snapshot_load_normalization`) and the dump-time ordinal-renumbering path (`qualified_name_segments.renumber_anonymous_closure_identities`), and both the parenthesized (`(anonymous union:...)`) and bare (`anonymous union at ...`) marker shapes.
- General invariant: (1) No `str`-subclass `Enum` member reachable from `_LAMBDA_IDENTITY_FIELDS` is ever passed to a free-text rewrite function or collected as free text, regardless of that enum's own vocabulary — stated and tested directly against the walk primitives (`TestWalkPrimitivesPreserveEnumIdentity`), not only through the one call site that happened to crash. (2) Already covered by the registered `identity.environment_taint` class's invariant ("canonical identity is a function of semantic scope and source identity — never of checkout root, absolute path spelling...") — this PR extends that class's existing property-based seed test (`test_anon_type_location_properties.py`'s `_KIND` strategy now includes the `anonymous <kind>` vocabulary) and adds `tests/test_source_graph_directory_taint.py` to its `seed_tests`.
- Must-merge / must-not-merge pair: `test_type_node_id_strips_checkout_directory_from_bare_anonymous_marker` (`tests/test_source_graph_directory_taint.py`) states both halves for the `identity.environment_taint` invariant this PR extends: must-merge — the identical `"anonymous union at ...concurrent_hash_map.h:649:9"` declaration under two different checkout roots (`/tmp/abicheck-eval/...` vs. `/mnt/pr-gate/...`) normalizes to the same node id; must-not-merge — a genuinely different header (`other.h`) at the identical coordinates, and a genuinely different anonymous-tag kind (`anonymous struct` vs. `anonymous union`) at the identical coordinates in the *same* header, both stay distinct from it. The companion property tests in `test_anon_type_location_properties.py` (`test_same_declaration_different_checkout_roots_match` / `test_different_headers_same_position_stay_distinct` / `test_different_positions_in_one_header_stay_distinct`) now run this exact must-merge/must-not-merge pair generatively over the new `anonymous <kind>` vocabulary too, not just the pre-existing `lambda`/`unnamed <kind>` ones.

## Checklist

- [ ] Docs updated if user-visible behaviour changed (bug fix only, no user-facing behavior/doc change)
- [x] `ruff check` / `ruff format --check` / `mypy abicheck/` all pass on every touched file

https://claude.ai/code/session_01TCTABQkrLsFxxLULvMfqBW


napetrov:
Pushed an additional commit (`49050e0`) with test coverage for the same bug class through a different reachable field, found independently against real oneDAL corpora rather than oneDNN/oneTBB: `Function.visibility`/`Variable.visibility` (`Visibility(str, Enum)`) gets demoted through the identical `_walk_rewrite_strings` path this PR fixes for `ParamKind`, crashing `diff_symbols._check_removed_function`'s `.visibility.value` access.

`TestStrSubclassEnumsSurviveTheWalkUnrewritten` in `tests/test_lambda_identity_ordinal.py` adds: three independent str-Enum types at the primitive level (`Visibility`/`AccessLevel`/`ParamKind`), container-position coverage, the real `renumber_anonymous_closure_identities` path with a sibling closure marker, and an end-to-end `snapshot_from_dict` load of a legacy (raw-marker) baseline. Verified locally to fail against `a193f81a` (pre-fix) and pass against this PR's fix — no changes needed to the fix itself, it already generalizes correctly. Also registered as an additional `seed_tests` entry on the `serialization.str_enum_downcast_via_generic_rewrite` bug class already added here.

---

PR #987, opened 2026-09-01
Fix str-Enum field demotion in _walk_rewrite_strings

napetrov:
## Summary

Fix a crash when comparing stored snapshots containing lambda/anonymous-type markers, caused by `_walk_rewrite_strings` demoting `str`-subclass `Enum` fields (like `Function.visibility`) to plain strings.

## Motivation

When a snapshot contained a lambda/anonymous-type marker anywhere in its `functions`/`variables`/`types`/`enums`, the `_walk_rewrite_strings` function would walk over all reachable strings — including enum members like `Visibility.HIDDEN`. The rewrite callbacks (`strip_anonymous_type_location`, `apply_anonymous_type_ordinals`) always return a new plain `str` object, even when content is unchanged. The walk's object-identity check (`new is old`) would then see a change and `setattr` the plain string back onto the dataclass field, permanently demoting `Function.visibility` from `Visibility.HIDDEN` to the bare string `"hidden"`. Every later `.visibility.value` access would crash with `AttributeError: 'str' object has no attribute 'value'` on any `compare()` involving that snapshot.

Closes #20260901

## Changes

- **`abicheck/qualified_name_segments.py`**: Modified `_walk_rewrite_strings` to skip rewriting any `str` subclass that is not exactly `str` (e.g., `Visibility`, `AccessLevel`, `ParamKind` enum members). These closed-vocabulary fields are never free-text spellings that could contain closure markers, and should pass through unrewritten.

- **`tests/test_lambda_identity_ordinal.py`**: Added comprehensive test class `TestStrSubclassEnumsSurviveTheWalkUnrewritten` with:
  - Tests verifying enum members survive the walk as the same object
  - Tests confirming plain strings are still rewritten normally
  - Tests covering enum members in containers alongside rewritten strings
  - Integration test through the real `renumber_anonymous_closure_identities` entry point
  - End-to-end test through the real load path (`snapshot_from_dict` with legacy snapshot format)

- **`changelog.d/20260901_193216_noreply_visibility_enum_demotion_snapshot_mgvhkn.md`**: Added changelog entry documenting the fix.

## Bug-fix test contract

- **Bug class**: Enum field type demotion via identity-preserving rewrite
- **Publicly observable failure**: `AttributeError: 'str' object has no attribute 'value'` when calling `compare()` on any snapshot with a lambda/anonymous-type marker
- **Regression test fails on base**: Yes — `TestStrSubclassEnumsSurviveTheWalkUnrewritten` tests all fail without the fix
- **Negative control**: `test_a_plain_str_sibling_is_still_rewritten` confirms the fix doesn't skip all string rewriting
- **Public-surface test**: `test_function_visibility_survives_renumbering_a_sibling_closure` and `test_loading_a_legacy_snapshot_preserves_visibility_enum_identity` exercise the real entry points
- **Axes covered**: 
  - All three enum types used in the model (`Visibility`, `AccessLevel`, `ParamKind`)
  - Multiple container shapes (tuple, list, dict)
  - Both production entry points (renumbering and on-load normalization)
- **General invariant**: Any `str`-subclass enum member that enters `_walk_rewrite_strings` must exit as the same object with the same type, regardless of whether sibling strings in the same container are rewritten

## Checklist

- [x] Tests added for new behaviour (comprehensive test class with 7 test methods)
- [x] No user-visible behaviour changed (internal fix)
- [x] Pre-commit checks pass
- [x] No new mypy or ruff errors

https://claude.ai/code/session_01V9wimXjzAmev3HCa7EWCdA


napetrov:
Closing as a duplicate of #985.

Both PRs fix the identical root cause in `_walk_rewrite_strings`/`_collect_strings` (`abicheck/qualified_name_segments.py`) — a `str`-subclass `Enum` field getting silently demoted to a plain `str` by the closure-marker rewrite walk. #985 was opened first, independently, via a different reachable field (`ParamKind` rather than `Visibility`), and its fix (`isinstance(value, str) and not isinstance(value, Enum)`) is a superset that also covers `Visibility`/`AccessLevel`/`ScopeOrigin`/`ElfVisibility`. It additionally fixes a related anonymous-tag vocabulary gap this PR doesn't touch.

The `TestStrSubclassEnumsSurviveTheWalkUnrewritten` regression suite this PR adds (Visibility-specific primitive coverage, the real `renumber_anonymous_closure_identities` path, and an end-to-end `snapshot_from_dict` load) has already been ported onto #985's branch and verified to pass against its fix — see [#985's comment](https://github.com/abicheck/abicheck/pull/985#issuecomment-5499400231). Merging both would conflict on the same lines in both `qualified_name_segments.py` and `tests/test_lambda_identity_ordinal.py` for no additional coverage.

---

PR #989, opened 2026-09-02
feat(adr-063): best-effort entity_id wiring for typedefs/constants (Phase 2)

napetrov:
## Summary

Wires `entity_id` onto `Change` for typedef and constant findings, closing one of the three items ADR-063 Phase 2 lists as still open.

## Motivation

Auditing ADR-063 Phase 0-4's status text against the real code (per the linked session) confirmed three genuinely open Phase 2 items:

1. `entity_id_for_typedef`/`entity_id_for_constant` had zero production call sites.
2. DWARF/PE/Mach-O/ELF-symbol-table detectors have no `EntityId`-carrying object to wire at all.
3. The `entity:` alias in `finding_identity.py` stays additive-only, never promoted to a real `primary_id` match tier.

This PR closes (1). (2) and (3) are **not** attempted here — see "Deliberately out of scope" below for why forcing either would be unsafe.

## Changes

- `abicheck/model/identity.py`: adds `scope_path_from_qualified_string()` — a documented, explicitly *lossy* `ScopePath` fallback for a flat `::`-qualified name string (every segment becomes `Namespace`, since a bare string can't distinguish namespace/record/inline-namespace).
- `abicheck/diff_types.py`: `_diff_typedefs` now builds `entity_id_for_typedef(...)` per alias and passes it to all three emitted kinds (`TYPEDEF_REMOVED`, `TYPEDEF_BASE_CHANGED`, `TYPEDEF_VERSION_SENTINEL`).
- `abicheck/diff_symbols.py`: `_diff_constants` now builds `entity_id_for_constant(...)` per name and passes it to all three emitted kinds (`CONSTANT_REMOVED`, `CONSTANT_CHANGED`, `CONSTANT_ADDED`).

## Deliberately out of scope (not silently dropped)

- **DWARF-side `EntityId` extension**: `dwarf_metadata.py`'s DFS scope tracking (`_walk_die_iter`) is a flat string concatenation, and the resulting struct/enum entries are stored in `DwarfMetadata.structs`/`.enums` **keyed by that flat string**, consumed at 12 other call sites. Giving this a typed `ScopePath` means redesigning that keying scheme, not adding a field — real extraction-side surgery I can't validate with confidence without exercising every DWARF edge case (anonymous namespaces, template instantiations, versioned inline namespaces) against real binaries.
- **Promoting the `entity:` alias to a real `primary_id` tier**: `model/identity.py`'s own docstring records this was attempted and reverted twice, because `Anonymous`/`LocalToFunction` ordinals are only stable within one process's lifetime, not across releases. Promoting the alias before that instability is solved would reintroduce the exact bug those two reverts exist to prevent.

## Limitations of what *is* included

This is explicitly the interim, string-derived identity — not the full fix. A real fix needs new structured `Typedef`/`Constant` model types, parser changes in both header-AST backends (castxml + clang), and a schema bump, so two distinct qualified names differing only in segment *kind* (a namespace vs. a same-named nested record scope) can still collide into the same synthesized identity here. Documented in the new helper's own docstring.

## Verification

- `mypy abicheck/` — clean (0 issues).
- `python scripts/check_ai_readiness.py` — 0 errors, same 137 warnings as baseline (no new warnings).
- `pytest tests/ -k "typedef or constant" -m "not integration and not libabigail and not abicc and not slow and not golden"` — 743 passed, 1 skipped.
- Full fast test suite run in progress at time of PR open; will report back if it surfaces anything.

https://claude.ai/code/session_01JabiEWZxZm2XdkCfypmhVr

---
_Generated by [Claude Code](https://claude.ai/code/session_01JabiEWZxZm2XdkCfypmhVr)_


napetrov:
Closing without merging: a concurrent PR (#982, commit `d0f4c4b`) landed on `main` while this one was open, implementing the *real* fix for the same gap — parse-time-resolved `AbiSnapshot.typedef_entity_ids`/`constant_entity_ids`, populated by both header-AST backends and persisted through `storage/entity_id_codec.py` — which fully supersedes this PR's best-effort, string-derived fallback (`scope_path_from_qualified_string()`).

Per ADR-063's own governing invariant ("one concept, one representation — a second implementation is a defect, full stop"), I merged `main` into this branch, took #982's implementation over this PR's own, and deleted the now-dead fallback helper rather than keep both around. The branch is now identical to `main` (`git diff origin/main...HEAD` is empty), so there's nothing left for this PR to contribute — closing rather than merging a no-op.

https://claude.ai/code/session_01JabiEWZxZm2XdkCfypmhVr

---

PR #1072, opened 2026-09-05
refactor: retire the dump CLI's dead ELF/PE execution helpers (ADR-063 T1)

napetrov:
## Summary

Retire `cli_dump_helpers.perform_elf_dump` and `cli_dump_non_elf.handle_non_elf_dump` — dead code since ADR-063 Phase 1 — after rehoming the assertions whose only home was one of them onto the path that actually runs (Track 1 of the duplication-and-convergence plan).

## Motivation

ADR-063 Phase 1 routed `dump_cmd`'s real run, for ELF *and* PE/Mach-O, through the shared typed executor (`service_dump_pipeline.execute_dump_request` via `frontends/cli/dump_execute.py`). Both legacy execution functions lost their production caller then and were kept only because their own unit tests still imported them — the exact shape `docs/contribute/plans/duplication-and-convergence-assessment.md`'s "explicit retirement table" names as T1, and the plan's Phase 6 item 1.

## Changes

- **Deleted** `cli_dump_helpers.perform_elf_dump` (661 lines), `abicheck/cli_dump_non_elf.py`, and `abicheck/cli_dump_protocols.py` (its `Protocol` types described only the callables `dump_cmd` passed into those two). No behaviour change: nothing invoked either function.
- **Rehomed** the assertions with no other home into a new `tests/test_dump_cli_execution_behaviors.py`, driving the real `dump` CLI:
  - `header_roots` restated as the cross-path **equivalence** it always was — asserted against the root set `service.run_dump`'s own dependency-scope choke point derives for the same invocation, not a second copy of that function's formula written into the test. Both binary formats, several `-H` shapes, plus the `--dump-manifest` case where the two computations are genuinely independent. Mutation-checked: dropping the manifest term from the CLI side fails the test.
  - public-root forwarding to the write-time L4 embed; the `--dwarf-only` `from_headers` stamp gate (stated over that whole axis); inferred-header-root vs. compile-context search ordering; context include dirs reaching `extra_hash_dirs` (over each token spelling); `--include-system-declarations` reaching the streaming pruner; the PE `--follow-deps` notice; extraction-failure translation (over the whole caught-exception set).
  - The rest of what those tests pinned was **not** unique — it re-asserted, at a retired call site, behaviour the shared pipeline already owns and tests at its own seam. The new module's docstring names each owner (`test_typed_dump_request.py`, `test_header_compile_context.py`, `test_service_unit.py`, `test_service_input_resolution.py`, `test_legacy_compile_db_typed_threading.py`, `test_dumper_scoping.py`).
- **Recorded, not reverted:** one assertion could not be rehomed because the behaviour changed with the earlier migration. A `-H <dir>` operand used to reach `dumper.dump`'s `scope_header_dirs` (extraction scope only, ADR-015 provenance tagging deliberately off); the typed request splits `-H` with `split_public_header_inputs` and passes the directory as a real `public_header_dirs` entry, so `dump` now tags provenance for it — matching what `compare` has always done with its own `-H` list. New `docs/contribute/known-gaps.md` entry covers why it stands and what is *not* established.
- Housekeeping: `CLI_CONTRACT_ALLOWLIST` loses its `cli_dump_helpers.py` entry (the first line to leave that list, by deletion rather than rerouting); `architecture/modules.yaml` drops the two deleted modules; `architecture/debt.yaml` ratchets `cli_dump_helpers.py` 1623 → 962 and drops the now-paid-off `test_cli_dump_helpers_coverage.py` entry; stale prose across `abicheck/` and the ADR/plan docs updated where it made a present-tense claim about a function that no longer exists.
- `tests/test_cli_coverage.py::test_elf_dump_error` was patching `cli_dump_helpers`' own inert `dump` alias rather than the binding `_resolve_input` actually resolves — the patch was a no-op and the expected `ClickException` came from a real parse failure. Retiring the alias surfaced it; now patches `abicheck.dumper.dump`.

## Verification

`python scripts/verify.py --profile pr`: lint, typecheck, architecture, fp-rate, tier-accuracy, docs/schema/metadata gates all pass; `unit-pr` reports **96.51 %** coverage (floor 95 %) with 38 108 passing. Four failures on that run are pre-existing in this environment and reproduce on a clean tree: three `adr-status-sync` receipts naming a commit not reachable from `origin/main` here, and `test_verify_profiles.py::test_isolated_module_runner_preserves_user_site_without_root_shadowing`. `fmt-check` flags ~200 files repo-wide including untouched ones (`abicheck/__init__.py`) — a ruff-version difference in this environment, not this diff.

## Checklist

- [x] `ruff`/`mypy`/`check_architecture`/`check_ai_readiness` clean (no new findings)

https://claude.ai/code/session_01TQrbuwHStmQYcdo6pMWG4u

---
_Generated by [Claude Code](https://claude.ai/code/session_01TQrbuwHStmQYcdo6pMWG4u)_


napetrov:
### `unit-tests (windows-latest, 3.13)` was cancelled — a job-timeout race, not a test failure, and not this PR's

Standing down on this one check, with the evidence, rather than silently re-running.

**Every test passed.** The job's own log ends with the full pytest summary *before* the cancellation:

```
== 37448 passed, 719 skipped, 4 xfailed, 175 warnings in 1758.64s (0:29:18) ===
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
##[error]The operation was canceled.
```

Zero failures. The `KeyboardInterrupt` lands in pytest's epilogue (it is printing the `slowest 25 durations` table), after the session had already reported success. The job ran 02:49:30 → 03:19:44 = **30m14s** against `unit-tests`' `timeout-minutes: 30`.

**It isn't this PR's.** Two independent checks:

1. **Base-branch comparison.** On `f51df94` — this PR's exact base — the same lane ran 02:06:56 → 02:36:06 = **29m10s** and passed ([run 33937268164](https://github.com/abicheck/abicheck/actions/runs/33937268164)). The lane sits ~50 seconds under the cap on main; this run was ~14 seconds over it. That is runner variance against a budget with no headroom, not a regression. (That same base run is itself not clean, incidentally: it failed `integration-tests (windows-latest)` — a job that **passes** here.)
2. **Direction of the diff.** This PR deletes ~1,800 lines of tests and adds ~300; it removes work from the suite. It cannot have made this lane slower.

**This is a documented, recurring class in this repo.** `ci.yml`'s own comment above `timeout-minutes: 30` records the previous occurrence verbatim — the budget was raised from 20 after the lane was *"cancelled at exactly 20m01s with every single step, coverage floor included, already reporting success (a job-level timeout racing its own last step, not a real failure)"* — and predicts this recurrence: *"a fixed budget with days-old headroom keeps re-drifting to zero as tests are added anywhere in the fast lane."* Headroom on the Windows leg has now drifted to roughly zero again.

**What I did and did not do.** I attempted the one permitted re-run; GitHub rejected it (`403 This workflow is already running`) because `test-action` is still in progress in the same run. I will retry it once the run completes. I have **not** raised `timeout-minutes`, sharded the Windows lane, or touched any test to get green — the first two are a repo-wide CI-budget decision well outside a dead-code retirement, and the third is never acceptable. If maintainers want the underlying drift addressed, that is its own change; the fix shape is either more headroom or splitting the Windows leg, and it should be decided on the whole matrix's numbers rather than bolted onto this PR.

Everything else on `47c3fec` is green, including the canonical coverage lane (`unit-tests (ubuntu-latest, 3.13)`, 96.51% against the 95% floor), `lint-and-types`, `ai-readiness`, `architecture`, all three integration lanes, both parity lanes, and `codecov/patch`.

---

PR #1091, opened 2026-09-05
fix: close ADR-063 T9's legacy-hybrid backfill blocker

napetrov:
## Summary

ADR-063 T9 ("Fact provenance and scope") landed its first slice on
2026-09-05 (`Fact[T]` gained a `producer` field, closing the PDB `vtable`
fabrication). This PR closes the item's second named blocker: **the
legacy-hybrid backfill blocker** holding seven `fact_provenance`-gated
case-(a) fields (`RecordType.deprecated`, `EnumType.deprecated`/`is_scoped`,
`Function.deprecated`, `Variable.deprecated`, `TypeField.deprecated`/
`default`).

### The bug

These seven fields are gated at compare time by
`AbiSnapshot.fact_provenance` (a per-declaration map) rather than a
snapshot-level flag, because a `--ast-frontend hybrid` merge's two backends
can disagree on which of them actually populated any one declaration's
fact. `clang_deprecation_facts_reliable` reads `True` unconditionally for a
hybrid producer (an ordinary, fresh hybrid dump states these facts
explicitly per declaration, so the flag has no reason to distrust it) — so
loading a document that predates this fact family's own schema version
previously reconstructed every one of these facts as a confirmed `PRESENT`,
including for a declaration neither backend's merge ever actually recorded
looking at, since the legacy JSON format always serializes some value
(there is no "omitted" concept once a dict round-trips through a dataclass
constructor).

### The fix

`storage/fact_backfill.py`'s legacy-load correction (`CaseAFactRule`)
gains a fifth, optional field, `fact_provenance_kind` — additive and
`None` by default, so every pre-existing rule and call site is
unaffected. Set on the seven affected rules, it makes
`apply_case_a_fact_backfill` consult the document's own per-declaration
`fact_provenance` map whenever `ast_producer == "hybrid"`: a declaration
with no recorded provenance entry (probed namespace-qualified first,
falling back to the former bare key only when this document's own side has
no other declaration sharing that bare name — the same shape
`fact_provenance.resolved_fact_producer` already applies at compare time)
downgrades the claim to `NOT_COLLECTED`, following the module's
pre-existing "downgrade the claim, never the value" convention (a real
non-resting legacy value is left untouched either way).

## Bug-fix test contract

- Bug class: `storage.legacy_availability_collapse` (new entry added to `tests/regressions/manifest.py`, since no existing class matched this shape — a legacy document's per-declaration fact must never be read as a confirmed, positively-observed value on a whole-snapshot reliability flag alone when a finer-grained availability signal recorded on that same document is silent about that exact declaration).
- Publicly observable failure: Loading a `--ast-frontend hybrid` snapshot from a pre-v40 (or pre-v39 for `TypeField`) JSON document reports `deprecated`/`is_scoped`/`default` as a confirmed `PRESENT` fact (e.g. "confirmed not deprecated") for a declaration neither backend's merge ever recorded looking at, so a `compare` run can silently treat an unconfirmed absence as a real observation instead of `NOT_COLLECTED`.
- Regression test fails on base: Yes — `tests/test_deprecation_family_facts.py::TestLegacyHybridProvenanceBackfill::test_no_provenance_entry_downgrades_a_resting_default` and its `RecordType`/`EnumType`/`Function`/`Variable`/`TypeField` siblings assert `NOT_COLLECTED`; on `main` (pre-fix) `CaseAFactRule` has no `fact_provenance_kind`, and `clang_deprecation_facts_reliable`/`clang_field_initializer_facts_reliable` read `True` unconditionally for a hybrid producer, so every one of these resolves to `PRESENT` instead.
- Negative control: `test_confirmed_by_qualified_provenance_key_stays_present`, `test_bare_key_fallback_when_unambiguous`, `test_no_provenance_entry_preserves_a_real_non_default_value`, and `test_non_hybrid_producer_is_unaffected` assert the fact stays `PRESENT` (or the real legacy value is preserved) whenever provenance genuinely confirms it, the producer isn't hybrid, or the value is a real non-resting one — the fix narrows exactly the ambiguous case, not the whole fact family.
- Public-surface test: Every new test goes through `abicheck.serialization.snapshot_from_dict` — the real, public JSON-load chokepoint every CLI/typed-API caller uses to read a stored/cached snapshot — not a hand-rolled shortcut directly into `apply_case_a_fact_backfill`/`_hybrid_provenance_confirms`.
- Axes covered: `ast_producer` (hybrid vs. castxml/clang single-backend); owner kind (`RecordType`/`EnumType`/`TypeField`/`Function`/`Variable`); provenance-key shape (namespace-qualified, bare-fallback unambiguous, bare-fallback declined-as-ambiguous, mangled-name-keyed).
- General invariant: `storage.legacy_availability_collapse` (`tests/regressions/manifest.py`) — backed by `TestLegacyHybridProvenanceBackfill`'s independently-chosen sibling cases across all five owner kinds plus all four provenance-key shapes, checked against the qualified-then-bare-with-ambiguity-guard shape `fact_provenance.resolved_fact_producer` already applies at compare time as the oracle (this fix's lookup logic mirrors that algorithm independently rather than calling it, since `storage/` may not import the flat-root `fact_provenance.py`).

Real-dependency test: every new test drives the real `snapshot_from_dict` JSON-decode chokepoint (the format boundary this fix corrects) end to end — building a raw legacy-shaped dict and reading it back through the actual decoder, not a hand-constructed shortcut into the correction function directly — the same pattern this file's own pre-existing `TestDeprecatedFamilyRoundTrip`/`TestEnumIsScopedFact` classes use.

## Testing

- New test class: `tests/test_deprecation_family_facts.py::TestLegacyHybridProvenanceBackfill`
  — covers the qualified-key, bare-key-fallback (unambiguous and
  ambiguous), mangled-name-keyed, value-preserving, and
  non-hybrid-producer-unaffected shapes.
- New bug-class registry entry: `tests/regressions/manifest.py`'s
  `storage.legacy_availability_collapse`.
- `ruff check`, `ruff format --check`, `mypy abicheck/`: clean.
- `python scripts/check_architecture.py`: 0 errors.
- `python scripts/check_ai_readiness.py`: no new errors (4 pre-existing,
  unrelated ADR-verification-receipt errors, confirmed present at the
  PR's base commit too).
- `python scripts/check_fp_rate.py`: 0 FP / 0 FN, unchanged.
- `python scripts/check_tier_accuracy.py`: unchanged.
- Verified the original investigation's own regression scenario still
  passes: `tests/test_dumper_hybrid.py::TestNamespaceQualifiedMerging::test_legacy_bare_keyed_hybrid_baseline_still_detects_transition`
  (which, on inspection, never exercises this JSON-load-time code path at
  all, since it constructs its snapshot in-memory rather than through
  `snapshot_from_dict`).
- Full fast unit-test suite (`pytest tests/ -m "not integration and not
  libabigail and not abicc and not slow and not golden"`): 39741 passed, 4
  failed. Confirmed all 4 failures pre-exist at this PR's base commit
  (`d1a486bb`), unrelated to this diff — `test_action_run_sh_baseline_set_fallback.py`,
  `test_ai_readiness.py::test_adr_status_sync_holds`/`test_main_returns_zero_on_clean_tree`
  (the same pre-existing ADR-verification-receipt issue noted above), and
  `test_verify_profiles.py::test_isolated_module_runner_preserves_user_site_without_root_shadowing`
  (an environment/venv isolation issue in this sandbox).

## Documentation

Records this slice in
`docs/contribute/plans/duplication-and-convergence-assessment.md`'s T9
tracking row/note, and adds a `changelog.d/` fragment.

Still open, unchanged by this PR: the DWARF per-translation-unit
completeness gap (needs new evidence infrastructure the model doesn't
carry yet) and the shared analysis accounting for declined comparisons —
both recorded as T9's remaining scope in the plan doc.

https://claude.ai/code/session_01PV4ZWN7nVotQukVwtURg8S

---
_Generated by [Claude Code](https://claude.ai/code/session_01PV4ZWN7nVotQukVwtURg8S)_

### DIFF
--- a/abicheck/serialization.py
+++ b/abicheck/serialization.py
@@ -1088,6 +1088,11 @@ def snapshot_from_dict(d: dict[str, Any]) -> AbiSnapshot:
         clang_deprecation_facts_reliable_value=clang_deprecation_facts_reliable_value,
         clang_restrict_facts_reliable_value=clang_restrict_facts_reliable_value,
         castxml_var_access_facts_reliable_value=castxml_var_access_facts_reliable_value,
+        # T9 / ADR-063 Phase 6 item 4 (legacy-hybrid backfill blocker): the
+        # raw per-declaration provenance map, same source as the
+        # `AbiSnapshot.fact_provenance` field constructed further below --
+        # read here too since this call happens before that field exists.
+        fact_provenance=dict(d.get("fact_provenance", {})),
     )
 
     # ADR-050 D1 (schema v12) — profile/scope fingerprints. Missing key (every
--- a/abicheck/storage/fact_backfill.py
+++ b/abicheck/storage/fact_backfill.py
@@ -29,11 +29,44 @@ module holding the per-field thresholds both this module and ``fact_codec``
 need. Importing them back from ``fact_codec`` instead would be a real import
 cycle (``fact_codec`` re-exports this module's three public names for
 compatibility), which the ``import-cycle-growth`` gate rejects.
+
+**T9 / ADR-063 Phase 6 item 4 (the "legacy-hybrid backfill blocker"):** the
+whole-snapshot ``reliable``/``evidenced`` framework above cannot see the one
+ambiguity that only exists on a ``--ast-frontend hybrid`` snapshot: seven
+fields (``RecordType.deprecated``, ``EnumType.deprecated``/``is_scoped``,
+``Function.deprecated``, ``Variable.deprecated``, ``TypeField.deprecated``/
+``default``) are gated, at COMPARE time, by ``fact_provenance.py``'s
+per-declaration ``AbiSnapshot.fact_provenance`` map rather than by a
+snapshot-level flag -- because a hybrid merge's own two backends can
+disagree on which of them actually populated any one declaration's fact,
+which no whole-snapshot boolean can express. ``clang_deprecation_facts_
+reliable`` reads ``True`` for a hybrid producer unconditionally (an
+ordinary, fresh hybrid dump states this fact explicitly per declaration, so
+the flag has no reason to distrust it), so the ``unreliable``/
+``unproduceable`` checks above both pass for a legacy (pre-``min_schema_
+version``) hybrid document, and the pre-existing ``__post_init__`` bridge's
+``Fact.present(raw_value)`` is left standing -- even for a declaration
+neither backend actually confirmed on that document, since the legacy JSON
+format always serializes SOME value (there is no "omitted" concept once a
+dict round-trips through a dataclass constructor). ``fact_provenance_kind``
+on :class:`CaseAFactRule`, consulted only when the
+document's ``ast_producer`` is exactly ``"hybrid"`` and only for a rule
+that names one, resolves this the same way ``fact_provenance.
+resolved_fact_producer`` already does at compare time: probe the
+declaration's namespace-qualified provenance key first, falling back to the
+bare pre-qualification key only when no OTHER declaration on this same
+snapshot shares that bare name (an ambiguous bare fallback would misattribute
+one declaration's real provenance to an unrelated sibling). No entry under
+either key means neither backend's merge recorded ever having looked --
+which downgrades the *claim* exactly like an ``evidenced``-only miss above
+(never the value, and only when the value is already the field's own
+resting default -- the existing "downgrade the claim, never the value"
+convention this module's own callers already rely on).
 """
 
 from __future__ import annotations
 
-from collections.abc import Iterator
+from collections.abc import Iterator, Mapping
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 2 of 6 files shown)


napetrov:
**`mutmut (detector core)` failure — not this PR's, root-caused, one re-run applied.**

The job aborted with `mutmut run` exiting 1 during its clean-tree baseline pass ("stopping after 1 failures" / "failed to collect stats"), before any mutant was even generated.

This isn't `--scope-run-to-diff` narrowing to `storage/fact_backfill.py`'s own tests — this PR's diff also touches `changelog.d/*.md`, `docs/contribute/plans/duplication-and-convergence-assessment.md`, and `tests/regressions/manifest.py`, all outside `only_mutate`, and `scripts/check_mutation_score.py`'s own `diff_touches_outside_only_mutate` docstring names exactly this shape ("this repo's own changelog-fragment convention means most real PRs touching one `only_mutate` module also touch `changelog.d/*.md`, which now disables scoping too") — so this run fell back to a full, unscoped baseline over the whole suite.

A full-suite baseline is exactly where this PR's own description already found and root-caused 4 pre-existing failures on this PR's base commit (`d1a486bb`, confirmed unrelated to this diff by running the same 4 against a clean checkout of that commit): `test_action_run_sh_baseline_set_fallback.py`, `test_ai_readiness.py::test_adr_status_sync_holds`/`test_main_returns_zero_on_clean_tree` (a pre-existing ADR-verification-receipt drift, deterministic regardless of environment), and `test_verify_profiles.py::test_isolated_module_runner_preserves_user_site_without_root_shadowing` (an environment/venv-isolation-sensitive test). The job's own log truncates `mutmut run`'s output to its last 5 lines (`check_mutation_score.py`'s own `tail = "\n".join(run_out.splitlines()[-5:])`), so I can't name which of the 4 fired the abort from the CI log alone — but the mechanism (unscoped full-suite baseline hitting an already-known-flaky/pre-existing failure) is the same either way, and none of the 4 involves code this PR touches.

No fix to port here — these are pre-existing gaps unrelated to `storage/fact_backfill.py`/`serialization.py`. Re-running the job once now to confirm.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PV4ZWN7nVotQukVwtURg8S)_

---

PR #1101, opened 2026-09-06
test: harden two environment-fragile tests against non-standard dev venvs

napetrov:
## Summary

Investigated "failures on main" and found: (1) `main`'s actual CI history is currently green — the two red runs in recent history (`Architecture contract (ADR-061)` on two near-simultaneous merge commits) were transient merge-race artifacts already resolved by subsequent merges, confirmed by a clean local `python scripts/check_architecture.py` (0 errors) and clean `ruff check`/`mypy abicheck/` on current `main`; (2) running the full fast unit suite locally surfaced two genuinely environment-fragile tests, fixed here.

## Motivation

`pytest tests/ -m "not integration and not libabigail and not abicc and not slow and not golden"` on current `main` fails 2 of ~39,843 tests when run inside a standard `python -m venv` (without `--system-site-packages`) with `abicheck` installed only into that venv — exactly the setup a contributor gets by following this repo's own documented `pip install -e ".[dev]"` instructions in a plain venv. Both tests pass in the real GitHub Actions CI because `actions/setup-python` installs directly onto the runner with no extra venv layer, which is why `main`'s CI has stayed green while these fail locally in a very standard dev setup.

## Changes

- `tests/test_action_run_sh_baseline_set_fallback.py::TestBaselineSetFallback::test_extraction_and_resolution_work_when_python3_is_absent`: build the fake `python3`-less-PATH interpreter from `sys.executable` via a `#!/bin/sh` exec wrapper instead of `shutil.which("python3")` + `Path.symlink_to()`.
  - `shutil.which("python3")` can resolve to an unrelated system interpreter with no `abicheck` installed whenever the dev environment installs `abicheck` into a separate venv than whichever `python3` is first on `PATH`.
  - A plain symlink to `sys.executable` isn't enough either: when `sys.executable` is itself a venv symlink chain down to a base interpreter (the ordinary venv shape), adding one more symlink hop from an unrelated directory makes CPython's own `pyvenv.cfg` discovery walk straight past the venv to the base interpreter, silently losing the editable `abicheck` install. An `exec` wrapper invokes `sys.executable`'s real path directly — the same single hop a normal invocation takes — so venv detection resolves correctly.
- `tests/test_verify_profiles.py::test_isolated_module_runner_preserves_user_site_without_root_shadowing`: skip when the running interpreter would not normally enable the user site (e.g. a venv created without `include-system-site-packages=true`). The test's premise — a user-site tool retains precedence over a same-named cwd module — cannot hold there regardless of `run_isolated_module.py`'s own correctness, since `site.ENABLE_USER_SITE` is already `False` for that interpreter with or without the `-I` isolation the test exercises. The skip reuses `run_isolated_module.py`'s own `_normally_enables_user_site()` predicate rather than restating the logic, so the two can't drift apart.

No `abicheck/**/*.py` touched — tests-only, no changelog fragment needed.

## Verification

- `ruff check abicheck/ tests/` — clean
- `mypy abicheck/` — clean (0 errors)
- `python scripts/check_architecture.py` — 0 errors
- `python scripts/check_ai_readiness.py` — 0 errors, pre-existing warnings only
- `pytest tests/test_action_run_sh_baseline_set_fallback.py tests/test_verify_profiles.py -q` — all pass (was 2 failing before this change)
- Full fast suite locally: 39,842 passed / 38 skipped / 4 xfailed (was 39,840 passed / 3 failed before this change)

## Checklist

- [x] Changelog fragment added (`scriv create`) if this touches `abicheck/**/*.py` — N/A, no `abicheck/**/*.py` changed
- [x] Docs updated if user-visible behaviour changed — N/A, no user-visible behavior changed
- [x] `pre-commit run --all-files` equivalent checks (`ruff check`, `mypy`) pass locally

https://claude.ai/code/session_017WvAEhC958VDntCdfav33D

---
_Generated by [Claude Code](https://claude.ai/code/session_017WvAEhC958VDntCdfav33D)_

### DIFF
--- a/tests/test_action_run_sh_baseline_set_fallback.py
+++ b/tests/test_action_run_sh_baseline_set_fallback.py
@@ -747,13 +747,13 @@ class TestBaselineSetFallback:
     @pytest.mark.skipif(
         sys.platform == "win32",
         reason=(
-            "This test builds a `python` symlink via Path.symlink_to() to "
-            "simulate a python3-less PATH -- creating a symlink on Windows "
-            "needs Developer Mode or an elevated process (raises OSError "
-            "otherwise), a privilege this test has no need to require just "
-            "to exercise a fallback that already has its own dedicated "
-            "coverage of the real Windows shape via the win32-only "
-            "`only 'python' on PATH` runners this whole PR targets."
+            "This test builds a `python` shell-script wrapper to simulate "
+            "a python3-less PATH -- a `#!/bin/sh` shebang script isn't "
+            "executable on Windows, a privilege this test has no need to "
+            "require just to exercise a fallback that already has its own "
+            "dedicated coverage of the real Windows shape via the "
+            "win32-only `only 'python' on PATH` runners this whole PR "
+            "targets."
         ),
     )
     def test_extraction_and_resolution_work_when_python3_is_absent(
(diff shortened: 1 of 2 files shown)


napetrov:
CI: `unit-tests (macos-latest, 3.13, false)` failed on this PR's head (`1312a39`), but not on anything this PR touches:

```
FAILED tests/test_dumper_scoping_dependency_retention.py::TestDirectlyReferencedDependencyRetention::test_long_typedef_chain_reachability_stays_fast - AssertionError: long typedef chain reachability took 5.05s
assert 5.052140667000003 < 5.0
```

This is a hardcoded wall-clock threshold (`elapsed < 5.0`) in a perf-regression test unrelated to this PR's diff (which only touches `tests/test_action_run_sh_baseline_set_fallback.py` and `tests/test_verify_profiles.py`). `5.052s` vs. a `5.0s` threshold on a shared macOS runner is consistent with ordinary runner-load variance, not a real regression — no code path this test exercises (`abicheck.dumper_scoping`) was touched by this PR. Re-ran the failed job once via `rerun_failed_jobs`; will follow up if it fails again.

---

PR #1110, opened 2026-09-06
feat: ADR-065 S3/S4 — package component inventories, support-promise findings, and the set-difference deletion

napetrov:
## Summary

Completes ADR-065 ("Comparison Scope, Member Selection, and Input Completeness") slices **S3** (package component inventories + support-promise findings under a contract policy) and **S4** (the deletion gate: the release fan-out's set-difference removal path), and updates the ADR's Status, its `adr/index.md` row and section A of `vision-api-abi-evolution.md` to record S1–S4 as landed.

## Motivation

After S1/S2, two things named in ADR-065 were still open in code:

* `abicheck/package.py` answered "what does this package ship?" with a **directory**, so a component the extractor failed to unpack and a component the package no longer ships were one state — and no live operand could ever supply the completeness proof D2 requires before an unmatched member becomes a removal. `AcquisitionState.EXPECTED_NOT_PRODUCED` was reserved by S2 with no producer at all.
* `_match_release_keys` still computed `old_keys - new_keys` / `new_keys - old_keys`. S2 had taken exit `8`, the verdict bump and the Markdown/PR-comment "removed" sections off it, but the plan's own deletion gate named the two readers that still consumed it *by name*: the JSON `unmatched_old`/`unmatched_new` keys and the fan-out's stderr warnings.

## Changes

**S3 — package component inventories**

- `abicheck/model/package_inventory.py` — `PackageInventory`/`PackageComponent`: declared components, an explicit `complete` flag, and a separate `unproduced` map. Keeping those three apart is the whole point; the module docstring states why collapsing any two is the failure it exists to prevent.
- `abicheck/package.py` — `ExtractResult.container_complete`, set by every archive extractor (each unpacks its container in full or raises) and deliberately **not** by `DirExtractor`; `package_component_inventory()` turns that plus the discovered members into a `PackageInventory`.
- `workflows/release_scope.release_inventory_evidence` gains the inventory as a second `PROVEN` source beside S2's stored `inventory_complete` assertion. **A package-archive operand is now a D2 completeness proof; a directory operand still proves nothing** (it may have been populated partially — the "absent is not removed" case).
- `build_release_scope_record` / `build_declared_selection_record` gain `old_unproduced`/`new_unproduced`, giving `EXPECTED_NOT_PRODUCED` its first producer: a declared component the extracted tree cannot reach (dangling link, unreadable file) is an *acquisition failure*, never an absence the other side's proof may read as a removal.

**S3 — support-promise findings (D1's fifth concept)**

- `abicheck/policy/support_promise.py` — the `--support-promise off|declared` contract-policy field (`off` by default) and the derivation, which reads *only* `proven_removed_members`/`proven_added_members`, so an unmatched member under an unproven inventory can never reach a finding whatever the policy says.
- Two new `ChangeKind`s: `support_promise_component_retired` (BREAKING) / `support_promise_component_introduced` (compatible addition), each carrying D2's completeness receipt in `old_value`/`new_value`.
- `abicheck/workflows/release_support_promise.py` places each finding in the fan-out's own `library_results` (with a real `DiffResult`), so the existing verdict fold, severity aggregation, JSON/Markdown/JUnit renderers and disposition audit see it without a parallel reporting path. Deliberately **not** a duplicate of `bundle_library_removed`, which fires only when a surviving sibling imports the missing library.

**S4 — deletion gate**

- `_match_release_keys` now returns only the matched keys and the two maps. Both remaining readers moved onto the acquisition record: the JSON keys (same meaning, `[]` rather than a re-derived difference when a driver has no record) and the stderr notices, now `report.comparison_scope.release_scope_warnings`, emitted after the record exists. That is what lets one line distinguish a **proven** removal (naming the inventory that proved it) from an unmatched member, an `expected_not_produced` one and an `out_of_scope` one — four states the set difference reported as one.
- `_collect_release_warnings` deleted; `_prepare_compare_release_inputs` no longer returns the two key lists (it returns the two inventories instead).

**Behaviour change worth calling out**

`--fail-on-removed-library` reaches exit `8` again for a **package-archive** pair (S2 had left it reachable only for a stored snapshot). A directory pair still exits `0` with an incomplete scope. Second migration note added to `docs/reference/exit-codes.md`.

**Housekeeping forced by the size gates (not incidental scope creep)**

- The wheel filename-tag/`METADATA` fact readers moved out of the `no_growth`-tracked `abicheck/package.py` into `abicheck/extract/wheel_tags.py` (ADR-061 routing: reading a declared build/packaging fact is `extract/`'s job). `abicheck.package` re-exports all of them, private helpers included, so every existing import path and test still resolves. This is what made room for S3 inside `package.py`'s baseline — per AGENTS.md, "the way to shrink an entry is to move responsibility out to a properly-owned module".
- The stored-side D2/D8 marker readers moved out of `workflows/release_scope.py` into `workflows/release_stored_inventory.py` to keep it under the 800-line production cap.

**Docs**

ADR-065 Status + `adr/index.md` row (they must not contradict — `adr-status-sync`); `vision-api-abi-evolution.md` section A (S3/S4 landed, plus a new "Still missing after S1–S4" and "Still open in S4, deliberately" paragraph); `docs/use/multi-binary.md` (inventory proof, `unmatched_old`'s new source, `--support-promise`); `docs/reference/exit-codes.md`; regenerated `detector-spec`, `cli-reference`; ChangeKind headline counts 399 → 401 across the doc set.

## Deliberately deferred (documented, not silently dropped)

* **Scalar/bundle operand convergence.** S4 names it "as a slice of `cli-cleanup-phase-two.md` PR I/J". PR I is open there with its own live/stored-driver and one-gate/report/dry-run design question, and that plan's own row already cross-references this workstream. Doing it here would be a second copy, so it stays with PR I; recorded in the plan's new "Still open in S4, deliberately" paragraph and in the ADR's slice list.
* **The silent canonical fallback in `compare_product_directories`** (D3's `ambiguous` diagnostic). The deletion gate this PR closes covers `_match_release_keys`' set difference; turning an ambiguous group into a refusal-to-compare in the whole-product path is a behaviour change with its own migration note, so it is recorded as open rather than half-done.
* **Action/project/aggregate parity** was *already* landed by S2 (`action/run.sh`'s `SCOPE_INCOMPLETE` verdict tier; aggregate report schema 1.8's `scope_completeness` axis). Verified rather than re-implemented; no second copy added.
* **`--support-promise` is a CLI flag, not an `.abicheck.yml` key** — mirroring `--on-incomplete-scope`, S2's precedent for a contract/scope policy on the release fan-out, rather than inventing a second config surface in this PR.

## Bug-fix test contract

Not applicable — this is a `feat:` PR, not a `fix:`/`perf:`/`security:` one.

## Verification

* `python scripts/verify.py --profile pr` — **20/21 checks pass**, `complete: true`. The one failure is `fmt-check`, which **fails identically on the unmodified base**: `ruff format --check abicheck/ tests/` reports 594 files would be reformatted at `origin/main`. Every one of the 8 files this PR touches that `fmt-check` flags was already unformatted at base (verified file-by-file against `origin/main`); all 5 new modules and both new test files are format-clean.
* `python scripts/check_architecture.py` — **0 errors** (including the `no_growth` baselines for `package.py`, `cli_compare_release.py`, `cli_compare_release_helpers.py` and the 800-line caps on `release_scope.py` / `cli_compare_release_matrix.py`).
* Full unit lane — **40127 passed, 39 skipped, 4 xfailed, 0 failed**.
* `mypy abicheck/` clean; `ruff check abicheck/ tests/` clean; `check_ai_readiness.py` 0 errors.

## Tests

* `tests/test_package_component_inventory.py` (81 cases) — primitive-level property tests per AGENTS.md: `PackageInventory`/`merge_unproduced` order-independence and declared-minus-produced stated against an oracle independent of the implementation; `package_component_inventory` through the **real** `TarExtractor`/`WheelExtractor`/`DirExtractor` on real containers with genuinely ELF-shaped members (a filename-only fixture would pass for the wrong reason); the support-promise safety property checked over the **whole cross-product** of the policy domain and every `AcquisitionState`; `release_scope_warnings` over every state.
* `tests/test_release_package_inventory_cli.py` (9 cases) — through the public `compare` CLI and the rendered JSON/stderr, contrasting an archive pair against a directory pair built from *identical* snapshots, so the only variable is the operand container.

## Checklist

- [x] Changelog fragment added (`changelog.d/1788100000_claude_adr065-s3-s4-package-inventory.md`)
- [x] `scripts/verify.py --profile pr` run locally (see Verification for the one pre-existing failure)

https://claude.ai/code/session_01Ncsmks7L81RdnenXDXWQCu

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ncsmks7L81RdnenXDXWQCu)_


napetrov:
`test-action (required)` (in `.github/workflows/ci.yml`) failed with a timeout: it polls for the separate `test-action summary` check (from `.github/workflows/test-action.yml`) to complete, and that check hadn't reported after 35 minutes.

This isn't a code failure — every sub-job of the `Test GitHub Action` matrix that has run so far on this PR's head (`7f3d9b8c`) has passed; the remaining ones were still sitting in GitHub's runner queue (shared runner-pool congestion, visible from several unrelated jobs on other branches queued at the same time), not failing. I'll re-run `test-action (required)` once the matrix run finishes.

---

PR #1125, opened 2026-09-07
fix: generalize cross_source_evolution identity, migrate private_header_leak onto it

napetrov:
## Summary

While this PR was open, `main` independently landed **two** distinct ADR-068 D3 evolution-state mechanisms:

- `checker_policy.FindingEvolution` (#1122/#1124/#1126): a finding's identity across a *chain* of separate `compare()` calls over time (longitudinal history).
- `checker_policy.CrossSourceEvolution` (#1115): how a cross-source hygiene check behaves OLD→NEW *within one* `compare()` call, opt-in via `compare(..., cross_source_checks=True)`, with `unversioned_exported_symbol` as its first migrated check.

This PR originally reused `FindingEvolution` for `private_header_leak`'s same-comparison evolution — the wrong axis, since `private_header_leak`'s behavior is exactly what `CrossSourceEvolution` was built for. This revision reconciles with `main` instead of duplicating it:

- Drops this branch's now-redundant `compare/finding_evolution.py` and `workflows/crosscheck_evolution.py` in favor of `main`'s already-reviewed `CrossSourceEvolution` mechanism.
- Migrates `private_header_leak` onto `workflows/cross_source_evolution.py` as its **second** check, alongside `unversioned_exported_symbol`.
- **Fixes a latent bug** in the newly-merged mechanism: `_run_one_side` keyed findings by a bare `Change.symbol`, which silently collides two distinct findings sharing one symbol. `private_header_leak` can emit more than one finding per symbol (one function leaking two distinct private types), so this generalizes the module to a per-check identity function (`_IDENTITY_FUNCS`), defaulting to `symbol` and registering `(symbol, new_value)` for `private_header_leak`.
- `not_evaluated` stays mandatory: a pre-existing leak invisible on a header-less baseline never reads as a manufactured `introduced`.
- Authority unchanged: `private_header_leak` stays `RISK`, never promoted to `BREAKING`, regardless of evolution state.
- Reverts this branch's `tests/parity` additions that assumed a migrated check drops its `gaps.py` row — `main`'s actual convention (since the migration is opt-in, not wired into `compare`'s default pipeline) keeps that row, since the scan-vs-compare parity harness only exercises default invocation.
- Corrects the schema-history addendum (no report schema shape change — `private_header_leak` reuses the already-published `cross_source_evolution` wire shape) and the changelog fragment to describe `CrossSourceEvolution` instead of the discarded `FindingEvolution`-based design.
- Raises `architecture/debt.yaml`'s `no_growth` baseline for `abicheck/schemas/__init__.py` (1250 → 1258) for the corrected history addendum.

## Testing

- `tests/test_cross_source_evolution.py` — extended with a full `private_header_leak` matrix (evidence-state × finding-state), an F8-style pre-existing-leak-reads-not_evaluated test, an F9-style fixed-reads-resolved test, an authority-unchanged test, an opt-in-wiring test, and a dedicated identity test proving two distinct leaks on one symbol are no longer collided.
- `tests/parity/*` — reverted to `main`'s versions (no change in premise needed).
- `ruff check`, `ruff format --check` (this PR's own files), `mypy abicheck/`, `python scripts/check_architecture.py`, `python scripts/check_ai_readiness.py`, and the full fast unit suite (`pytest tests/ -m "not integration and not libabigail and not abicc and not slow and not golden"`, 40602 passed) all clean.

## Bug-fix test contract

- Bug class: `matching.dedup_key_soundness` (`tests/regressions/manifest.py`) — injectivity for semantically distinct findings. `workflows/cross_source_evolution.py`'s per-side OLD/NEW lookup keyed a `Change` by a bare `symbol`, which is non-injective for `private_header_leak` (`crosscheck.py` keys that check's own findings by `(mangled_or_name, leaked_type)`), so one function leaking two distinct private types silently collided into one `Change`.
- Publicly observable failure: `compare(old, new, cross_source_checks=True)` against a library where one function leaks two distinct private types would drop one of the two leaks from the reported `changes`/`cross_source_evolution` output, and if the two leaks had different evolution states (e.g. one resolved, one newly introduced) the surviving `Change`'s reported state could be wrong for the dropped one.
- Regression test fails on base: yes — `test_private_header_leak_identity_distinguishes_two_leaks_on_one_symbol` (`tests/test_cross_source_evolution.py`) fails against the pre-fix `_run_one_side` (bare `c.symbol` key): the `OldImpl`/`NewImpl` leaks on the shared `_Z3usev` symbol collide into one dict entry instead of two.
- Negative control: `unversioned_exported_symbol` (uniquely keyed by symbol already) keeps the default identity function unchanged; its own evolution-matrix assertions in the same file are unaffected by the per-check identity generalization.
- Public-surface test: `test_private_header_leak_wired_into_compare_opt_in` goes through the real `compare(old, new, cross_source_checks=True)` entry point, not only `compute_cross_source_evolution` directly.
- Axes covered: both migrated checks (`unversioned_exported_symbol`, `private_header_leak`); all evidence-state combinations per side (no headers / headers-clean / headers-with-leak) via `test_private_header_leak_evolution_matrix`'s parametrization.
- General invariant: `matching.dedup_key_soundness`'s injectivity clause, restated for `_run_one_side`'s per-check identity function: two `Change`s from the same check with different `new_value` never key-collide. `test_private_header_leak_identity_distinguishes_two_leaks_on_one_symbol` states this directly with an independently-chosen sibling case (two leaks, one symbol) against the oracle of `crosscheck.py`'s own per-check identity, not the fixed lookup itself. Added `tests/test_cross_source_evolution.py` to this class's `seed_tests` in `tests/regressions/manifest.py`.

## Related

ADR-068 D3, `docs/contribute/plans/one-comparison-product.md` §5 P2 / §3 #3-#4.

https://claude.ai/code/session_01UJjEZZgfzxfxeSQQcCfwyy


napetrov:
## codecov/patch (76.81%) — investigated, not a real coverage gap

The `codecov/patch` check flags low diff coverage on `abicheck/compare/finding_evolution.py` (46.15%) and `abicheck/report/finding_evolution.py` (77.77%). I traced this line-by-line against the uploaded report and it's a measurement artifact, not an untested code path:

- In `compare/finding_evolution.py`, codecov marks `evolve_check_findings`'s `def` line, its two dict-comprehension assignments (`old_by_id`/`new_by_id`), `results: list[Change] = []`, and the final `return results` as **misses** — while the loop body strictly *inside* that same function (`for key in sorted(...)`, the branch dispatch, `results.append(...)`) shows real hit counts (1–2). That's not physically possible under real execution: the loop can't iterate over `old_by_id`/`new_by_id` with real hits if the comprehensions that build them never ran.
- Same pattern in `report/finding_evolution.py`: `compute_change_evolution`'s body (`evolution = getattr(...)`, `return evolution.value if ...`) shows as a miss, while `render_change_evolution`'s `if value is not None: entry["evolution"] = value` shows real hits on both branches — which requires `compute_change_evolution` to have actually returned a non-`None` value at least once.

Locally, both files show **100% line+branch coverage** running the exact same tests (`tests/parity/test_evolution_state_gap.py::test_evolve_check_findings_exhaustive_matrix` exhaustively drives all 9 evidence combinations through `evolve_check_findings`, plus the real-snapshot matrix and end-to-end `compare()` wiring tests). The discrepancy only appears with CI's `COVERAGE_CORE=sysmon` env var; this repo's own `coverage.py` pin warns `Can't use core=sysmon: sys.monitoring can't measure branches in this version` and falls back — so the sysmon path is exercising an edge case in how hits get attributed/merged for these newly-added files under `-n auto --dist worksteal`, not a real gap in test coverage.

Per this repo's `codecov.yml`, `codecov/*` statuses are advisory (`require_ci_to_pass: false`) unless separately marked required in branch protection, so I'm not treating this as blocking. Flagging the finding here rather than adding vacuous tests to chase a reporting artifact. Happy to dig further if this is actually gating merge.

---

PR #1139, opened 2026-09-07
feat(cli): unconditional pattern-verdict modulation and audit-suppressions; repeatable --write

napetrov:
## Summary

ADR-068 D4 / `docs/contribute/plans/one-comparison-product.md` Phase 5:
pattern-verdict modulation and the suppression audit's underlying fact
become unconditional, presentation-only flags stop being able to change
analysis, and `--write` becomes repeatable on `compare`.

## Motivation

`one-comparison-product.md` §3/§4.1/§6 ("Product decisions and change
routing" -- AGENTS.md) states the rule this phase makes executable:
*presentation must never change analysis*. Before this PR, `compare OLD NEW
--explain-patterns` (with no `--pattern-verdicts`) silently turned
pattern-verdict modulation **on** as a side effect of `apply_patterns =
pattern_verdicts or explain_patterns` -- asking *why* a modulation happened
could also decide *whether* it happened, changing the reported verdict and
exit code on identical operands. That is a real correctness bug, not a
style issue: a user could get `BREAKING` without `--explain-patterns` and
`COMPATIBLE_WITH_RISK` with it.

## Changes

- **`--pattern-verdicts`/`--no-pattern-verdicts` removed.** Pattern-verdict
  modulation (opaque-pointer/PIMPL demotion, lost-invariant raises) now runs
  unconditionally on every `compare`, gated only by its own existing
  evidence requirement (header-aware-or-better tier for demotion) -- never
  by a flag. It never fires without genuine idiom evidence and never turns
  a break into a pass on its own, so making it unconditional never
  manufactures a false negative, only removes an opt-in a user could
  forget.
- **`--explain-patterns` is now pure rendering.** It only controls whether
  the idiom evidence behind an already-decided modulation is printed --
  never whether the modulation happened. **Behavior change**: a run that
  previously relied on `--explain-patterns` alone to enable modulation now
  gets modulation unconditionally (same or richer result, never a
  regression); a run that explicitly passed `--pattern-verdicts`/
  `--no-pattern-verdicts` must drop the flag -- the removed spelling now
  exits `64` with `No such option`.
- **`--surface-metrics` is deliberately left alone.** Unlike modulation, a
  surface-metric finding is a brand-new `Change` record merged directly
  into the result with no separate suppressible-but-hideable channel --
  making it unconditional would add a new, always-visible finding to most
  real comparisons, a larger behavior change than this phase budgeted for.
  It stays the pre-existing opt-in flag; folding it into AUTO is deferred
  to a follow-up slice (see `adr027_compare_options`'s docstring in
  `abicheck/frontends/cli/options/release.py`).
- **`--audit-suppressions`'s audit is now always computed** when
  `--suppress` is given, present in JSON/SARIF/JUnit/HTML regardless of the
  flag; the flag now only controls whether markdown/text/review *also*
  prints the human-readable section. Previously forgetting the flag
  silently withheld the fact from every output, machine-readable included.
- **`--write FORMAT=PATH` is repeatable on `compare`** -- pass it more than
  once (`--write json=a.json --write markdown=b.md`) to emit several
  artifacts from one already-computed comparison, each rendered full and
  unfiltered, demangled per its own format. `scan --against`'s `--write` is
  unchanged (still single-value). The directory/package release engine
  still accepts only one `--write` (a second one is a clear `UsageError`,
  not silent truncation).
- New `tests/test_presentation_analysis_separation.py`: F-19 (rendering
  permutations across `--format`/`--write`/demangle never change the
  canonical result or exit code) and F-20 (the specific --explain-patterns
  bug above is fixed and stays fixed).

**`compare`'s option count**: 86 -> 85 (`--pattern-verdicts` removed;
`secondary_fmt` becomes the repeatable `secondary_writes`, still one
param). This PR does **not** attempt the plan's larger `--view` collapse
(`--report-mode`/`--show-only`/`--demangle` folding into one `--view`
option, plan §4.1) -- that is out of scope for this phase and left for a
later PR in the sequence; nothing here blocks it.

Full `pr`-profile verification (lint, fmt-check pre-existing repo-wide drift
aside -- see note below, typecheck, ai-readiness, architecture) is clean;
40738 tests pass (0 failed) on the fast+coverage lane.

**Note on `ruff format --check`**: this environment's `ruff format --check
abicheck/ tests/` reports 603 pre-existing files needing reformatting
identically on the base commit (verified via `git stash`) -- an environment
drift unrelated to this change, not something this PR introduces or
attempts to fix wholesale.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GD2nKTWjYyh4zKeWLgPtCP)_


napetrov:
## "Full example matrix" CI failure — root cause and fix

**What actually flipped (and what didn't):** the failing check reported Runtime smoke drifting from `90 DEMONSTRATED / 78 NO_RUNTIME_SIGNAL` to `89/79`. The obvious suspect — this PR's `--pattern-verdicts` removal in `tests/validate_examples.py` — turned out **not** to be the cause: `run_example_runtime_smoke.py` doesn't build its abicheck compare command through `validate_examples.py` at all; it just builds each case's consumer app via CMake and runs it, so a CLI-flag change here can't touch its output.

Bisecting by re-running the whole runtime-smoke suite repeatedly at HEAD found the real culprit: **`case199_public_function_parameter_added`** flips between `DEMONSTRATED` and `NO_RUNTIME_SIGNAL` from run to run, on both this PR's head commit *and* the pre-PR base commit (`cfee969d`) — confirming it predates this PR entirely. The case's own README already documents why: the consumer calls `chan_open()` with one argument while `libv2`'s version now reads a second one out of whatever the calling convention leaves in that register — *"chan_open -> 3 or -1, depending on register residue"*. That residue is address-derived, so with ASLR enabled it varies run to run, independent of any code change.

**Fix:** `validation/scripts/run_example_runtime_smoke.py`'s baseline/swapped app invocations now clear ASLR for the child process (`personality(ADDR_NO_RANDOMIZE)` on Linux, best-effort no-op elsewhere) before executing. This is a general fix for the bug *class* (any runtime-smoke demo relying on real undefined behavior), not a patch to `case199` alone. Verified stable across many repeated full-catalog runs locally (previously flipped within 2-3 runs; now consistently `90/78/1/39`, matching the catalog's already-committed table).

**Also fixed:** `catalog/README.md`'s Release headers / Stripped headers rows were separately stale (`157/152 PASS`, `46 SKIP` vs. the real `164/158 PASS`, `39 SKIP` a fresh run produces) — confirmed pre-existing at the base commit too, unrelated to this PR. Refreshed via the documented `check_examples_validation_status_sync.py` (no `--check`) path, plus a narrative update naming the one additional (also pre-existing) stripped-headers reduced-evidence case, `case89_inline_accessor_renamed_pimpl_member`.

**Verification:** `check_examples_validation_status_sync.py --check` with the exact `--gcc/--clang/--runtime/--release/--stripped/--build-source/--full-matrix` flags CI uses now exits 0 (reproduced locally end-to-end: full gcc/clang/release-headers/stripped-headers/build-source `validate_examples.py` runs + `run_example_owner_proofs.py`/`run_bundle_examples.py`/`run_special_cli_examples.py`/`collect_full_example_matrix.py`, matching CI's `208/208 COVERED`). `pytest tests/test_validate_examples_unit.py`, `ruff check`, and `python scripts/check_ai_readiness.py` all pass on the changed files.

Commit: `ee73968f`

---

PR #1140, opened 2026-09-07
fix: Mach-O mangled-name identity normalization and PR #1138 CodeRabbit follow-ups

napetrov:
## Summary

Follow-up to #1138 (already merged), which shipped with two classes of known, unaddressed defects: a real cross-platform correctness bug caught by CI's macOS `integration-tests` job (still red on `main`'s `13eb153b`), and three CodeRabbit review findings on that PR that landed before its fixes could be pushed. This PR closes all four, based on current `main`, plus two further defects CI's macOS runner and its "Full example matrix" job surfaced during review of this PR itself.

## Fixes

1. **Mach-O leading-underscore identity normalization (the macOS CI failure), three independent layers deep.** Darwin's Itanium mangling convention decorates every real mangled name with an extra leading underscore (`_Z...` → `__Z...`), and even a plain C-linkage export gets one leading underscore. Three separate, independent bugs all needed fixing before self-comparing an ordinary Mach-O library stopped falsely reporting a risk:
   - `dumper_clang.py`/`extract/headers/clang/functions.py`: a bare `--ast-frontend clang` dump left `Function.mangled`/`Variable.mangled` decorated for genuine C++ (Itanium) declarations — only the castxml+clang hybrid-merge path stripped this. Fixed via a new shared `extract.headers.clang.context.strip_darwin_itanium_decoration` helper.
   - The same path left a genuine `extern "C"`/plain-C declaration's Darwin-decorated name (`_c_func`) unstripped even though `is_extern_c` detection already correctly recognized it. Fixed by reusing that exact already-computed boolean to gate stripping a single leading underscore only when `mangled == "_" + name` exactly.
   - `abicheck/model/export_index.py`: `default_versioned_names`'s Mach-O branch **re-stripped** a Mach-O export's leading underscore a second time, even though `macho_metadata.py` already strips it once when parsing the real export trie/symtab — corrupting every Itanium C++ export (`_ZN2ns3fooEv` → `ZN2ns3fooEv`) and breaking correlation against the by-then-correctly-normalized declared identity. This bug predates this whole PR chain (latent, invisible only because `Function.mangled` used to be doubly-stricken too) — fixed by making the Mach-O branch an unconditional passthrough, matching the already-correct `linked_export_names`.
2. **`_cache_key()` didn't hash `force_cpp`** (only `force_cpp20`) — a plain header could cache a C-mode AST, then a later request correctly wanting C++ (via #1138's export-evidence fallback) would silently hit the stale C-mode cache entry. Fixed.
3. **Export-evidence C++ fallback wasn't header-scoped**, in two ways: it used the whole binary's export-symbol union rather than correlating with the specific header being parsed (an unrelated C++ export elsewhere in the binary could wrongly force `-x c++` on a genuinely plain C header — CodeRabbit's example: `struct options { int new; };` is valid C, invalid C++), and its raw-token collection didn't exclude comments/string literals/inactive `#if 0` blocks (a name only appearing in a comment could falsely correlate). Both fixed: real header-scoped correlation reusing the existing comment/string/`#if 0` stripper, gated on Itanium/MSVC mangling shape.
4. **`scope.public_header_dirs` (#1138's P4 config boundary) didn't reach the directory/package release fan-out or its stranded-library fallback.** Threaded through `service_compare_pipeline.run_compare`, `cli_compare_release_pairwise.py`, and `cli_compare_release.py`'s `_resolve_stranded_library`, matching the direct two-file compare path #1138 already wired.
5. **CI's "Full example matrix" job's `Artifact modes (release-headers)`/`Artifact modes (stripped-headers)` steps silently skip 7 `source_smoke`-run-mode example cases** — investigated as a possible regression from this PR, confirmed instead to be a pre-existing `.github/workflows/examples-validation.yml` configuration gap: those two jobs never set `ABICHECK_TRUSTED_SOURCE_SMOKE_RUN=1`, unlike the sharded debug-headers job that already does. Proven via a clean `origin/main` worktree run reproducing byte-identical (buggy) numbers before this PR's changes. Fixed the workflow to set the same env var on both jobs. `catalog/README.md`'s validation-status table needed no changes — `check_examples_validation_status_sync.py --check` is clean once the workflow fix is applied.

## Bug-fix test contract

- **Bug class:** new `extraction.macho_mangled_identity_normalization` (`tests/regressions/manifest.py`), with an explicit, honest `KnownGap` noting no Mach-O toolchain is available in this sandbox to compile a real end-to-end reproduction — verified instead by code-level correctness, targeted unit tests of each normalization helper, and (for the final export-index layer) CI's own real macOS runner across three consecutive rounds. Also extends the existing `extraction.language_mode_export_evidence` invariant for the header-scoping fix.
- **Publicly observable failure:** on Darwin, `abicheck compare` (or `scan`) on any library exposing a plain C++ function, a plain C-linkage function, or built via the header-AST+export-index pipeline reported a self-contradictory `exported_not_public`/`public_not_exported` pair for an unchanged declaration — exactly mirroring the Linux bug #1138 fixed for ELF. CI's `integration-tests (macos-latest)` job on this PR is the direct evidence, across three consecutive real-macOS-runner failures as each layer was found and fixed in turn.
- **Regression test fails on base:** yes for every fix — `tests/test_dumper_hybrid_macho_idempotence.py`, `tests/test_dumper_clang_extern_c_identity.py`, `tests/test_crosscheck_macho_export_index_normalization.py`, extensions to `tests/test_dumper_language_mode_export_evidence.py`/`tests/test_crosscheck.py`/`tests/test_export_index.py`, and `tests/test_compare_config_public_header_dirs.py`'s new release-path assertions all fail against `main`'s current tip; `test_public_not_exported_normalizes_macho_underscore` was actively asserting the buggy double-strip behavior and is corrected in place.
- **Negative control:** the correlation fixes include positive/negative pairs throughout (a genuine correlated C++ export forces C++ mode, an unrelated export or a comment/string-only occurrence does not; a real C-linkage symbol's decoration is stripped, a name that merely starts with `_` without matching the exact Darwin-decoration shape is untouched).
- **Public-surface test:** yes — real `compare`/`dump` CLI paths and the release fan-out CLI path.
- **Real-dependency test:** yes — `tests/test_dumper_clang_extern_c_identity.py`, the extended `tests/test_dumper_language_mode_export_evidence.py`/`test_crosscheck_language_mode_export_evidence.py` cases, and `tests/test_source_extractors_clang.py` exercise the real `clang`-backed extraction path end to end, not a hand-constructed shortcut into a lower-level API; the Mach-O-specific normalization pieces this sandbox cannot compile for are covered by direct unit tests of each shared helper plus CI's own real macOS runner, which is what actually proved each of the three layers fixed, in turn, across this PR's review cycle.
- **Malicious fixture + side-effect absence:** this diff touches `.github/workflows/examples-validation.yml`, but only to add one existing env var (`ABICHECK_TRUSTED_SOURCE_SMOKE_RUN=1`) to two jobs, copying the exact value and justification an adjacent job in the same file already uses — no new script, action, or logic that processes untrusted input is introduced, so there is no new attack surface for a malicious-fixture test to exercise. `check_examples_validation_status_sync.py --check` and a real local run of both affected jobs' commands (`validate_examples.py --artifact-variant release-headers`/`stripped-headers`) confirm the change's only effect is the previously-silently-skipped 7 cases now running, with no change to any other job's behavior.
- **Axes covered:** frontend (clang direct path, castxml+clang hybrid path) × platform mangling shape (ELF Itanium, Mach-O-decorated Itanium, Mach-O-decorated C-linkage) × correlation scope (header-local vs whole-binary, active-declaration-text vs raw comment/string tokens) × compare topology (direct two-file, directory/package release fan-out, stranded-library fallback) × pipeline layer (header-AST extraction, export-index normalization).
- **General invariant:** a declaration's identity/mangled-name normalization must be idempotent and applied consistently at *every* extraction/index path that produces or consumes it — this PR found three separate, independently-latent violations of that single invariant across the header-AST layer and the binary export-index layer, which is why each fix was verified in isolation and the whole chain re-verified together; C++-mode inference from export evidence must correlate with the specific header being parsed using only its active declaration text, never the whole binary's export set or incidental comment/string tokens; a cache key must hash every input that affects its cached value; a project-level scope/boundary config value must reach every compare topology that resolves an `InputSpec`, not just the simplest one; and a CI job's local reproduction must match the job's exact environment (including env vars) before a discrepancy is trusted as a real regression rather than a reproduction gap.

## Verification

- Full non-integration suite: 40,932 passed, 0 failed (one pre-existing NaN-equality flake in `test_dedup_key.py`, confirmed unrelated via stash-and-rerun against clean `main`).
- Full `integration`-marked suite (real castxml + gcc): 533 passed, 0 failed.
- CI's real macOS runner (`integration-tests (macos-latest)`): red on three consecutive commits as each of the three independent Mach-O normalization bugs was found and fixed in turn; this PR's latest commit is expected to be the first to pass it cleanly.
- `mypy abicheck/`, `ruff check`, `check_architecture.py`, `check_ai_readiness.py`: all clean.
- `docs/reference/python-api-reference.md` regenerated; `test_service_unit.py`'s positional-argument pin updated for `run_compare`'s new trailing parameter.
- `architecture/debt.yaml`: reviewed, justified adoption-baseline bumps for genuine necessary growth; every other touched file stayed under its existing cap.

## Checklist

- [x] Docs updated (`python-api-reference.md` regenerated)
- [x] Full non-integration and integration suites pass locally

## Summary by CodeRabbit

* **New Features**
  * Project-configured public header directories are now honored throughout release and package comparisons, including stranded-library checks.

* **Bug Fixes**
  * Improved Darwin symbol normalization for C++, C, and `extern "C"` declarations.
  * Prevented incorrect Mach-O export-name rewriting during comparison.
  * Improved C/C++ parsing detection by correlating exports with active declarations in the relevant headers.
  * Prevented AST cache results from being reused across incompatible language modes.

* **Documentation**
  * Updated the Python API reference and changelog with the new comparison option and fixes.


napetrov:
`unit-tests (windows-latest, 3.13, false)` is red and failed identically on re-run: `test_presentation_analysis_separation.py::TestF19OutputFormatInvariance::test_write_is_repeatable_and_does_not_change_the_primary_result` raises `UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d` when reading back a written Markdown report on Windows. This is unrelated to this PR — the diff here touches only header-AST extraction (`dumper_clang.py`, `dumper_ast_config.py`, `extract/headers/clang/`), the Mach-O export index (`model/export_index.py`), release/compare-pipeline config threading, and one CI workflow env var; nothing in `--write`/report-encoding is touched. No fix for this pre-existing Windows-encoding gap exists yet in this codebase that I'm aware of. Every other check — `integration-tests` on macOS/Ubuntu/Windows, `abicc-parity`, `libabigail-parity`, `Full example matrix`, `lint-and-types`, `ai-readiness`, `test-contract`, and `unit-tests` on macOS/Ubuntu (3.12/3.13/3.14) — is green on this PR's current head.

---

PR #1173, opened 2026-09-09
refactor: rename Phase 6 scan-survivor engine primitives off the scan identity

napetrov:
## Summary

Phase 6 step 4 of `docs/contribute/plans/one-comparison-product.md`, pulled forward so it can land before `scan` itself is deleted. Pure rename/re-home plus a documentation truth sweep — no behavior change anywhere.

## Motivation

`scan`'s eventual retirement (ADR-068) needs the survivor engine primitives it shares with `compare` (§3 #34) renamed off the `scan` identity ahead of time, and this pass also surfaced status prose that no longer matched the code.

## Changes

**Part A — rename the survivors off the `scan` identity.** The plan named six files; the tree actually has eight in the family, so I verified each one's real callers via an AST scan (`tests/parity/test_engine_primitive_call_sites.py`) rather than trusting the plan's prose:

- `abicheck/buildsource/{crosscheck,crosscheck_base,crosscheck_coherence}.py` → `cross_source_checks{,_base,_coherence}.py`. Stay physically in `buildsource/`, not moved to `abicheck/compare/`: the main engine file imports `export_accounting.py` (`extract`-classified) and `source_graph_query.py` (unclassified), which `compare/`'s `may_import: [model]` restriction forbids without reclassifying those two modules first — out of scope for a pure rename. (I did attempt the physical move first; `scripts/check_architecture.py` caught the real layering violation, so I reverted to a rename-in-place.)
- `abicheck/buildsource/{pattern_scan,preprocessor_scan}.py` → `pattern_facts.py`/`preprocessor_facts.py` (stay in `buildsource/`, the L3-L5 evidence package). `scan_files`/`PatternScanResult` → `find_pattern_facts`/`PatternFactsResult`; `run_preprocessor_scan`/`PreprocessorScanResult` → `collect_preprocessor_facts`/`PreprocessorFactsResult`.
- `abicheck/buildsource/scan_levels.py` → `abicheck/model/evidence_depth_levels.py` (a value-type vocabulary with no buildsource-specific dependency, consumed at least as widely by `compare` as by `scan`).
- **Correction to the plan's own Phase 6 prose:** `poi.py` and `risk.py` were also named there as surviving engine primitives, but the same AST call-site audit found neither has a `compare`-pipeline caller today — every caller is scan-only. They are **not** renamed here; the plan is corrected to reclassify them as delete-with-`scan` candidates pending a genuine `compare`-side caller.
- Every caller and test updated; `tests/parity/test_engine_primitive_call_sites.py` updated (not relaxed) for the new paths/symbols and fixture shapes.
- `scripts/check_architecture.py` taught to recognize a git-detected rename when validating the no-growth debt ledger — previously any `git mv` of a debt-tracked file read as "brand-new file" against the PR base, which would make a legitimate rename like this one permanently fail that gate.

**Part B — record vs. code.**

- `one-comparison-product.md`'s Phase 4 section rewritten to match what's actually landed: the Action `mode:scan`→`mode:compare` translation was built (`7f8109a1`) then unconditionally reverted the same day (`9f2166e5c`) because `compare` lacks `scan`'s advisory-only cross-source-finding stripping for a baseline comparison and `compare --no-baseline` still crashes; the typed-API absorption and docs sweep are far short of done. (Another PR is working the code side of this gap — this only corrects the written record.)
- Fixed three docs pages that recommended `scan --against ...` for an ordinary baseline comparison when `compare` already owns that case per every other already-migrated page: `docs/start/getting-started.md`, `docs/start/real-world-example.md`, `docs/learn/where-in-the-pipeline.md`.
- ADR-049's stale `**Verified:**` receipt fixed: the prior sha was a branch-local commit not reachable from `origin/main` (the exact failure mode `adr-status-sync` flags) — re-verified the Status paragraph against the seven commits that touched its named modules since (all additive, none contradicts a claim there) and recorded a real `origin/main` commit.

**Part C — deletion-order checklist** appended to the plan's Phase 6 section: verified file/line counts for the scan-only `abicheck/` modules (16 files, 10,074 lines — the plan's "~10,000 lines" held up almost exactly) and scan-only tests (33 files, 21,111 lines — corrected from "40 test modules", which over-counted by including renamed-primitive tests and unrelated false positives), the `run.sh` branch count (17 explicit conditionals, not "~40"), one file needing a split before deletion (`tests/test_preprocessor_scan.py`), and the naming-sweep false positives. **Nothing is deleted in this PR** — the `scan` deletion semantics blocker is still open.

## Verification

- Full `abicheck` package import-walks cleanly (0 errors across every submodule)
- `ruff check`/`ruff format --check` clean on every file this PR touches
- `mypy abicheck/`: 0 errors, 720 files
- `scripts/check_architecture.py`: 0 errors
- `scripts/check_ai_readiness.py`: 0 errors (down from 1 before this PR — the ADR-049 receipt fix)
- `python scripts/verify.py --profile pr`: 19/21 checks passed. The 2 failures are unrelated to this PR: `fmt-check` reflects pre-existing repo-wide `ruff format` drift (628 files, reproduced identically against `origin/main` before any of my edits) that this PR does not touch further; `docs-build`'s one real finding (a broken anchor from a heading reword in this PR's own docs sweep) is fixed in the second commit.

## Files avoided per the task's parallel-work note

`abicheck/checker.py`, `abicheck/policy/**`, `action/run.sh`, and compare's/dump's Click option definitions are untouched. Two named files (`abicheck/cli_scan_baseline.py`, `abicheck/service_scan.py`) needed a 2-line mechanical import-path fix each — unavoidable, since they import the renamed modules by name and the package would not import without it (verified by the whole-package import-walk above). Several `tests/parity/**` files needed the same kind of import-path fix, plus the explicitly-required update to `test_engine_primitive_call_sites.py` itself.

https://claude.ai/code/session_012u1UCYgNcEutc6JdTYobGp

---
_Generated by [Claude Code](https://claude.ai/code/session_012u1UCYgNcEutc6JdTYobGp)_


napetrov:
Pushed `d6c83cdb` addressing all three Codex review findings:

- **P1** (`docs/start/real-world-example.md`): unprefixed `--sources .` applies to both compare operands, so it was silently requesting source evidence for the stored, headers-only baseline snapshot too. Fixed to `--sources new=.` there and in the analogous command in `docs/learn/where-in-the-pipeline.md`.
- **P2** (`scripts/check_architecture.py`): the git-rename-detection fix this PR added trusted any git-detected rename (default 50% similarity) as evidence of an established-debt file being renamed. Now also requires the renamed-from path to have genuinely been listed in `architecture/debt.yaml` at the base revision before trusting it — closing the spoofing path Codex described (rename an ordinary file, grow it while keeping similarity, add it to the ledger at its new size).
- **P2** (`docs/contribute/plans/one-comparison-product.md`): two references paired an already-renamed module path with its pre-rename function name (`pattern_facts.scan_files`, `preprocessor_facts.run_preprocessor_scan`) — fixed to the real current names (`find_pattern_facts`, `collect_preprocessor_facts`).

Re-verified: `scripts/check_architecture.py`'s own 53-test suite passes, mypy/ruff clean, `mkdocs build --strict` clean.

**CI: `unit-tests (macos-latest, 3.13, false)` failure.** Not this PR's — the identical failure (`tests/test_action_release_topology_config.py::TestReleaseTopologyOverlayCleansUpOnExit::test_release_topology_overlay_is_removed_after_the_trap_fires`, `CMD[-1]: unbound variable` / "bad array subscript") is present on `main`'s own latest push at this PR's exact base commit (`3930cfce`, [run 34315674516](https://github.com/abicheck/abicheck/actions/runs/34315674516)) — before this PR existed. Root cause: the test's embedded bash fixture uses `${CMD[-1]}` (negative array indexing, bash ≥4.3), which fails on macOS's system bash (3.2, Apple's GPLv3 freeze) with "bad array subscript." This PR touches neither that test file nor `action/run.sh`. No fix exists yet on `main`; a portable one-line patch would be:

```diff
-_overlay_path="${CMD[-1]}"
+_overlay_path="${CMD[$((${#CMD[@]} - 1))]}"
```

Not pushing it here since it's unrelated code outside this PR's scope. I'll re-run the check once as the one permitted re-run, though it's not expected to help (deterministic, not a flake).

---

napetrov:
Two more CI signals on `120b6b97`, both unrelated to this PR's changes:

**`unit-tests (windows-latest, 3.13, false)`** — 2 failures in `tests/test_action_release_topology_config.py` (`test_same_drive_letter_shaped_path_is_prefixed_on_non_windows`, `test_ordinary_relative_path_still_gets_pwd_prefix_on_windows`): the test expects a Windows-style `C:\Users\...` path but got an MSYS-style `/c/Users/...` path, i.e. the Windows runner's shell environment produced a Git-Bash-mangled path instead of a native one. `tests/test_action_release_topology_config.py` and `action/run.sh` are byte-identical between this PR's base commit (`3930cfce`) and this branch's HEAD (`git diff 3930cfce -- tests/test_action_release_topology_config.py action/run.sh` is empty) — this PR touches neither file, so the failure is pre-existing runner-environment behavior, not something this branch introduced.

**`Baseline regression (PR vs base)`** — flagged `opaque_filter @ size=1000` (+31%, 0.722s→0.949s) and `nested_types @ size=500` (+19%, 0.643s→0.766s) against the gate's `max(15%, 0.100s)` tolerance. This PR makes no changes to `diff_types.py`/`checker.py`/the comparison engine itself — only renames and docstring/import-path updates in `buildsource/`/`model/`, none of which are on either scenario's hot path. Given the shared CI environment's very high concurrent load, this reads as measurement noise from a busy runner rather than a real regression.

Re-ran both as the one permitted re-run for each; the windows job's re-run was already in flight (queued together with the earlier macOS re-run). Will follow up if either recurs.

---

napetrov:
**`mutmut (detector core)`** — `mutmut run` timed out after 20700s (345 min, the script's own internal ceiling, just short of the job's 360-min platform-max wall clock) without producing a result.

Why this lane ran in full rather than diff-scoped: this PR touches no `only_mutate` production module (`diff_symbols.py`/`diff_types.py`/etc.), but it does touch `tests/test_surface_graph.py` (an unavoidable `crosscheck`→`cross_source_checks` import-path fix, matching the workflow's `mutated_tests` path filter for that module), which sets `require_baseline=true` — a diff-scoped-only gate isn't enough once a mutated module's own test changed, since there may be no production function in scope to score against. `check_mutation_score.py`'s own diff-scoping safety check then disables `--scope-run-to-diff` because the diff also touches many files outside `only_mutate` (the required `changelog.d/` fragment, `docs/`, and every other renamed/updated file in this 144-file PR) — by design, per that script's documented "no allowlist, fails closed to a full run" rule. That forces the full, unscoped `only_mutate` suite, which this workflow's own comments note has previously timed out at smaller budgets (240min on the 2026-08-31 weekly run, before the ceiling was raised to the current 360min platform max) as the suite has grown.

No detector-quality regression to root-cause here — this PR changes zero lines in any `only_mutate` module, so there is nothing for a surviving mutant to indict. This is a pre-existing scoping/timeout limitation in the mutation lane's fallback path (any large, broad-touching PR that also touches a mutated module's own test hits it), not a defect this PR introduced. Re-running once, since the run got within ~5% of its own budget and a faster iteration may complete.

---

PR #1174, opened 2026-09-09
Phase 7 CLI cleanup: 7d remainder + 7g resource-limit calibration

napetrov:
## Summary

Closes out Phase 7d's remainder and Phase 7g of `docs/contribute/plans/one-comparison-product.md`'s CLI/config cleanup. `compare`'s accepted-option count: **62 → 59**.

## Motivation

Four `compare` flags survived the release-topology demotion (`1e9d59698`) without an explicit ADR-068 D5 ruling, and `--max-json-object-nodes` needed a real calibration before it could move to config (per this workstream's task instructions: measure, don't pick a number and call it calibrated).

## Changes

- **`--keep-extracted`/`--no-bundle-analysis` removed outright, no config replacement.** Neither survives D5's three guards — a local-debug retention knob with no per-run evidence content, and the "escape hatch that disables real analysis" D4 rules out. Extraction cleanup on a directory/package `compare` is now unconditional; bundle-level cross-library analysis always runs. The Action's `keep-extracted` input is removed too, front-end parity in the same PR.
- **`--bundle-facts-out`/`--bundle-facts-library-manifest` affirmed as KEEP**, ruled explicitly rather than left "pending" — a per-run operand (same shape as `-o/--output`) and a document-naming CLI flag (same class as `--policy`/`--suppress`), respectively. See their own docstrings in `frontends/cli/options/release.py`/`bundle_facts.py`.
- **`--max-json-object-nodes` replaced by `.abicheck.yml`'s `resource_limits.max_bundle_facts_decode_nodes`.** Did the real calibration first: measured actual bytes-per-node density (~9.3, stable across scale) against a synthetic corpus sized like the flag's own named use case (oneDAL-scale, via `scripts/benchmark_scaling._build_onedal_large_surface`) — the previous `DEFAULT_MAX_JSON_OBJECT_NODES=1_000_000` was already ~6x below a real single-library, 25k-function facts blob's own 5.8M nodes. Recalibrated to `20_000_000`. Deliberately kept node-based rather than re-expressed as a memory size (the superseded `cli-cleanup-phase-two.md`'s own proposal): converting a memory budget to a node budget via this measured *legitimate*-payload ratio would size the node budget for an adversarial payload's much lower bytes/node density too, weakening the exact container-count defense `storage.json_budget` provides. Full measurement table: `bundle_facts.py`'s own docstring / the plan doc's Phase 7g entry.
- Regenerated `docs/reference/{cli-reference,config-keys-reference,github-action-inputs}.md`. Added a `resource_limits:` section to `docs/reference/config-file.md`.
- Registered both retired-surface families in `scripts/check_docs_contract.py`'s `_RETIRED_SURFACES`.
- Raised `architecture/debt.yaml`'s `no_growth` baselines for the five files this change's minimal, irreducible new-field/test cost landed in (`build_config.py`, `cli_helpers_compare.py`, `test_bundle.py`, `test_package.py`), each with a recorded rationale following the file's own existing raise-baseline precedent. `cli_compare_release_helpers.py` and `compare_bundle_facts.py` stayed within their existing baselines after trimming.
- Changelog fragment added.

## Checklist

- [x] Docs updated (CLI/config/Action reference regenerated, `config-file.md`/`github-action.md`/`multi-binary.md` updated, plan doc's Phase 7 section updated)
- [x] `mypy abicheck/` and `ruff check abicheck/ tests/` clean
- [x] `python scripts/check_architecture.py` clean (0 findings)
- [x] `python scripts/check_ai_readiness.py` / `python scripts/check_docs_contract.py` clean (only pre-existing, unrelated baseline findings remain — verified via `git stash` diff before/after this change)

Not a `fix:`/`perf:`/`security:` PR, so the bug-fix test contract section is omitted per the template's own instruction.

https://claude.ai/code/session_015uRDuV6c4wRV1cp7RZatZE

---
_Generated by [Claude Code](https://claude.ai/code/session_015uRDuV6c4wRV1cp7RZatZE)_

## Summary by CodeRabbit

- **New Features**
  - Added `.abicheck.yml` support for configuring BundleFacts JSON decode limits with `resource_limits.max_bundle_facts_decode_nodes`.
  - The setting is applied when supplied through an explicit `--config`.

- **Changes**
  - Bundle analysis now runs automatically for comparisons.
  - Temporary extracted files are always cleaned up.
  - The default decode limit remains conservatively set to 1,000,000.

- **Removed Options**
  - Removed `--keep-extracted`, `--no-bundle-analysis`, and `--max-json-object-nodes`.
  - Removed the GitHub Action’s `keep-extracted` input.

- **Documentation**
  - Updated CLI, configuration, and GitHub Action references.


napetrov:
Pushed `6e1d816c` addressing:

**CI failure — "Ubuntu package and bundle smoke"**: `.github/workflows/realworld-validation.yml` still passed `--no-bundle-analysis` to two live `compare` invocations. Removed both (the flag is gone per this PR) and updated the JSON-contract assertion from `"bundle_findings" not in data` to `data["bundle_findings"] == []`, since bundle analysis always runs now.

**CodeRabbit findings** (both addressed):
- `docs/reference/config-keys-reference.md` documented `max_bundle_facts_decode_nodes` as `unspecified` — `gen_config_reference.py` didn't know about `INT_SUBKEYS`. Taught it, regenerated; now documents as `int`.
- `tests/test_bundle.py`'s new `test_compare_release_bundle_analysis_always_runs` now asserts `bundle_findings == []`, not just that the key exists.

**CI failure — "unit-tests (macos-latest, 3.13, false)"**: `TestReleaseTopologyOverlayCleansUpOnExit::test_release_topology_overlay_is_removed_after_the_trap_fires` (a bash-array-manipulation test in `add_release_topology_config_flags`'s own harness, unrelated to anything this PR touches). Confirmed pre-existing: it fails identically on `main` at this PR's exact base commit (`3930cfce`, [CI run 34315674516](https://github.com/abicheck/abicheck/actions/runs/34315674516/job/102351260102)) — same test, same `CMD: bad array subscript` / `CMD[-1]: unbound variable` error. No fix exists yet upstream; not attempting one here since it's outside this PR's scope (CLI/config cleanup) and would widen the diff into unrelated `add_release_topology_config_flags` logic. Will keep watching.

---

napetrov:
Pushed `52869dcf` addressing both new Codex findings:

**P1 (`bundle_facts.py:121`) — "Keep the conservative decode budget as the default"**: agreed, this was a real regression. Recalibrating `DEFAULT_MAX_JSON_OBJECT_NODES` up to `20_000_000` to match the measured oneDAL-scale need also raised the decode-bomb ceiling for every *unconfigured, untrusted* `BundleFacts` blob by the same 20x (~75 MB → ~1.5 GB worst-case RSS). Reverted the default to the original `1_000_000`. The measurement itself stays (that's what justifies the new `resource_limits.max_bundle_facts_decode_nodes` config key existing) — only the fix is corrected: a project with a real large-blob need sets that key explicitly in its own `.abicheck.yml`, rather than the ambient default moving for everyone. Updated the plan doc, `config-file.md`, and the changelog fragment to match.

**P2 (`docs/use/multi-binary.md:514`) — "Preserve the documented omitted-key failure case"**: also correct — `_run_bundle_analysis()` can still return `None` (no libraries matched, or the bundle-level snapshot build itself failed), in which case `_format_release_json()` omits both `bundle_verdict`/`bundle_findings`. Fixed the wording to name those two pre-existing, unrelated cases instead of claiming an unconditional guarantee.

---

napetrov:
Agreed — this was a real gap. Fixed in 9cb349d8: `compare_bundle_facts.dispatch()` now takes a `config_explicit` parameter, threaded through unchanged from `compare_cmd`'s own `ctx.get_parameter_source("config") == COMMANDLINE` check (the same one already computed there for `resolve_stored_bundle_lang`). `resource_limits.max_bundle_facts_decode_nodes` is now only honored when `config_explicit` is true; an auto-discovered `.abicheck.yml` falls back to the conservative `DEFAULT_MAX_JSON_OBJECT_NODES`, the same way `resolve_dispatch_compile_context()` already treats auto-discovered config as untrusted for `compile:`.

Added `test_max_json_object_nodes_override_ignored_when_config_auto_discovered` (next to the existing explicit-`--config` test) asserting the auto-discovered budget-of-1 is *not* honored.

---

napetrov:
Agreed — narrowing the ceiling is never a decode-bomb risk, only raising it is. Fixed in 90445a3b: extracted the resolution rule into `resolve_max_json_object_nodes_cfg()` (`compare_bundle_facts_rejections.py`) which now applies `min(configured, default)` for an auto-discovered config while letting an explicit `--config` raise it freely. Direct unit tests cover both directions plus the no-configured-value case, and the sibling CLI test that (incorrectly, per the prior round) asserted a *lowered* auto-discovered budget was ignored is corrected to assert it's honored.

This also caught and fixed a real mypy regression the first round's fix introduced (a static import into `bundle_facts` pulled this module into the pre-existing CLI-registration cycle and broke type inference of `main` elsewhere) — reverted to the `importlib`-based read this file already uses for its other cross-cycle imports.

---

napetrov:
Two doc nits fixed in 5f05d8a9:

- **"anymore" wording** (`docs/use/multi-binary.md:513`) — fixed.
- **Explicit-`--config` requirement not documented** (`docs/contribute/plans/one-comparison-product.md:222`/702-703) — updated, with the more precise rule (an auto-discovered `.abicheck.yml` can still *lower* the budget, only *raising* it needs an explicit `--config`) rather than the suggested wording's blanket "explicitly supplied `.abicheck.yml` is required," which would have been slightly too strong now that the second-round fix lets auto-discovery lower the budget. Also added the same clarification to `docs/reference/config-file.md`'s `resource_limits:` section, which had the same gap.

---

napetrov:
Good catch — confirmed real. Fixed in 46c744e7: `_merge_config_overlay_with_discovered_project_config()` in `action/run.sh` now caps `resource_limits.max_bundle_facts_decode_nodes` the same way `build.query`/`compile.compiler`/`build.compile_db` are already stripped there, using the identical `resolve_max_json_object_nodes_cfg()` the native CLI calls (so the two can't drift). Capped rather than stripped — a lower value is never a decode-bomb risk, matching this PR's second-round fix. An explicit build-config input is unaffected (deliberate operator action, same as the other three guards).

Added two tests (raise-is-capped, lower-survives-uncapped) plus one assertion on the existing explicit-build-config test proving raising still works when the operator opts in explicitly.

---

napetrov:
**CI failure — "unit-tests (windows-latest, 3.13, false)"**: `TestReleaseTopologyOverlayPreservesWindowsQualifiedBuildConfigPath::test_same_drive_letter_shaped_path_is_prefixed_on_non_windows`/`test_ordinary_relative_path_still_gets_pwd_prefix_on_windows` (a Windows/git-bash `$PWD` POSIX-vs-native path-representation mismatch in this test's own assertion, unrelated to anything this PR touches). Confirmed pre-existing: fails identically on `main` at this PR's exact base commit (`3930cfce`, [CI run 34315674516](https://github.com/abicheck/abicheck/actions/runs/34315674516/job/102351260054)) — same two tests, same `assert '/c/Users/...' == 'C:\\Users\\...'` error. No fix exists yet upstream; not attempting one here since it's outside this PR's scope and would widen the diff into unrelated Windows-path-representation test-harness logic. Will keep watching (alongside the other pre-existing macOS bash-array failure already noted on this PR).

---

---

napetrov:
**CI failure — "test-action (required)"**: this job only polls for the separate "Test GitHub Action" workflow's own "test-action summary" check and timed out after 35 minutes because that workflow's run ([34329789280](https://github.com/abicheck/abicheck/actions/runs/34329789280)) never left `queued` on standard `ubuntu-latest` runners — a GitHub Actions runner-queue stall, not a code issue (nothing in this diff affects scheduling). Attempted the one allowed re-run; the API refused it (`This workflow is already running`) since the run is still queued, not completed. Will re-check and re-run once it actually completes or clears; keeping the PR watched.

---

---

PR #1180, opened 2026-09-09
feat!: close one-comparison-product Phase 5 and audit the whole compare/dump flag surface (7f + new 7i)

napetrov:
## Summary

Closes `docs/contribute/plans/one-comparison-product.md` **Phase 5** and **Phase 7f**, and adds **Phase 7i** — a per-flag audit of every surviving `compare` and `dump` option against ADR-068 D5's three guards, with an explicit written ruling for each, including the ones that stay. `compare` goes **56 → 50** options, `dump` **24 → 21** (Click-introspected, not read off the plan's prose).

## Motivation

ADR-068 D4: presentation must never change what gets analyzed, only what gets rendered. Phase 5 left three `compare` flags that gated *rendering* of data the run already computed unconditionally, and `--surface-metrics` had already been reduced to an accepted no-op — which D5 counts as public surface all the same. 7f's `dump` provenance merge was the last open Phase 7 row. 7d established the precedent that every flag gets a ruling written down, keeps included; 7i applies that to the whole remaining surface so no option is left unexamined.

## Changes

**Phase 5 — the three remaining AUTO rows, removed rather than left inert**

- `compare --surface-metrics` **removed, no replacement selector.** ADR-027's metric-drift findings are ordinary `changes[]` entries that every projection already renders, so nothing was left for the flag to select — not even a rendering choice. `surface_metrics` is also no longer a *parameter* of the documented Tier-2 verb `workflows.compare_policy.compare_snapshots`, which forces it on for every caller: a keyword the public Python API accepts and then ignores is the same dead surface one layer down.
- `compare --show-filtered` → **`--view filtered`.** The ADR-067 S1 scope/disposition ledger has been unconditional for a while and is always in `--format json`.
- `compare --audit-suppressions` → **`--view suppressions`.** The audit is computed on every run given `--suppress` and carried in JSON/SARIF/JUnit/HTML unconditionally. It inherits the no-op-without-`--suppress` rule verbatim.

After this, **no surviving `compare` flag decides whether a piece of the canonical result is computed.**

**Phase 7f — `dump` provenance merge**

`--git-tag` / `--build-id` / `--no-git` → one repeatable `--provenance KEY=VALUE` (`git-tag=<tag>`, `build-id=<id>`, `git=auto|off`). Grammar and parsing in a new pure-parsing leaf `abicheck/frontends/cli/options/provenance.py`, validated eagerly by its own Click callback — the same two-step shape `--view` uses on `compare`. Nothing is silently dropped: an unknown key, a missing `=`, an empty value and a bad `git=` value are each exit 64.

**Phase 7i — four more removals falling out of the audit**

- `compare --reconcile-build-context` → **AUTO.** ADR-039's reconciliation is strictly evidence-gated (a no-op unless *both* snapshots carry `build_context_defines` and per-field guards) and can only ever move a phantom finding out of the verdict into the audit bucket, never manufacture one — so an opt-in switch could only mean "leave a known false positive in your verdict because you forgot a flag". Forced on at the one Tier-2 chokepoint, so CLI, typed API and Action change together; `CompareRequest.reconcile_build_context` is removed with it, and the release fan-out now *gets* the behavior it used to reject.
- `compare --pdb-path` → **`debug.pdb_path`** (the key `dump` has read since Phase 7c; ADR-037 D8.1 forbids the two commands' debug context from drifting). The flag's per-side `old=`/`new=` scoping has no config spelling; that case is preserved by `--debug-root old=`/`new=`, which the debug resolver already searches for a PDB.
- `compare --support-promise` → **`release.support_promise`.** ADR-065 D1/D6 and the flag's own help text already called it a *contract-policy* field.
- `dump --compile-db-filter` → **`build.compile_db_filter`**, beside the `build.compile_db` it scopes.

**Measured keep, not an oversight — the dependency-walk family**

§4.1/§4.2 classify `--follow-deps` AUTO with a "cost check" prerequisite. The check was run and it **rejects** the row (the 7g precedent: measure, then let the measurement decide, including against the plan). A real `dump` with and without the flag differs by exactly one snapshot field, `provenance.dependency_info`, which embeds **absolute host paths**. Making the walk unconditional would put host-specific paths into every snapshot (breaking dump reproducibility and the ADR-050 comparability contract), add a host filesystem walk to every dump and both sides of every compare, and on `compare` let findings depend on which libraries the runner happens to have installed. `--search-path`/`--ld-library-path` stay with it as that walk's only inputs. Recorded in the plan with the measurement and the revisit condition.

**F-16 / F-19 — the executable invariant test (verification deliverable)**

The existing test was **not** an invariant test and had to be generalized. Two independent defects:

1. *It sampled; it did not cross the axes.* The exit-code half enumerated 192 invocations but asserted **only the exit code**; the result-block half used 8 hand-listed combinations, all at `--format json`. `--write` was not an axis at all, `show=` was not an axis, and format was never crossed with view for *content*.
2. *Its oracle was the code under test.* `_canonical_facts` projected the JSON report the implementation had just rendered and compared it to another rendering of the same report — which can only prove the renderings agree with each other.

Both fixed in `TestF19CanonicalResultIsInvariantOverTheWholeRenderingSpace`:

- **The full product**: 7 `--format` × 4 `--view` report modes × 3 demangle states × 2 `patterns` × 2 `filtered` × 2 `suppressions` × 3 `show=` states × 3 extra `--write` sets = **6048 invocations**. The `slow`-marked test runs every one; the default-lane test runs a *seeded random sample* (150 points, `ABICHECK_F19_SEED` to re-seed) of the same space rather than a hand-picked list. A third test pins the space's size against the product of the axis tuples, so a silently shrunk space fails instead of passing with less coverage.
- **The oracle is not the report projection**: it calls the Tier-2 engine verb (`workflows.compare_policy.compare_snapshots`) directly and reads verdict, finding identities, counts and assurance status off the returned `DiffResult` — a different code path from `report/render_json.py`, which every CLI invocation goes through.

**Counts and residuals** (Click-introspected on this branch):

| Command | Before | After | Target | Residual, itemized |
|---|---|---|---|---|
| `compare` | 56 | **50** | ≤40 | `--scope-public-headers`, `--post-manifest` (Phase 9) · `--env-matrix`, `--require-complete-analysis` (`scan`-owned, Phase 6) · `--instantiation-manifest`, `--use-cases` (prerequisite) · `--follow-deps`, `--search-path`, `--ld-library-path` (measured keep) · the declined `--old-variant`/`--new-variant` merge (−1) = 10 |
| `dump` | 24 | **21** | ≤16 | `--follow-deps`, `--search-path`, `--ld-library-path` (measured keep) · `--compression`, `--build-target` (documented Action inputs; removing the flag without the input would silently ignore supplied input) = 5 |

Both targets are arithmetically reachable, and every option still standing between here and them has a named owner and a named blocker rather than an unexamined one. Two of `compare`'s ten and two of `dump`'s five are *deliberate keeps*, so the honest reading is that §4.5's numbers predate Phase 2's five added flags and the dependency-walk measurement — recorded in the plan as still-useful pressure, not as arithmetic that survives the per-flag rulings.

**Merge criteria, per flag touched**

- Old spellings exit **64** with `No such option`, no hidden alias, no deprecation window (pinned in `tests/test_presentation_analysis_separation.py` and `tests/test_diff_reconcile.py`).
- Front-end parity in this PR: CLI, typed API (`CompareRequest.reconcile_build_context` removed; `compare_snapshots`'s `surface_metrics`/`reconcile_build_context` parameters removed) and the composite Action (no Action input or `run.sh` branch drove any removed flag — verified; `scan` routing deliberately untouched, it is another agent's area).
- No machine-readable contract changed shape, so no schema bump: the report schema, snapshot schema and `SCAN_SCHEMA_VERSION` are all untouched. The `.abicheck.yml` schema gains two additive keys (`release.support_promise`, `build.compile_db_filter`), which is what `config-keys-reference.md`'s regeneration records.
- Verdict / gate / exit code / coverage / assurance asserted unchanged **separately** by the 6048-point invariant test above against an engine-computed oracle, plus the rewritten reconciliation tests.
- Every removed spelling registered in `scripts/check_docs_contract.py`'s `_RETIRED_SURFACES` (via `scripts/retired_surfaces.py`).
- `docs/reference/cli-reference.md`, `config-keys-reference.md`, `python-api-reference.md`, the example-catalog pages and `repo_facts.json` regenerated by their generators.
- Changelog fragment added.

**Notes for the reviewer**

- `architecture/debt.yaml`: two baselines raised (`buildsource/build_config.py` 868→877, `cli_helpers_compare.py` 1359→1367) for the irreducible dataclass-field/parse/serialize cost of two new config keys, each with the usual detailed rationale. `frontends/cli/commands/dump.py` stayed under its baseline by moving the new `--provenance` option declaration into its own decorator module rather than trimming to fit.
- `release.support_promise: off` must be **quoted** in YAML — bare `off` is a YAML boolean, which the schema rejects rather than coerces. Same trap and same answer as `python.abi3_floor`'s bare `3.9`; documented at the key.
- 14 `docs-contract` **warnings** remain, all on narrative `docs/**` pages that still name a retired flag (down from 35). Narrative-page rewrites were explicitly out of this PR's scope and belong with the page owner.
- `ruff format --check` reports ~638 files needing reformatting in this environment, including files this PR never touches (`abicheck/detectors.py`, `abicheck/checker.py`) — a pre-existing toolchain-version difference here, not something this branch introduces.

## Checklist

- [x] Docs updated if user-visible behaviour changed (generated references + the owning plan; narrative pages out of scope)
- [ ] `pre-commit run --all-files` passes locally — the full unit lane is far slower than CI in this environment; `mypy abicheck/`, `ruff check`, `check_ai_readiness.py`, `check_architecture.py`, `check_docs_contract.py` and every affected test module were run clean, and the full suite is left to CI

https://claude.ai/code/session_014W5ov5GHqdqng3Ye9eNmXo

---
_Generated by [Claude Code](https://claude.ai/code/session_014W5ov5GHqdqng3Ye9eNmXo)_


napetrov:
## CI: `clang-plugin` / "C.6 + scan" jobs are failing, but not because of this PR

All eight `C.6 + scan (LLVM …)` matrix jobs fail at the same step, "Validate the Plugin injection scan flow (dump → merge → compare)":

```
+ abicheck dump /tmp/.../libwidget.so --build-info ./abicheck_inputs/ -o widget.baseline.json
merged baseline has no embedded build_source payload
##[error]Process completed with exit code 1.
```

This PR never touches `dump`'s `--build-info` inline-embedding/merge path, `dumper.py`, `service_dump_pipeline.py`, or any of the L4 source-extraction code the scan-flow fixture exercises. I confirmed it's not this PR's: the identical failure, in the identical step, across every LLVM matrix leg, already reproduces on `main`'s own tip (commit `e7bc4cb35`, [run 34391113483](https://github.com/abicheck/abicheck/actions/runs/34391113483)) — before this branch's merge commit existed. It's consistent (not a transient/infra failure — every leg fails at the same assertion after a real compile+dump ran), so it isn't a flake worth re-running.

Root cause, not yet fixed: `contrib/abicheck-clang-plugin/tests/scan_flow.py` runs `abicheck dump <so> --build-info <pack-dir> -o out.json` and asserts the resulting JSON carries a non-empty `build_source` key. It doesn't any more — the real ELF dump path was migrated onto the shared typed `execute_dump_request` pipeline in recent `main` history (ADR-063 Track 1/T4, e.g. commits `301e9639f`/`47c3fec59`/`16a693887`), and something in that migration stopped folding an inline `--build-info` pack into the emitted snapshot for a real binary dump. I haven't pinned the exact commit or the fix — that migration is unrelated to this PR's flag-consolidation work and I don't want to widen this PR's scope to chase it.

No fix exists yet on `main` to port. Since this doesn't block `test-action (required)` or the PR's own gating checks, I'm not spending further effort on it here — flagging it so whoever owns the ADR-063 dump-execution migration can pick it up, and keeping this PR watched.

---

napetrov:
## Re: "Pass the discovered config into inline dumps" (P1, `cli_compare_helpers.py:1741`)

Verified, and declining the suggested fix (threading `cfg_path` instead of the raw `config` into `_embed_inline_source_sides`'s `build_config=` kwarg) — it would reopen a trust-boundary hole this exact parameter already has a documented fix for.

`_embed_inline_source_side`'s own docstring (`frontends/cli/commands/compare.py`, "CLI cleanup phase two, Block 7 — PR C's tail") spells out why `build_config` here must stay the raw, explicit-only `--config` value and never `compare`'s own auto-discovered `cfg_path`:

> **Must stay the raw, explicit-only value, never `compare`'s own auto-discovery fallback** (`_resolve_compare_config`'s resolved `cfg_path`) — `embed_build_source` treats a non-`None` `build_config` as operator-authorized to execute `build.query` (ADR-032 D5); forwarding an auto-discovered path here would let an untrusted, PR-controlled `.abicheck.yml` in the sources tree authorize its own subprocess execution.

`discover_project_config()` walks up from the *current working directory* — in a realistic CI shape (`compare --old-sources ./old --new-sources .` run from inside a checked-out PR), that can resolve to an attacker-influenced `.abicheck.yml` sitting in the same checkout, not just a config genuinely under the operator's control. So `cfg_path` isn't a safe stand-in for "the operator's real config" here, even though it looks that way from `run_compare`'s own perspective.

The actual gap is narrower than the finding suggests: only `compile_db_filter` (a plain string read, no subprocess execution) is affected, while `build_config`'s *other* consumer on this same parameter (`build.query` authorization) is correctly gated. A real fix would thread `compile_db_filter` through its own, separate channel — never reusing the `build_config` parameter that also gates `build.query` — so a `compile_db_filter`-only value can't be mistaken for `build.query` authorization anywhere downstream. That's a real (if narrow) design gap worth closing, but it's a new parameter/threading change through `_embed_inline_source_side`/`_embed_inline_source_sides` and the dump-side resolvers, not a one-line fix, and risks quietly reintroducing the exact hole above if done carelessly. Leaving it for a follow-up rather than rushing it into this PR; the current behavior (silently falling back to the anchored-only `discover_build_config(sources)`, i.e. "no `compile_db_filter`" unless `.abicheck.yml` sits at the `--sources` root or `--config` was given explicitly) is a known limitation, not a silent wrong answer.

---

PR #1198, opened 2026-09-10
fix: use a real DLL, not a bare .exe, for the Windows live-artifact test

napetrov:
## Summary

`test_a_candidate_version_label_is_honoured` (`tests/test_compare_no_baseline_cli.py`) fails deterministically on `windows-latest` CI. This fixes the test's fixture, which was Windows-incorrect, not `compare`'s own validation logic (which is behaving correctly).

## Motivation

While investigating a separate CI-timeout PR (#1197), raising that job's timeout budget let the Windows unit-tests suite run to completion for the first time in a while — surfacing this test failure, which had previously been masked because CI kept getting cancelled by the timeout before reaching this test's place in the suite.

Root cause: the test stands in "any native binary" with `shutil.which("true")`, copying its bytes under a `.so` name and asserting `compare --no-baseline` succeeds against it. That holds on Linux/macOS, where a dynamically-linked ELF/Mach-O executable still carries a real dynamic symbol table — but on Windows, `shutil.which("true")` resolves to Git for Windows' `true.exe`, a plain PE executable with **no export directory at all**. `compare` reads the file's actual content (not its extension) and correctly rejects it: `Error: PE file '...' has no exports (named or ordinal). Verify the file is a valid DLL.` — exit code 64, not 0 as the test expected.

Confirmed as a real, reproducible failure (not a flake): it failed identically on two independent windows-latest CI runs (job [102767887923](https://github.com/abicheck/abicheck/actions/runs/34445033617/job/102767887923) and its rerun, job [102792300467](https://github.com/abicheck/abicheck/actions/runs/34445033617/job/102792300467)), same error both times.

## Changes

- `tests/test_compare_no_baseline_cli.py`: added `_native_library_with_exports()` — on Windows, uses a real system DLL (`kernel32.dll`, present on every supported Windows image) which genuinely has an export table; on every other platform, keeps the existing `shutil.which("true")` behavior unchanged, since it already works there.

## Bug-fix test contract

- Bug class: a test fixture's "any native binary works" assumption is false on Windows specifically, because a plain PE executable (unlike a dynamically-linked ELF/Mach-O executable) has no export table — so a Windows-only stand-in fixture is required wherever a test needs a "live artifact with exports."
- Publicly observable failure: `windows-latest, 3.13` `unit-tests` job fails `test_a_candidate_version_label_is_honoured` with `Error: PE file '...' has no exports ... Verify the file is a valid DLL` / `assert 64 == 0`, reproduced on two independent CI runs.
- Regression test fails on base: Yes — `test_a_candidate_version_label_is_honoured` itself is the regression test; it fails on `main` on `windows-latest` (confirmed via the two CI runs cited above; verified locally that the same test passes on Linux with no change needed there).
- Negative control: The fix is scoped to `sys.platform == "win32"` only; the existing `shutil.which("true")` path for Linux/macOS is left completely unchanged, so a hypothetical regression in the Windows-only branch cannot silently pass by falling through to the old behavior on those platforms — verified locally (`pytest tests/test_compare_no_baseline_cli.py -q`, 44 passed, on Linux).
- Public-surface test: Yes — this is itself a public-CLI-surface test (`invoke_cli("compare", "--no-baseline", ...)` through the real Click entry point), unchanged in that respect; only the input fixture changed.
- Axes covered: Single, narrow fixture-construction bug — one platform axis (`win32` vs. everything else), which is exactly what the fix branches on.
- General invariant: N/A — this is not a case where a broader class exists to generalize: `shutil.which("true")`-as-a-stand-in-native-binary appears nowhere else in the test suite (verified via repo-wide grep), so there is no sibling call site this bug's root cause could also be lurking in.
- Malicious fixture + side-effect absence: Not applicable — no action/workflow trust boundary touched; this is a pure test-fixture change reading a well-known, always-present local system file (`kernel32.dll`) and asserting existing CLI behavior.

## Checklist

- [x] Tests added / updated for new behaviour — the test itself is the fix; verified passing locally
- [ ] Changelog fragment added (`scriv create`) if this touches `abicheck/**/*.py` — N/A, `tests/` only
- [ ] Docs updated if user-visible behaviour changed — N/A, no user-visible behavior changed
- [x] `pre-commit run --all-files` passes locally — `ruff check`/`ruff format --check` both pass on the changed file
- [x] No new `mypy` or `ruff` errors introduced — `ruff check`/`ruff format --check` clean; pre-existing (unrelated) mypy debt in this test file is untouched, and CI's `mypy abicheck/` gate doesn't scope `tests/` anyway

https://claude.ai/code/session_014nMSZvqQiUbm6vLpxFXayw

---
_Generated by [Claude Code](https://claude.ai/code/session_014nMSZvqQiUbm6vLpxFXayw)_

### DIFF
--- a/tests/test_compare_no_baseline_cli.py
+++ b/tests/test_compare_no_baseline_cli.py
@@ -44,6 +44,35 @@ def invoke_cli(*args: str):
     return CliRunner().invoke(abicheck_main, list(args))
 
 
+def _native_library_with_exports() -> Path | None:
+    """A real, already-built native binary that exports symbols.
+
+    ``compare``'s live-artifact path requires an export table (a "DLL", in
+    its own error wording) -- a plain executable copied in under a ``.so``
+    name reads its file *content*, not its name, so it fails that check
+    regardless of extension. ``shutil.which("true")`` used to stand in for
+    "any native binary" on every platform, which holds on Linux/macOS
+    (a dynamically-linked ELF/Mach-O executable there still carries a real
+    dynamic symbol table) but not on Windows: Git for Windows' ``true.exe``
+    is a plain PE executable with no export directory at all, so the copy
+    fails abicheck's own "has no exports ... Verify the file is a valid
+    DLL" check every time -- reproduced on windows-latest CI (previously
+    masked by an unrelated job timeout that cancelled the run before this
+    test's turn in the suite). A real system DLL is the Windows-correct
+    stand-in; every supported Windows image ships one at this path.
+    """
+    import shutil
+
+    if sys.platform == "win32":
+        import os
+
+        system_root = Path(os.environ.get("SystemRoot", r"C:\Windows"))
+        dll = system_root / "System32" / "kernel32.dll"
+        return dll if dll.is_file() else None
+    which = shutil.which("true")
+    return Path(which) if which is not None else None
+
+
 @pytest.fixture
 def candidate() -> Path:
     """A committed G20 audit fixture that reports exactly one finding."""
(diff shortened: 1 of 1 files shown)


napetrov:
Closing as redundant — this fix was ported into and merged via #1197 (commit 1f0a5c35, carried through to main in a448081). Same change reaches `main` either way; no need to merge both.

---

PR #1241, opened 2026-09-12
fix: separate declared from exported on Function/Variable (ADR-071)

napetrov:
## Summary

Separate *declared in the source surface* from *dynamically exported* on `Function`/`Variable`, so a declaration that never left the public headers stops being reported as a source removal when the binary stops exporting it.

## Motivation

The fourth defect from the report that produced ADR-069 (the other three are already fixed on main). `Visibility` folds two independent facts into one member:

```python
PUBLIC = "public"      # default visibility / exported
HIDDEN = "hidden"      # __attribute__((visibility("hidden")))
ELF_ONLY = "elf_only"  # present in ELF symbol table, not in headers
```

`PUBLIC` means "declared in a parsed header **and** dynamically exported", and there is no member at all for the ordinary fourth combination — declared and *not* exported (an inline member, a fully-inlined function, one a `-fvisibility=hidden`/version-script change stopped exporting). The 39 `visibility == Visibility.PUBLIC` guard sites each got the question its author meant right and the other one by accident, and no test of either question alone could see it.

Observed on two real artifacts of one library with **byte-identical headers**, differing only in export emission: the still-declared public inline `svs::runtime::v0::DynamicVamanaIndex::add(size_t, const float*)` dropped out of `diff_namespaces`' source-surface index and was reported as `EXPERIMENTAL_REMOVED_WITHOUT_REPLACEMENT`. Both GCC 14 and Clang 17 compile an explicitly-qualified call to it against either header set.

Design recorded in [ADR-071](docs/contribute/adr/071-declared-and-exported-are-independent-facts.md); the known-gaps entry is rewritten to record what landed and what is still open.

## Changes

- **Model** — `Function`/`Variable` gain `declared_fact`/`exported_fact` (`Fact[bool]`). Neither has a legacy sibling, so an omitted field resolves to `Fact.not_collected()`: a headerless dump reports declaration presence as *unknown* and a snapshot with no observed export table reports export status as *unknown*, never a `False` that would fabricate findings. `Visibility` is untouched.
- **Persistence** — snapshot schema **v46**, additive, via `storage/fact_codec.py`. A pre-v46 snapshot loads both as "no evidence" and produces bit-for-bit identical findings. The ADR-050 comparability contract does not gate on `schema_version`, and neither the removal guard nor the new kinds fire without positive evidence on the side they read, so a pre-v46 side compared against a v46 side degrades to the pre-fix answer rather than a wrong one.
- **Producers** — `extract/declaration_surface_stamp.py` stamps both facts for every header-AST declaration at the two choke points that hold the export sets (`dumper_manifest.resolve_header_ast_result` for ELF, `extract/header_ast_fields.parse_header_ast_fields` for PE/Mach-O), deriving the export answer from the backend's own `Visibility` decision rather than re-matching mangled names. No observed export table ⇒ `NOT_COLLECTED`. Export-table-only synthesis stamps the export fact and leaves declaration unknown.
- **Consumers** — all 39 guards audited. `diff_namespaces`' five (source axis) moved to `in_source_surface` plus a `source_removal_supported` guard; the rest meant the conjunction and now say so via `in_exported_public_api`, which is behaviourally identical. `model/declaration_surface.py` owns the reading rule.
- **Reporting** — `FUNC_EXPORT_REMOVED_STILL_DECLARED` / `VAR_EXPORT_REMOVED_STILL_DECLARED`, both BREAKING, replace `FUNC_REMOVED`/`VAR_REMOVED` when the export went away and the declaration did not. The verdict is deliberately unchanged (an already-linked consumer still fails to resolve the symbol); only the false "removed" claim goes.
- Housekeeping the no-growth gates forced, each a real move rather than a trim: `ElfHeaderAstResult` → `extract/header_ast_result.py`, `_EQUIVALENT_CHANGE_CATEGORIES` → `compare/equivalent_change_categories.py`, the export-axis finding → `compare/export_axis_findings.py`, the removal guard → `compare/source_surface_removal.py`. `diff_namespaces`' baseline is raised 1253 → 1283 with a recorded rationale (the growth there is the per-guard comments).

## Bug-fix test contract

- Bug class: `evidence.independent_facts_conflated_into_one_signal` (new entry, `tests/regressions/manifest_evidence.py`; the sibling `classification.name_shape_as_contract_membership` is about naming, this one about evidence selection)
- Publicly observable failure: comparing two builds of one library with byte-identical headers that differ only in export emission reports `EXPERIMENTAL_REMOVED_WITHOUT_REPLACEMENT` for a public inline method still declared in both header sets
- Regression test fails on base: yes — `tests/test_declaration_surface_properties.py::test_the_reproducing_defect_is_not_a_source_removal` and `tests/test_declared_vs_exported_end_to_end.py::test_export_dropped_is_not_reported_as_a_source_removal`
- Negative control: `test_a_genuine_source_removal_is_still_reported`, `test_a_genuine_experimental_namespace_removal_is_still_reported` and `test_a_public_vtable_or_layout_change_is_still_breaking` — the fix must not work by suppressing real findings; plus the new FP-corpus case `export_dropped_declaration_retained`, filed as a real-break sentinel rather than internal noise because the export loss *is* breaking
- Public-surface test: yes — `tests/test_declared_vs_exported_end_to_end.py` runs the whole scenario through `checker.compare()` in both directions and each side against itself, and round-trips both facts through `save_snapshot`/`load_snapshot`
- Axes covered: declared ∈ {yes, no, unknown} × exported ∈ {yes, no, unknown}, on both sides (81 combinations); direction ∈ {old→new, new→old, self}; declaration kind ∈ {Function, Variable}; every `FactStatus` spelling of "no evidence"; schema ∈ {v46, pre-v46}
- Real-dependency test: yes — the schema half goes through the real `save_snapshot`/`load_snapshot` chokepoints (`test_new_snapshot_round_trips_both_facts`, and the pre-v46 test which rewrites a genuine dump via `from_sectioned_document` rather than hand-building a document); the two new keys also pushed `tests/test_bounded_decoded_prefix_properties.py`'s low-compression-ratio fixture out of its regime, which its own precondition assertion caught and which is re-tuned here rather than silenced
- Malicious fixture + side-effect absence: not applicable and deliberately not faked — this diff adds no action/workflow surface and parses no untrusted input; the new code paths read already-parsed model objects
- Must-merge / must-not-merge pair: `test_the_export_axis_kind_dedupes_against_a_plain_removal_of_the_same_symbol` (the new kinds map onto the existing `func_removal`/`var_removal` dedup categories, so a second detector's plain removal of that symbol collapses onto it) and `test_it_stays_distinct_from_the_same_kind_on_a_different_symbol` (which "collapse everything" would also have satisfied)
- False-positive removed / real break preserved: `test_export_dropped_is_not_reported_as_a_source_removal` alongside `test_a_genuine_source_removal_is_still_reported`, `test_a_genuine_experimental_namespace_removal_is_still_reported`, `test_a_public_vtable_or_layout_change_is_still_breaking`, and the FP-rate corpus's new `export_dropped_declaration_retained` real-break sentinel (0/0 FP/FN, unchanged)
- Verdict, gate and exit code checked independently: `test_verdict_gate_and_exit_code_each_say_breaking` — `Verdict.BREAKING`, the finding present in the scored change set, and `compute_exit_code(...) == 4` asserted separately
- General invariant: a declaration's presence in the source surface and its availability as a dynamic export are independent facts — neither may be inferred from the other, and absent evidence for one may never be read as a negative answer for it. `tests/test_declaration_surface_properties.py` enumerates the full 9×9 quadrant cross-product through the real detectors and checks each cell against a hand-written oracle matrix written out in the test file, not against `source_removal_supported`/`export_only_loss` (the code under test); it additionally asserts axis-independence, order-independence and Function/Variable symmetry of the selection primitive.

## Checklist

- [ ] `pre-commit run --all-files` passes locally

https://claude.ai/code/session_01Lov6uAXcQNi1BuwoEArC7Q

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lov6uAXcQNi1BuwoEArC7Q)_


napetrov:
**Blocked: this PR is superseded by #1235, and I am not resolving the conflicts.**

#1235 ("fix: split the three facts `Visibility` conflated", merged as `c29824f8`) landed the same fix while this branch was in flight. It is not a partial overlap — it is the same change, done more thoroughly:

| | #1235 (on `main`) | this PR |
|---|---|---|
| Model | `declared_in_headers_fact` / `in_public_contract_fact` / `binary_exported_fact` | `declared_fact` / `exported_fact` |
| Reader | `model/surface_facts.py` | `model/declaration_surface.py` |
| Schema | v46 | v46 — same number, different fields |
| Export-axis kinds | `FUNC_EXPORT_REMOVED`/`_ADDED` + variable counterparts | `FUNC_EXPORT_REMOVED_STILL_DECLARED` + variable counterpart |
| Bug class | `evidence.export_presence_as_declaration_presence` | `evidence.independent_facts_conflated_into_one_signal` |
| Quadrant test | `tests/test_surface_fact_split.py`, 3-way `itertools.product` | `tests/test_declaration_surface_properties.py`, 2-way, 81 cells |

`main`'s split is strictly richer: it separates *declared in headers* from *member of the promised public contract*, which this branch folds together through `ScopeOrigin`.

The 45 conflicting files are that duplication, not an ordinary merge. Resolving them would leave the repository with two parallel models of the same facts and two claimants on schema v46 — the second parallel path `AGENTS.md` requires be deleted rather than added. So the merge was aborted deliberately; nothing here is waiting on CI or review.

**Recommended: close as superseded.** Two pieces are not in #1235 and port onto `main` independently of the model split, if wanted:

1. `export_dropped_declaration_retained` in `scripts/check_fp_rate.py` — byte-identical headers, one export dropped, filed as a *real-break* sentinel (not internal noise: the export loss genuinely breaks an already-linked consumer). ~25 lines plus a `CASE_CATEGORY` row and `+1` to the `exports` shadow-loss budget.
2. `docs/contribute/adr/071-declared-and-exported-are-independent-facts.md` — #1235 landed with no ADR, and the known-gaps entry it closes said the fix needed one. It would need rewriting against the three-fact model, but its alternatives-rejected section (why not a fourth `Visibility` member, why not "ignore export changes for inline functions", why not patch `diff_namespaces` alone) applies unchanged.

No bot findings are outstanding: Codex's security review completed clean on `4ebd846`, and both CodeRabbit and Codex code review hit usage limits without reviewing.

---

napetrov:
Noting the one red check so it isn't mistaken for a real failure: `test-action summary` failed on `4ebd846`, and every job in that run reports `"result": "cancelled"` — the run was superseded by the push of `4fdc115` (the current head) and killed mid-flight. No test executed and failed; nothing to fix or re-run.

The blocker stands as described above: this PR duplicates #1235, which is already on `main`, and should be closed as superseded rather than merged.

---

PR #1243, opened 2026-09-12
perf: cut repeated whole-repo work and fix a vacuous differential test

napetrov:
## Summary

Acts on a test-level audit of this suite: repairs one test that was silently asserting nothing, gives three weak-oracle/repeated-work areas real contracts, removes redundant whole-tree work from both the suite and a shared gate, and works the duplicate-test queue to zero in the category that matters.

## Motivation

The audit's central claim is that the largest savings are in test *infrastructure* and repeated whole-repository scans, not in deleting regression tests — and that some of the most expensive tests have the weakest oracles. Checking each finding against the code confirmed that, turned up one outright bug (a differential test comparing a result with itself), and **refuted two of the audit's own claims** with measurement.

Every change is validated by **falsification** — reintroducing the defect, or substituting a deliberately wrong implementation — not by the fix's own passing run. Per `AGENTS.md`, a test written to confirm the fix just made only encodes the bug its author already thought of.

## Changes

- **A real bug: a differential test comparing a result with itself.** `_isolate_ast_cache` in `tests/test_clang_header_backend_integration.py` derived the same `tmp_path / "xdg-cache"` root on *every* call, so both double-`dump()` streaming-pruner tests pointed their pruning-off and pruning-on runs at one cache. The second was served the first's AST, the pruner never parsed anything, and both the equivalence and method-count assertions were vacuously true. The comment on the second call asserted "a *different*, still-fresh cache dir"; it never was. Each configuration now gets a distinct root with the in-process memo cleared alongside it, and a `_PruneSpy` asserts **within the same comparison** that the pruning path executed.
- **The autouse cache fixture.** `tmp_path_factory.mktemp` allocates a *numbered* directory, enumerating every sibling to pick the next number; the fixture is autouse, so test N+1 pays a scan of N entries — charged even to tests that only compare two enum values. `tempfile.mkdtemp` inside pytest's own basetemp gives the identical guarantee from an atomic random name. `tests/test_conftest_cache_isolation.py` states the isolation contract as invariants so the next "make this faster" round cannot reach for a shared session directory.
- **`TestExhaustiveMatrix`'s oracle.** 1,608 parametrized cases asserted only that the result was one of four enum members. Replaced with batched checks against `_expected_verdict` — an independent second derivation from the intrinsic `*_KINDS` partitions and the two named downgrade sets, deliberately **not** from `policy_kind_sets`, which is what `compute_verdict` folds over. A vacuity guard covers the oracle itself. Also removed four `A & B ⊆ A` tests: set algebra, true for any two sets including two wrong ones.
- **Stopped re-running the repository to test argument dispatch.** `test_main_returns_zero_on_clean_tree` drove the live tree through nearly the whole readiness registry to assert `main()` returns 0 — work the dedicated `ai-readiness` CI job already does with *no* skips. `tests/test_ai_readiness_main_dispatch.py` keeps `main()`'s own contract against the real `main()` over a small instrumented registry.
- **Made the shared gate faster instead of relocating its tests.** The two sibling live-tree scans (`fact-detector-misuse`, `fact-field-readers`) are deliberately **kept** in the unit lane — unlike the `main()` case they assert something substantive, so moving them would stop a contributor learning locally that they introduced a violation. Profiling found the cause: `fact_equality_misuse_sites` computes three whole-tree helpers, then calls `_fact_aliases`, which computes all three again — 4×/2×/2× full `ast.walk` per file over 787 files, the bulk of 13.7M walk calls. `_memoize_per_tree` caches each on the tree node's own `__dict__`.
- **The duplicate-test queue, worked to zero in its useful category.** `scripts/find_duplicate_tests.py` (committed, documented, deliberately *not* gated) screens same-module clones by body + parameters + decorators. Clones across two differently-named classes are usually intentional; the category that matters is same scope with **different names**, because a test whose name promises input X while its body tests input Y makes a gap look filled. All nine such groups resolved — four were one assertion under two names, and five hid something, including a wholly untested **workflow-wiring** claim (`performance.yml` must source the PR's whole current label set, not the delivered event's) now asserted by `TestWorkflowWiring`.
- One exact in-class duplicate in `tests/test_finding_identity.py` folded into its surviving twin.
- `AGENTS.md` records each lesson under "Test-quality gates"; `tests/regressions/manifest.py` gains the `guard.differential_test_shares_state_with_itself` bug class with its residual tracked as a `KnownGap`.

### Validation

| Change | Falsification | Result |
|---|---|---|
| Pruner cache isolation | Reintroduce the shared root | `spy.calls == 0` — the pruner had **never run**; both tests fail. All 4 pass with the fix. |
| Cache fixture | Regress to a shared session directory | 16 of 24 isolation checks fail |
| Policy matrix | Substitute always-`COMPATIBLE`, then always-`BREAKING` | Both passed all 1,608 cases before; both now rejected |
| Gate memoization | Capture per-file site lists memo-on vs memo-off | **Byte-identical across all 787 files** |
| Memo cache key | Mutate to `key = ()` | Caught (see caution below) |
| Memo actually caches | Mutate to pass-through | Caught — output equivalence alone cannot detect an absent cache |
| Workflow wiring | Source labels from `github.event.label.name`; gate the env on `labeled` | Each caught by its own named test |

**Full fast suite, twice on the merged tree: 43,179 and 43,176 passed**, 44 skipped, 4 xfailed. The single failure in each is a local-environment artifact of this container's shallow clone, not a CI failure — see "Pre-existing, not from this PR" below. The 3-test delta between runs matches the clone folds exactly, offset by the 3 new `TestWorkflowWiring` tests — the check that no fold silently took a third test with it.

Measured: fixture A/B on the same 1,107 tests at 0/8,000/20,000 pre-existing siblings — **2.33/2.67/2.87s → 1.72/1.80/2.15s**. `tests/test_ai_readiness.py` 5.57s (was ~5m for one test). `tests/test_policy_override_matrix.py` 1.63s / 53 tests (was ~58s / 1,608). Gate scan **25.56s → 18.96s**.

**Coverage, since removing test definitions was the real risk here:** measured on the canonical lane's exact command, including `golden` and instrumentation — **96.22%** on this branch against **96.21%** on `main`'s own canonical CI lane, i.e. marginally above `main` rather than below it, against a 95% floor. The canonical `unit-tests (ubuntu-latest, 3.13)` lane passes on this head.

### Two measurements that refute audit claims

1. The audit proposed "possibly sharing a parsed repository inventory between checks." Reading and parsing all 787 files costs **~1.1s total**, against **~43s** for those two tests — the scan logic dominates by roughly 40×, so a shared parse would buy ~2s of 43s.
2. The audit reported **61** clone groups / 64 excess definitions. Re-screening (body + parameters + decorators) finds **29 / 31**. I can't reconcile that without their script, which is why mine is committed: the next reader gets a reproducible baseline rather than a number to trust.

Both recorded in `AGENTS.md` so neither is retried.

### Three cautions worth not rediscovering

- **An untested cache key is not an unnecessary one.** The first version of `tests/test_fact_detector_misuse_memoization.py` had 77 tests, and a `key = ()` mutation — dropping the second argument from the cache key — **passed all of them**, because every real caller happens to pass the one memoized spans object per tree. The second argument is plainly observable, so `test_cache_key_includes_the_second_argument` now calls the helper twice on one tree with two genuinely different span arguments.
- **Mutate before claiming a gap.** One clone group looked like a missing one-sided-evidence case in the ELF alignment detector, and I wrote a test whose comment asserted that gap. A mutation test showed `test_declared_alignment_known_one_side_only_falls_back` already covered it — the speculative test was withdrawn and the clone folded instead. Otherwise a false claim would have shipped in a code comment.
- **Strip comments before asserting against workflow text.** The first `performance.yml` assertion failed on its own first run because the file documents the rejected `github.event.label.name` spelling *in a comment*, so a raw substring search reported the thing it was checking for as absent.

### Deliberately not done

- **Two of three audit-flagged "duplicates" kept.** Both `tests/test_signature_normalization.py` cases are cross-class guards that already name their counterpart in a comment. The audit concedes zero coverage change and no meaningful runtime saving, so removing them trades a stated regression guard for nothing.
- **`test_sc_offline_snapshot` was not removable at all** — `tests/test_scenarios.py` separately asserts one `test_sc_*` per automated catalog scenario. Strengthened to assert the offline property rather than inherit it from the `_compare` helper.
- **Native-test lane ownership, storage-scale lanes, presentation/analysis split** — CI-topology decisions, not edits.
- **One stated residual:** no test pairs a *narrowed* header-alias pass against an unmarked side in `test_l3l4l5_new_kinds.py`. Stated rather than guessed, since the expected value is a behaviour question for whoever owns that alias.
- **A gap found but not closed:** `mypy` reports 160 errors across 33 of 88 files in `scripts/`, which CI never type-checks (`mypy abicheck/` only). That is how 7 type errors in this PR's new script reached CodeFactor rather than a local gate. Fixed in this PR's own file; extending the gate needs its own baseline mechanism and belongs in a separate change.

### Pre-existing, not from this PR

Every failing check on this head has a named failing test traced to a commit outside this diff, and `git diff --stat origin/main...HEAD -- abicheck/` is **empty** — this branch changes no product code at all. Each is documented in its own comment below, with the reproduction:

| Failure | Origin established by |
|---|---|
| `case96_hidden_friend_removed` — `expected='API_BREAK' got='BREAKING'` (`Validate examples` shard 1/3 on gcc **and** clang, plus `integration-tests` on all three platforms) | Pristine `origin/main` worktree, both toolchains |
| 2 × `tests/test_release_assurance_cli.py` (`integration-tests (ubuntu-24.04)`) | Pristine worktree of the merge-base `324815f7` |
| 71 action/workflow-shell failures (`unit-tests (windows-latest)`) | Reproduce on `main` `851b1bd4`; zero file overlap with this diff |
| 12 collection errors — `read_text()` with no `encoding=` vs Windows cp1252 (`unit-tests (windows-latest)`) | All three files present on `origin/main` with the same unqualified call |
| `mutmut (detector core)` — clean pre-flight run aborts | Two candidate explanations set out in a comment; needs runner visibility I do not have |

**One retraction, kept here rather than only in a comment.** An earlier version of this section claimed `test_adr_status_sync_holds` and the `adr-status-sync` gate fail on clean `main` because ADR-049's `**Verified:**` sha is unreachable. That was wrong: `ai-readiness` **passes** in CI on this head, including its architecture-contract step. The sha resolves fine there and fails only in *this* container, whose clone lacks that commit — a local-environment artifact, not a CI failure, and it explained none of the reds it was cited for.

Separately, `ruff format --check` already flags 618 files on `main` under the pinned ruff 0.16.3; this PR's delta there is zero (both edited files were already flagged; all new files are clean).

## Bug-fix test contract

- Bug class: `guard.differential_test_shares_state_with_itself` (new entry in `tests/regressions/manifest.py`)
- Publicly observable failure: two streaming-pruner tests reported success while the pruner never executed, so a regression in the pruning path would have shipped undetected
- Regression test fails on base: yes — restoring the shared cache root makes `test_streaming_pruner_produces_an_equivalent_public_model` and `test_streaming_pruner_never_prunes_a_method_shaped_node_end_to_end` both fail on `spy.calls == 0`
- Negative control: `key = ()` and pass-through mutations of `_memoize_per_tree` are each caught by a distinct named test; always-`COMPATIBLE`/always-`BREAKING` substitutions are rejected by the policy matrix
- Public-surface test: none — every seed test calls an internal entry point (`dump()`, `_clang_header_dump`, a pytest fixture's allocation), so no public surface is claimed
- Axes covered: `shared_state` (on_disk_cache, in_process_memo), `configuration_pair` (pruning_off_vs_on, allocator_old_vs_new), `observation` (outputs_only, mechanism_spy)
- General invariant: a test asserting two configurations agree must establish, within that same comparison, that each configuration actually executed — observed via the mechanism, never inferred from the outputs being compared, since outputs are exactly what a stale cache makes indistinguishable

## Checklist

- [x] Changelog fragment — n/a, no `abicheck/**/*.py` change (`scripts/` and `tests/` only)
- [x] Docs updated if user-visible behaviour changed — n/a for behaviour; `AGENTS.md` and `scripts/CLAUDE.md` updated for contributor guidance
- [ ] `pre-commit run --all-files` passes locally — not run in full; `ruff check`, `ruff format --check`, the readiness gate, `mypy` on changed files, and the full fast suite were run individually

https://claude.ai/code/session_01H6Ddp5JySQNDsSrBSVQseG


napetrov:
## Two failing checks are the base branch's, not this PR's

Recording this once, with the evidence, rather than leaving them unexplained. **This PR's diff touches no product code at all** — `git diff --stat origin/main...HEAD -- abicheck/` is empty; all 24 changed files are under `tests/`, `scripts/`, or are Markdown — so it cannot move a compatibility verdict or an ADR receipt.

### 1. `Validate examples (clang, debug-headers, shard 1/3)` — `case96_hidden_friend_removed`

```
FAILED tests/test_example_autodiscovery.py::test_example_source_smoke[case96_hidden_friend_removed]
AssertionError: case96_hidden_friend_removed: source-smoke validation FAIL —
  expected='API_BREAK' got='BREAKING'
  (consumer hidden-friend smoke: old equality expression compiles, new header removes hidden friend)
```

Reproduced on a pristine worktree of `origin/main` (`324815f7`), with no part of this branch applied:

```
$ git worktree add /tmp/mainwt origin/main
$ PYTHONPATH=. pytest tests/test_example_autodiscovery.py -m integration \
    -k case96_hidden_friend_removed
E  AssertionError: case96_hidden_friend_removed: source-smoke validation FAIL —
   expected='API_BREAK' got='BREAKING'
1 failed, 1 skipped, 229 deselected
```

Byte-identical expected/got. `catalog/ground_truth.json` pins this case to `expected: API_BREAK` with `expected_kinds: ["hidden_friend_removed"]`, and its own description notes the detection is "derived from castxml's `befriending` attribute" — so the verdict landing on `BREAKING` is a detector/classification question in `abicheck/`, which this PR does not touch.

**No fix ported, because none exists to port.** Nothing in `origin/main`'s recent history touches `examples/cases/case96_hidden_friend_removed` or the hidden-friend detector, and `docs/contribute/known-gaps.md` has no entry for it. Widening this PR — a test-suite-cost change — to alter a detector's verdict classification would be exactly the scope creep `AGENTS.md` warns against, and it needs someone who owns that classification to decide whether `API_BREAK` or `BREAKING` is correct here.

**Not re-run**, because it is not a flake: it reproduces deterministically on clean `main`, so a re-run would spend the one available attempt to confirm something already confirmed against the base branch directly.

The same log also reports two non-blocking `KINDS_MISMATCH` lines on cases marked `[PASS]` (`case16_inline_to_non_inline` missing `func_added`; `case191_header_only_graph_field_type` missing `struct_size_changed`). Flagging them as observed; they did not fail the job.

### 2. `ai-readiness` / `test_adr_status_sync_holds` — ADR-049's `**Verified:**` sha

ADR-049 names commit `3930cfce`, which is not reachable from `origin/main`, so the gate errors. This also fails on clean `main` at `c29824f8`, and the gate's own message describes the cause: a receipt anchored to the branch that added it disappears on merge and then fails permanently on `main`.

Worth separating from the above: an **earlier** `ai-readiness` failure on this PR *was* mine — a `debt-no-growth` breach on `tests/test_clang_header_backend_integration.py`, which I had initially mis-attributed to this ADR issue before reading the job log. That is fixed in `ee30b0ad` by moving the cache-isolation helpers into their own module (1678 → 1601 lines, `check_architecture.py` clean), not by raising the baseline.

### Where this PR stands

Locally on the merged tree: the full fast suite passes twice at **43,179** and **43,176** tests, with only `test_adr_status_sync_holds` failing — i.e. only the base-branch item above. `CodeFactor` is green. `check_architecture.py` reports 0 errors.

I am keeping this PR watched and will act on any failure that is genuinely its own.

---

napetrov:
One addition to the above, because it sharpens the diagnosis rather than repeating it: `case96_hidden_friend_removed` fails on the **gcc** leg of the same shard too (`Validate examples (gcc, debug-headers, shard 1/3)`, `CC: gcc` / `CXX: g++`), with a byte-identical message and the same two non-blocking `KINDS_MISMATCH` lines.

So it is not a clang-backend limitation. It reproduces on both toolchains in CI and on clean `origin/main` locally under neither explicit `CC` nor `CXX` — consistent with a compiler-independent classification question about whether removing a hidden friend should read `API_BREAK` (as `catalog/ground_truth.json` pins) or `BREAKING` (as the detector currently returns).

Conclusion and disposition are unchanged: not this PR's, no fix exists to port, and no re-run spent on a deterministic base-branch failure.

---

napetrov:
## `test-action (required)` timed out waiting on a starved runner queue — not a code failure

Recording this once for the whole class, so subsequent instances on superseded heads don't need their own notes.

`test-action (required)` on `15465aa3` failed like this:

```
test-action summary not reported yet -- waiting      (repeated ~22×, over 6 minutes)
##[error]Timed out waiting for test-action summary to complete
```

It is an aggregator that waits for the real `test-action` matrix jobs to publish a summary. Those jobs never reported, so it timed out. No test body ran inside it at all — nothing was executed and then found wrong.

The cause is queue starvation, observed directly: jobs created at 07:44Z were still `queued` more than 20 minutes later without starting, and a fresh check suite created for the current head at 08:33Z was likewise entirely queued. **This PR's diff touches neither `action/` nor `.github/workflows/`** (`git diff --name-only origin/main...HEAD -- action/ .github/workflows/` is empty), so it cannot affect how that composite action behaves.

**Not re-run, deliberately.** The rules would allow a re-run for a job that died before executing anything, and this qualifies — but `15466aa3` is four commits behind the current head `ee30b0ad`, which already has its own check suite queued. Re-running a superseded head's job would gate nothing and would consume more capacity from the very queue that caused the timeout. If `test-action` times out the same way on the current head, that is the signal worth acting on, and I will.

### Running tally of this PR's CI failures

| Failure | Whose | Disposition |
|---|---|---|
| `CodeFactor` — 7 type errors in `scripts/find_duplicate_tests.py` | **Mine** | Fixed at the type level in `245b187f` |
| `ai-readiness` — `debt-no-growth` on `tests/test_clang_header_backend_integration.py` | **Mine** | Fixed in `ee30b0ad` by moving the cache-isolation helpers to their own module (1678 → 1601 lines), not by raising the baseline |
| `ai-readiness` / `test_adr_status_sync_holds` — ADR-049 `**Verified:**` sha | Base branch | Reproduces on clean `main`; carries no pytest marker, so it also fails every `unit-tests` platform lane |
| `Validate examples` shard 1/3 (gcc **and** clang) — `case96_hidden_friend_removed` | Base branch | Reproduced on a pristine `origin/main` worktree, both toolchains |
| `test-action (required)` | Infrastructure | This comment |

Two of the five were mine and are fixed. One open question remains that could still be mine, and I am measuring it rather than assuming: whether the canonical lane's **95% coverage floor** still holds, given this PR removes test definitions (four clone folds, the whole-tree readiness test, 1,608 matrix cases collapsed to a handful of batched ones). Every full-suite run I have reported was uninstrumented *and* excluded `golden`, so none of them is evidence about coverage. I am running the lane's exact command locally and will report the number. If the floor is breached, the fix is restoring the lost coverage — `AGENTS.md` forbids lowering it.

---

napetrov:
## Correction: my earlier account of the `unit-tests` failures was wrong

I need to correct something I stated in a comment above, because it was a confident claim and it does not hold up.

**What I claimed:** that `test_adr_status_sync_holds` carries no pytest marker, therefore fails in every `unit-tests` platform lane, and that this alone explained all five of them.

**What is actually true:** `ai-readiness` **passes** on this PR's current head `ee30b0ad` — including its "Architecture contract (ADR-061)" step — and `main`'s own canonical unit lane passes its unit run too. The ADR-049 `**Verified:**` sha resolves fine in CI; it fails only in *this* container, whose clone lacks commit `3930cfce`. So that item is a **local-environment artifact, not a CI failure at all**, and it explains none of the unit-lane reds. I should have checked a CI result before generalising from my local one.

### What the `unit-tests` failures actually are

Both reproduce on `main` at `851b1bd4` — a merge commit containing nothing from this branch. That run fails five jobs, two of them unit lanes:

| Lane on `main` `851b1bd4` | Result |
|---|---|
| `unit-tests (windows-latest, 3.13)` | **71 failed**, 43360 passed, 1200 skipped (43:13) |
| `unit-tests (ubuntu-latest, 3.13)` | unit run **passed at 96.21% coverage**; the *slow* step failed, 1 failed / 169 passed |

The 71 Windows failures fall in exactly seven files:

```
tests/test_action_cli_surface_generated.py
tests/test_action_run_sh_annotate_renderer.py
tests/test_action_run_sh_compare_pr_json_write.py
tests/test_action_run_sh_helpers.py
tests/test_extra_args_is_value_option_completeness.py
tests/test_reusable_workflows_assurance_overlay_extra_args_config.py
tests/test_source_input_completeness.py
```

and the Ubuntu slow-step failure is in `tests/test_presentation_analysis_separation.py`.

**None of those eight files appears in this PR's diff** — checked mechanically against `git diff --name-only origin/main...HEAD`, zero overlap. They are shell/workflow-parsing and presentation tests; this PR touches none of that surface.

### Coverage, since removing tests was the real risk here

This PR removes test definitions (four clone folds, the whole-tree readiness test, 1,608 matrix cases collapsed to a handful of batched ones), so the honest question was whether executed-line coverage fell. Measured on the canonical lane's exact command, including `golden` and instrumentation:

| | Total coverage |
|---|---|
| `main` `851b1bd4`, CI canonical lane | **96.21%** |
| This branch `ee30b0ad`, same command locally | **96.22%** |

Against a 95% floor, and marginally *above* `main` rather than below it. 43,196 passed locally.

### Corrected tally

| Failure | Whose | Status |
|---|---|---|
| `CodeFactor` — 7 type errors in `scripts/find_duplicate_tests.py` | **Mine** | Fixed, `245b187f` |
| `ai-readiness` — `debt-no-growth` on `tests/test_clang_header_backend_integration.py` | **Mine** | Fixed, `ee30b0ad`; `ai-readiness` is now green on the current head |
| `unit-tests (windows-latest)` — 71 action/workflow tests | Base branch | Reproduces on `main` `851b1bd4`; zero file overlap with this diff |
| `unit-tests (ubuntu-latest)` slow step — `test_presentation_analysis_separation` | Base branch | Same run on `main` |
| `Validate examples` shard 1/3 — `case96_hidden_friend_removed` | Base branch | Reproduced on a pristine `origin/main` worktree, gcc **and** clang |
| `test-action` timeout / cancelled legs | Infrastructure | Queue starvation; superseded heads |
| ~~`test_adr_status_sync_holds`~~ | **Neither** | My local artifact only; passes in CI. Retracted. |

Green on the current head so far: `ai-readiness`, `lint-and-types`, `e2e`, `docs-pr (required)`, `packaging (windows)`, `MSVC + PDB end-to-end`, both `Native PE/Mach-O compare workflows`, `CodeFactor`, `AgentReady`. `unit-tests` on Ubuntu/macOS and `test-action` are still running.

---

napetrov:
## `test-action (required)` on the current head: timed out while its own matrix was still passing

Discharging the commitment in my earlier comment — I said a `test-action` timeout on the *current* head would be the signal worth acting on. It happened on `ee30b0ad`, so here is what I found and what I am doing.

The failure is the same aggregator timeout, not a mirrored result:

```
test-action summary not reported yet -- waiting   (repeated)
##[error]Timed out waiting for test-action summary to complete
```

It waits on the `Test GitHub Action` workflow (run `34683578516`, 24 jobs) for this same head. Checking that run's jobs directly is the part that matters:

| Job | Status |
|---|---|
| `Appcompat: detect breaking change affecting app` | **success** (ran 09:46) |
| `Compare: audit-only dry run` | **success** (ran 09:51) |
| `Verify: action infrastructure with compare mode` | **success** (ran 09:58) |
| `Multiplatform: ubuntu-latest` | still **queued** — created 09:17, unstarted 41 min later |

So the matrix is **passing** where it has run at all; jobs were still *starting* at 09:58, which is exactly when the aggregator gave up. This is queue starvation reaching the gate, not a defect in the action surface — and it is consistent with the rest of this run: jobs created at 08:33Z were still queued 45+ minutes later.

**Action taken: deliberately not re-running yet.** The rules allow one re-run for a job that executed nothing, and this qualifies — but the thing it waits on has not finished, so re-running now would time out again for the same reason and spend the single attempt for nothing. The re-run becomes useful once the matrix drains, and that is when I will use it if the gate has not cleared on its own by then.

Nothing in this PR's diff touches `action/` or `.github/workflows/` (`git diff --name-only origin/main...HEAD -- action/ .github/workflows/` is empty), so it cannot affect either the aggregator or the matrix it waits on.

### Meanwhile, the substantive picture on `ee30b0ad`

Green: `ai-readiness` (including its ADR-061 architecture-contract step, which is what my `debt-no-growth` fix in this head was for), `lint-and-types`, `e2e`, `docs-pr (required)`, `packaging (windows-latest)`, `MSVC + PDB end-to-end`, `Native PE/Mach-O compare workflows` on both macOS and Windows, `CodeFactor`, `AgentReady`, plus Codecov reporting all modified and coverable lines covered.

Every remaining red is traced to `main` with its failing test named — the `case96_hidden_friend_removed` classification question (six runs across both toolchains and three platforms, plus a pristine `origin/main` worktree) and the Windows unit lane's 71 action/workflow-shell tests. Neither overlaps this diff.

---

napetrov:
## `mutmut (detector core)` — mutmut's *clean* run fails, and I cannot attribute it to `main`

Flagging this one differently from the other reds on this PR, because I have **not** been able to call it pre-existing, and it is closer to my diff than anything else here.

### What fails

mutmut aborts during its pre-flight clean run — before any mutant is tested:

```
FAILED tests/test_release_evidence_contract_axis.py::
  TestReleaseEvidenceContractAxisIsPreserved::
  test_rank_and_preservation[kwargs3-7-evidence_contract_error]
AssertionError: ExitDecision(code=8, reasons=(ExitReason.REMOVED_REQUIRED_LIBRARY, ...),
  removed_required_library_contribution=8, incomplete_scope_contribution=0, ...)
assert 8 == 7
!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!
Failed to run clean test
ERROR: `mutmut run` exited 1 — the run aborted, so any results on disk are from
       an earlier run, not this one. Not reading them.
ERROR: --run requested but mutmut produced no output ... Failing so the mutation
       gate is not a silent no-op.
```

So the gate never measured a single mutant. Worth saying plainly: this is the gate's own fail-closed design working (`AGENTS.md` — "an unresolved run ... is a *failed measurement*, not zero survivors"), not a survivor count I can shrink.

### Why I took it seriously rather than filing it with the others

Per `AGENTS.md`, the mutation PR lane auto-runs when a diff touches a mutated module **or that module's own tests**, and gates `--diff-scoped` absolutely — no baseline to be under. This PR edits `tests/test_policy_override_matrix.py` and `tests/test_finding_identity.py`, and `[tool.mutmut].only_mutate` covers `checker_policy` and identity. I removed 1,608 matrix cases and four tautologies from exactly those files, so "my removals stopped killing mutants" was the obvious hypothesis and had to be ruled out, not assumed away.

### What I established

| Check | Result |
|---|---|
| Is `test_release_evidence_contract_axis.py` in this PR's diff? | **No** |
| Does the whole file pass on this head locally? | **Yes** — 26 passed |
| Does the failing parametrisation pass on a clean `origin/main` worktree? | **Yes** — 5 passed |
| Is it in mutmut's deselect list? | No |
| Could my one `pyproject.toml` change cause it? | **No.** The `--deselect` I removed named `test_main_returns_zero_on_clean_tree`, which no longer exists on this branch (`grep` → 0), so keeping or dropping that line selects an identical test set. It cannot alter another test's `ExitDecision` ranking. |

So it is not a survivor from my removed tests, and it reproduces in **no** ordinary run — only inside mutmut's `mutants/` tree.

### What I cannot yet explain, stated rather than glossed

Other open PRs' `Mutation testing` runs are succeeding (#1246, #1242 both green within the last half-hour), so I will **not** claim this is simply broken on `main` — that would be the same unfounded generalisation I made earlier on this PR about `adr-status-sync` and had to retract. Two candidate explanations I have not distinguished:

1. **Base skew.** The failing assertion is in release exit-axis ranking (`REMOVED_REQUIRED_LIBRARY` = 8 outranking `evidence_contract_error` = 7) — precisely the area `main` changed in ADR-070/071 at `324815f7`. This branch merged that; the green mutation runs may be on branches that have not.
2. **A stale or partial `mutants/` tree.** The run's own message says results on disk are from an earlier run, and the log shows warnings emitted from `mutants/abicheck/product_baseline.py`. A `mutants/` copy whose `abicheck/` predates ADR-070/071 while `tests/` is current would produce exactly this disagreement.

Distinguishing those needs someone who can inspect the runner's `mutants/` tree or re-run the lane on a branch that definitively includes `324815f7`; I can reproduce it in no local configuration.

**No fix pushed, because I have nothing to fix**: the test is not mine, passes everywhere outside this one environment, and the one config line I touched provably has no effect on it. If the answer is (1), the fix belongs with whoever owns the ADR-070/071 exit-axis ranking, not in a test-suite-cost PR. Keeping the PR watched.

---

napetrov:
## New failure cause on `integration-tests (ubuntu-24.04)`: two release-assurance CLI tests — reproduced on this PR's own base commit

The required gate cleared, so this is the one thing left worth reporting. `test-action (required)` is now **green** on `ee30b0ad` — the single re-run I had been holding back until its matrix drained is what cleared it, exactly as I said I would spend it. That budget is now gone.

The three integration legs that just went red are *not* all the same failure I documented before. macOS and Windows are only `case96_hidden_friend_removed` again, byte-identical, already covered twice above — no second comment for those. **Ubuntu has two additional failures I had not seen**, so they get their own account:

```
FAILED tests/test_release_assurance_cli.py::TestReleaseAssuranceThroughTheCli::
  test_a_single_member_shortfall_is_not_masked_by_complete_siblings
  AssertionError: assert 'complete' != 'complete'

FAILED tests/test_release_assurance_cli.py::TestReleaseAssuranceThroughTheCli::
  test_the_notice_never_claims_a_floor_beside_a_real_break
  assert 'Analysis assurance incomplete' in ''
```

### Reproduced on a pristine worktree of this PR's base

Not inferred from file overlap this time — run directly against `origin/main` with nothing from this branch applied:

```
$ git worktree add --detach /tmp/mainwt2 origin/main     # 324815f7
$ PYTHONPATH=/tmp/mainwt2 pytest tests/test_release_assurance_cli.py -m integration \
    -k "single_member_shortfall or never_claims_a_floor"
2 failed, 11 deselected
```

Both fail there with the same messages. `324815f7` is precisely the merge-base this branch sits on, so the failure predates every commit here.

### Why it cannot be this PR's

`git diff --stat origin/main...HEAD -- abicheck/` is **empty** — this branch changes no product code whatsoever. The two assertions are about the release fan-out's assurance fold (`analysis_assurance.status` and the "Analysis assurance incomplete" stderr notice), which lives entirely in `abicheck/`. `tests/test_release_assurance_cli.py` is not in this PR's 24 changed files either.

Reading the assertions substantively rather than just clearing myself: the first says a release whose headers cover two of three members must **not** report `status: "complete"`, and it currently does — a complete sibling masking a single member's shortfall, which is the exact failure mode its own docstring names. The second says the incomplete notice must still be emitted beside a real break (exit 4), and stderr is empty. Both are one-directional: assurance reading *stronger* than the evidence supports. That is the "weaker evidence narrows conclusions" invariant in `AGENTS.md`, so it is worth someone's attention on its own merits, independent of this PR.

**No fix ported, because none exists to port.** Nothing in `origin/main`'s history carries a fix for this, and there is no open entry for it in `docs/contribute/known-gaps.md`. Changing how the release fan-out folds member assurance from inside a test-suite-cost PR would be a product-behaviour change with no ADR behind it — it needs whoever owns that fold to decide what `complete` should mean when one member is short.

**Not re-run.** It reproduces deterministically on the base commit, so it is not a flake, and the one re-run available on this PR was already spent clearing `test-action`.

### Where the PR stands on `ee30b0ad`

Green: `test-action (required)`, `docs-pr (required)`, `ai-readiness`, `lint-and-types`, `unit-tests` on ubuntu 3.12/3.13/3.14 and macOS 3.13, `libabigail-parity`, `abicc-parity`, `heavy-parity-gate`, `e2e`, both `packaging`, `FAIR metadata`, `MSVC + PDB`, both `Native PE/Mach-O`, `CodeQL`, `Security Scan`, `CodeFactor`, Codecov (all modified and coverable lines covered).

Every red now has a named failing test traced to a base commit: `case96_hidden_friend_removed`, the two above, the Windows unit lane's 71 action/workflow-shell tests, and the `mutmut` clean-run abort whose two candidate explanations I set out earlier and still cannot distinguish without runner visibility. Coverage measured at **96.22%** against `main`'s 96.21% on the canonical lane's exact command, so removing test definitions did not cost executed lines.

Keeping the PR watched.

---

napetrov:
## `unit-tests (windows-latest, 3.13)`: the 71 failures are the documented set, but there are also **12 collection errors** I had not recorded

The 71 failures land in exactly the seven files I traced to `main` earlier, so nothing new there. But the same run also reports `12 errors`, which my earlier account of this lane did not mention, so they get a diagnosis rather than being folded into "already covered".

```
= 71 failed, 41884 passed, 1200 skipped, 4 xfailed, 333 warnings, 12 errors in 1040.70s =
```

All twelve are collection errors in three files, four each:

```
________ ERROR collecting tests/test_workflow_concurrency_grouping.py _________
tests\test_workflow_concurrency_grouping.py:125: in _cancelling_workflows
    doc = yaml.safe_load(path.read_text())
...\encodings\cp1252.py:23: in decode
E   UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 67453
```

### Root cause

`Path.read_text()` with no `encoding=`. On Linux and macOS the default locale encoding is UTF-8 and it works; on Windows it is cp1252, which cannot decode the UTF-8 em dashes (`0xe2 0x80 0x94`) and section signs (`0xc2 0xa7`) that **every** file in `.github/workflows/` contains — I checked all 25, and each one carries non-ASCII bytes. So the read fails for any workflow the scan reaches, and because the call sits at module scope the whole file errors during collection instead of failing one test.

Six such call sites across the three files:

```
tests/test_workflow_concurrency_grouping.py:125          yaml.safe_load(path.read_text())
tests/test_workflow_path_filter_closure.py:125           _executed_files(action_yml.read_text())
tests/test_workflow_path_filter_closure.py:132           text = path.read_text()
tests/test_workflow_untrusted_text_interpolation.py:91    yaml.safe_load(path.read_text())
tests/test_workflow_untrusted_text_interpolation.py:171   body.read_text() if body.exists() else None
tests/test_workflow_untrusted_text_interpolation.py:179   yaml.safe_load((... / "bugfix-test-contract.yml").read_text())
```

The fix is `read_text(encoding="utf-8")` at each.

### Why it is not this PR's, verifiable from the tree rather than from a log

All three files are present on `origin/main` (`git ls-tree origin/main` lists each), added there today by `ebc32d0e`, `8c390537` and `a9144995` — none of them mine — and they carry the unqualified `read_text()` on that commit. This PR's diff contains none of the three, and changes no workflow file either (`git diff --name-only origin/main...HEAD -- .github/workflows/` is empty), so it can affect neither the reader nor the bytes being read.

**No fix pushed, and I want to be explicit about the judgement rather than hide behind scope.** This one is tempting: it is in `tests/`, which is this PR's whole subject, and it is six one-word edits. I am leaving it out because the three files belong to three commits that landed hours ago and are still settling, the lane stays red either way on the 71 unrelated action/workflow-shell failures, and the honest general fix is not six edits but a gate — no unqualified `read_text`/`write_text`/`open` anywhere under `tests/` and `scripts/` — which is a repo-wide change with its own consequences and does not belong bolted onto a test-cost PR without the owner's agreement. Per `AGENTS.md` I would rather record the class here, with the root cause and the exact call sites, than ship a narrow patch inside an unrelated diff. Happy to do either version on request.

**Not re-run:** deterministic, platform-derived, and the single re-run on this PR was already spent clearing `test-action (required)` — which is now green.

### Final state of this PR's CI on `ee30b0ad`

All jobs have now reported. Every red has a named failing test and an established origin outside this diff:

| Red | Cause | Origin established by |
|---|---|---|
| `Validate examples` shard 1/3 (gcc + clang), `integration-tests` ×3 | `case96_hidden_friend_removed`, `expected='API_BREAK' got='BREAKING'` | Pristine `origin/main` worktree, both toolchains |
| `integration-tests (ubuntu-24.04)` | 2 × `test_release_assurance_cli.py` | Pristine worktree of the merge-base `324815f7` |
| `unit-tests (windows-latest)` | 71 action/workflow-shell failures | Reproduce on `main` `851b1bd4`, zero diff overlap |
| `unit-tests (windows-latest)` | 12 collection errors | This comment |
| `mutmut (detector core)` | Clean pre-flight run aborts | Two candidate explanations set out earlier; needs runner visibility I do not have |

Green: `test-action (required)`, `docs-pr (required)`, `ai-readiness`, `lint-and-types`, `unit-tests` on ubuntu 3.12/3.13/3.14 and macOS 3.13, `libabigail-parity`, `abicc-parity`, `heavy-parity-gate`, `e2e`, both `packaging`, `FAIR metadata`, `MSVC + PDB`, both `Native PE/Mach-O`, `CodeQL`, `Security Scan`, `CodeFactor`, Codecov. Coverage **96.22%** against `main`'s 96.21% on the canonical lane's exact command, and `check_architecture.py` reports 0 errors.

Nothing of mine is outstanding. Staying subscribed.

---

PR #1245, opened 2026-09-12
perf+fix: memoize the snapshot content digest; stop `declaration_identity_reconciled` over-claiming

napetrov:
## Summary

Two independent defects from the v19 oneAPI scan report: the ~1.7-1.8x graph-compare performance regression (P1), and `OUTCOME_RECONCILED` making a claim its inputs cannot support (P2). P5 is answered below and recorded as a `known-gaps.md` entry with the measurement behind it; P3 and P4 are **not** in this PR — see "Not in this PR".

## Motivation

**P1.** Every graph-shaped compare regressed ~1.7-1.8x between v18 and v19. cProfile on the self-compare put 404.3s of 620s (65%) inside `serialization.snapshot_content_digest` at `n=6` over two snapshots; v18 was `n=2`/134.6s. The digest is pure, and two consumers reach it three times per run: `workflows.gate.snapshot_identity_digest`, and `storage.snapshot_encode.same_persisted_content` (which serializes both sides), through `checker.compare`'s own assurance attach, the front end's re-attach with the real evidence pack, and the same-binary-compared fallback.

**P2.** `graph_reconcile_outcome.classify` ended with an unguarded fall-through to `OUTCOME_RECONCILED`, whose catalog `impact` asserts both the qualified name and the declaring-file evidence changed together, at `COMPATIBLE_WITH_RISK`. All 234 of a oneTBB header graph's reconciled calls were `source_decl` pairs whose `qualified_name`, `source_relative` and two-sided `declaring_file` were byte-identical; 13 findings had `old_value == new_value`. #1232 corrected the rendered description for this case and the claim survived in the kind's own `impact` and verdict tier.

## Changes

- **P1** — `abicheck/storage/snapshot_digest_cache.py`: a run-scoped memo for `snapshot_content_digest`, entered by `cli_compare_helpers.run_compare` and `service_compare_pipeline.classify_compare_pair`. `same_persisted_content` short-circuits on object identity.
- **P2** — two new outcomes/`ChangeKind`s: `declaration_identity_unchanged` (COMPATIBLE) and `declaration_identity_reconciled_unresolved` (COMPATIBLE_WITH_RISK). `declaration_identity_reconciled` narrowed to the combined rename-and-move; `diff_graph_reconciliation_findings` indexes its outcome→kind map instead of defaulting to the strongest kind; #1232's now-unreachable render-time prose override removed. ADR-048 second amendment, catalog case197's expected kind, detector-spec and count docs regenerated. Both kinds registered in every completeness registry the full unit lane checks (`ASSERTED_CHANGE_KINDS`, `scripts/evidence_tiers.py`, `tests/evidence_provenance_contract.py`, `tests/canonical_identity_contract.py`, and — for the new storage module — `tests/unit/storage/adr062_scope.py`).
- **P5** — recorded in `docs/contribute/known-gaps.md` with its measurement.
- Bug classes moved to themed `tests/regressions/manifest_classification.py` / `manifest_performance.py` siblings, to keep `manifest.py` under its `architecture/debt.yaml` no-growth baseline (that file's own precedent: move responsibility out, never raise the baseline).

### Why these choices (P1)

- **Not `id()`-keyed alone.** The cache holds a *strong* reference to each snapshot it answered for, so id reuse across a GC is unrepresentable rather than merely unlikely. Within a run that costs nothing — the run holds them anyway.
- **Not a lazily-populated field on the DTO.** `AbiSnapshot` is a plain mutable dataclass and the resolve half of a comparison genuinely mutates it (evidence backfill, provenance tagging, pack capping). A field would have to be invalidated by every one of those writers — the same bug with more places to forget it. A scope is only ever entered where the snapshots are already final, and outside one the function behaves exactly as before.
- **Not a structural `==` pre-filter.** `__eq__` is sound in the *accepting* direction but not the rejecting one: two unequal snapshots can persist identically (a rounded float, a derived id — the exact class of the four P2s that followed #1229). It would answer only the case object identity already answers while paying a deep comparison on every case it cannot. Object identity is kept; `==` is not.
- **Both consumers already share one digest provider** — the duplication was never two implementations, it was one function called six times.

**Correctness constraints held.** `same_persisted_content` answers exactly what it answered (equal persisted content; the added `old is new` short-circuit could only ever have returned `True`, and `test_same_persisted_content_short_circuits_on_object_identity` asserts it answers without serializing anything). The digest string is byte-identical — still a plain sha256 over the same canonical JSON — asserted by `test_digest_string_is_unchanged_by_memoization`.

### Why COMPATIBLE for `declaration_identity_unchanged` (P2)

It is strictly *less* of a change than `declaration_coordinates_shifted`, which at least has a `:line:col` shift to report and is already argued down to COMPATIBLE for that reason: scoring at risk would inflate `risk_changes` with pairs the reconciliation itself proves identical. It is restricted to type-shaped kinds, and that restriction is load-bearing rather than inherited: for a `source_decl` the classifier's dimensions are *not* exhaustive, so "no dimension I read differs" does not entail "nothing differs" — a pair only reaches reconciliation because its node identities were unequal. Catalog case197 is exactly that (equal name, equal header, a parameter-type change that moved the mangled name), and calling it "unchanged" would downgrade a real difference on evidence that never looked at it. `declaration_identity_reconciled_unresolved` therefore keeps RISK.

## Bug-fix test contract

- Bug class: `perf.pure_content_digest_recomputed_per_consumer` and `classification.default_branch_asserts_more_than_inputs` (both added to `tests/regressions/`)
- Publicly observable failure: (P1) a `compare` of two snapshots serialized each snapshot three times, costing 65% of wall time on a real oneAPI operand; (P2) `abicheck compare` reported `declaration_identity_reconciled` at severity risk — "both the qualified name and the declaring-file evidence changed together" — for 234 pairs where neither did
- Regression test fails on base: yes — `tests/test_snapshot_digest_recomputation.py` 11 of 13 fail with the scope neutralized; `tests/test_graph_reconcile_outcome_properties.py::test_a_claim_of_change_is_entailed_by_the_inputs` fails on 14 of 96 generated pairs against the old classifier
- Negative control: `test_consumers_outnumber_serializations` (the bound would also hold in a build where the digest was never consulted); `test_every_outcome_in_the_vocabulary_is_reachable` (narrowing `OUTCOME_RECONCILED` must not narrow it out of existence)
- Public-surface test: yes — P1 runs the real `abicheck compare` CLI via `CliRunner` and the typed `run_compare_request` API; P2 is covered end-to-end by catalog case197 through `tests/test_l3l4l5_examples.py`
- Axes covered: P1 — front end (cli, typed_api) x pairing (content-identical, genuinely-different) x fixture; P2 — name evidence (identical, coordinates-only, renamed) x location evidence (identical, differing, absent-both, absent-one) x kind (type, source_decl) x signature tail (same, differing), 96 pairs enumerated
- General invariant: (P1) actual serializations counted during a real comparison are bounded by the number of *distinct snapshots*, not the number of consumers — so a fifth caller cannot reintroduce the defect; (P2) an outcome whose prose claims a name or location change may only be returned where the raw `qualified_name`s or `source_relative`s actually differ, checked against an oracle built from the raw inputs rather than from the classifier's own normalization helpers

## Not in this PR

- **P3 (dump-side edge ordering nondeterminism)** — not attempted. It needs the oneDPL corpus to reproduce the two-dump diff and to answer which of the implicit-ctor answers (`+3` vs `+1`) is correct; neither is available in this environment.
- **P4 (removed-option migration hints)** — mechanism designed, not landed. Probing the live CLI found the removed set is far larger than the flags the report named (`--require-complete-analysis`, `--env-matrix`, `dump --build-target`, `--exit-code-scheme`, `--header-graph`, `--surface-metrics`, `--pattern-verdicts`, `--crosscheck`, `--explain-patterns`, `--show-filtered`, `--audit-suppressions`, `--demangle`, `--profile`, … — `compare` is down to 50 options). The right shape is a `REMOVED_OPTIONS` registry beside `frontends/cli/options/rulings.py` (same curated-registry-plus-exact-gate pattern) with a hint injected at `_AbicheckGroup.main`'s single `ClickException` choke point, so the per-flag assertion list the report warns about never exists. Landing it needs a per-flag replacement audit across ~40 removals that is its own PR.
- **P5** — answered, no code change. The prior question ("what do consumers actually do with `quality_issues`?") outranks the presentation one, and the answer is a real defect: `quality_issues` is one of the four `SeverityConfig` categories and `PRESET_STRICT` sets every category to `ERROR`. Measured directly, `declaration_coordinates_shifted` and `declaration_identity_unchanged` each give `default rc: 0, strict rc: 1` — a finding whose whole content is "this declaration did not change" gates the build. `compute_gate_decision` also names the category in `blocking_categories`, and `pr_comment_render` renders it as "⛔ Quality policy violation". The fix is a third `COMPATIBLE` sub-partition (`INFORMATIONAL_KINDS`) with no `SeverityConfig` category of its own — a `checker_policy` partition change, a `report/` `compute_*`/`render_*` pair change, a report-schema bump and a `changekind-partition` gate update — not a renderer special case and not a per-kind gate exemption. Recorded in `known-gaps.md` with the measurement; the decision is left open rather than taken here.

## Verification

As of head `5d09dabb`:

- **CI: 42 checks green, 3 skipped, 0 failures**, remainder still running.  Green so far include `ai-readiness`, `test-contract`, `docs-pr (required)`, `e2e`, `build-docs`, `changelog-fragment`, `CodeQL Analysis (python)`.
- **Local `verify.py --profile pr`: 21 of 22 steps pass.** The exception is `fmt-check`, which is red on the base commit too (`btf_metadata.py`, `ctf_metadata.py`, `elf_metadata.py` are flagged and none is in this diff); this PR's own files are formatted with the pinned ruff.
- **Full Linux unit suite: 44,944 passed**, 43 skipped, 4 xfailed. One failure, unrelated and invisible to CI: `test_agent_evals.py::test_manifest_base_commit_exists_in_history[add-change-kind-small]`, whose manifest pins a `base_commit` absent from the repository; it `pytest.skip`s on a shallow checkout, which is `actions/checkout`'s default.
- `python scripts/check_ai_readiness.py` — **0 errors**. `ARCHITECTURE_BASE=324815f7 python scripts/check_architecture.py` — **0 errors**. `ruff check` and `mypy abicheck/` clean.
- **P1 measured through the real CLI and typed API: 6 serializations → 2**, for 2 distinct snapshots — restoring v18's `n=2`.
- Codecov patch coverage **94.74%** against a 90% target once the canonical lane's upload landed; `snapshot_digest_cache.py` subsequently taken to **100%** line and branch.
- The oneTBB / self-compare wall-time re-run the report asks for could not be performed here — the oneAPI corpus is not in this environment. The serialization count is the mechanism the profile attributed 404s to, restored to its v18 value.

### Red checks that are not this PR's

Verified against `main` @ `324815f7` (this PR's base) — identical node IDs, identical messages, identical pass/skip counts; the only numeric difference anywhere is `deselected`, which rises because this PR adds tests those marker lanes deselect. Two comments below carry the detail.

| root cause | checks affected | on `main`? |
|---|---|---|
| `case96_hidden_friend_removed` (`expected='API_BREAK' got='BREAKING'`) | integration-tests ×3 OSes, Validate examples (clang + gcc) | yes, identical |
| WSL-`bash` unavailable + cp1252 decode on the Windows runner | unit-tests (windows-latest, 3.13) | yes, identical |
| `test_release_assurance_cli.py` ×2 | integration-tests (ubuntu-24.04) | yes, identical |

Earlier `test-action summary` failures were on superseded SHAs, with every matrix job `cancelled` by a subsequent push of mine; in the last such run 7 Action jobs did complete and all 7 succeeded.

**Corrections to earlier revisions of this description**, kept rather than silently edited away: (1) it claimed `check_ai_readiness.py` had 1 pre-existing error (ADR-049's `**Verified:**` receipt). That was an artifact of this build container's *shallow* clone — after `git fetch --unshallow` the receipt resolves and the gate is clean; there is no ADR-049 problem. (2) An earlier "ruff format clean" claim used the ruff on `PATH`, not the pinned one `verify.py` invokes; three of this PR's test files were unformatted by the real gate and have been fixed.

## Checklist

- [x] Changelog fragment added (`scriv create`) if this touches `abicheck/**/*.py`
- [x] `verify.py --profile pr` passes except the base-branch-red `fmt-check`


napetrov:
### `unit-tests (windows-latest, 3.13, false)` is red here, and it is red on `main` — not this PR's

Standing down on this one, with the evidence, rather than leaving it unexplained.

**The numbers line up exactly.** This PR's head `f39e8348` and this PR's own base commit produce the identical failure set:

| | failed | passed | errors |
|---|---|---|---|
| this PR (`f39e8348`) | 71 | 43,650 | 12 |
| `main` @ `324815f7` (this PR's base) | 71 | 43,413 | 12 |

Same 71 failures, same 12 collection errors. The `passed` delta is just the tests this PR adds. The job is also `failure` on the three most recent completed `ci.yml` runs on `main` (`324815f7`, `851b1bd4`, `393c9c0f`), so it is not a one-off either.

**Two Windows-only clusters, neither reachable from this diff.**

1. *Action shell-helper tests* — on this runner `bash` resolves to `wsl.exe`, which answers `Windows Subsystem for Linux has no installed distributions.`, so every helper invocation returns empty output and `returncode 1`. Hence `assert '' == 'json'`, `assert '' == 'out.json'`, and friends across `test_action_run_sh_helpers.py`, `test_action_run_sh_compare_pr_json_write.py`, and `test_extra_args_is_value_option_completeness.py`.
2. *12 collection errors* — `UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 67453`, i.e. a workflow file read under Windows' cp1252 default encoding, in `test_workflow_concurrency_grouping.py`, `test_workflow_path_filter_closure.py`, and `test_workflow_untrusted_text_interpolation.py`.

Both are properties of the runner environment and of `action/`/workflow-file reading. This PR touches `storage/snapshot_digest_cache.py`, `buildsource/graph_reconcile*.py`, the change catalog, and test registries — none of the failing modules, and none of the failing tests import anything it changes.

**No fix exists to port.** Neither currently open PR (#1242, #1244) addresses the WSL-`bash` or cp1252 problem, so there is nothing to bring into this branch; a ported fix would be the right move if one appeared, and this note should be read as "still waiting", not "closed".

**Not re-running it.** A re-run is for confirming a suspected one-off, and three consecutive `main` runs with byte-identical failure counts is stronger evidence than a fourth run would be — both causes are deterministic environment facts, not flakes. Nothing here is being skipped, disabled, or quarantined.

For the record, on Linux this PR's full unit suite is green: 44,944 passed, with one unrelated failure (`test_agent_evals.py::test_manifest_base_commit_exists_in_history[add-change-kind-small]`, whose manifest pins a `base_commit` that no longer exists in the repository; that test `pytest.skip`s on a shallow checkout, which is `actions/checkout`'s default, so it never runs in CI at all).

---

napetrov:
### Addendum: `integration-tests (macos-latest)` is the same story

A second red check, same conclusion, checked the same way.

| | failed | passed | skipped | deselected |
|---|---|---|---|---|
| this PR (`f39e8348`) | 1 | 244 | 402 | 45,279 |
| `main` @ `324815f7` (this PR's base) | 1 | 244 | 402 | 45,042 |

One failure on each side, and it is the *same* test with the *same* message:

```
FAILED tests/test_example_autodiscovery.py::test_example_source_smoke[case96_hidden_friend_removed]
AssertionError: case96_hidden_friend_removed: source-smoke validation FAIL
  — expected='API_BREAK' got='BREAKING'
```

No new failure from this PR; the only numeric difference is `deselected`, which rises because this PR adds tests that this lane does not select. `integration-tests` is also red on `ubuntu-24.04` and `windows-latest` on that same `main` run, so the whole lane is currently red on the base branch.

Same disposition as the Windows lane above: nothing to port (no open PR addresses it), nothing skipped or quarantined, and no re-run — the failure is a deterministic verdict mismatch (`API_BREAK` vs `BREAKING`) reproducing identically on `main`, not a flake.

---

napetrov:
### `Baseline regression (PR vs base)`: measured against base locally, no regression — re-run triggered

```
PERFORMANCE GATE FAILED:
  - opaque_filter @ size=1000: 1.037s vs baseline 0.836s
    (+24%, +0.201s; allowed +0.125s = max(15%, 0.100s))
```

This one is about *this PR* by construction, so "it's red on main too" is not available as an answer. I reproduced the comparison instead, using the same method the job uses — HEAD's own `scripts/benchmark_scaling.py`, run against a base checkout (`324815f7`) and this head in turn, `--scenario opaque_filter --sizes 1000 --repeat 5 --no-memory`, on one machine, three paired rounds:

| round | base | head | delta |
|---|---|---|---|
| 1 | 1.162s | 1.155s | −0.6% |
| 2 | 1.156s | 1.159s | +0.3% |
| 3 | 1.210s | 1.174s | −3.0% |

Head is marginally **faster** on median, and every difference sits inside the runs' own 1.6–3.0% coefficient of variation. Nothing resembling +24%.

**There is also a mechanism argument, which matters more than the numbers.** This scenario times `checker.compare` directly, not `run_compare`/`classify_compare_pair` — and those two are the only places a `digest_scope` is ever opened. So on this code path no scope exists, `memoized_digest` takes one `ContextVar.get()`, sees `None`, and calls straight through to the same computation base performs. Per compare that is two extra context-variable lookups, against a ~1.16s measurement. The memo cannot make this scenario slower, and on this path it does not make it faster either — the scenario simply does not exercise the thing this PR changes.

Note also that base's own absolute number differs a lot between environments (0.836s in CI vs 1.16s here), which is consistent with the gate reading sub-second medians on shared runners; the workflow's own comment anticipates this and says to loosen `--regress-tolerance`/`--regress-min-delta-seconds` rather than reinstate `continue-on-error` if variance proves too noisy. I am not proposing that change here — it is a gate-policy decision for a maintainer, not something to slip into an unrelated PR.

I have used the one permitted re-run on this job. If it fails a second time, I will treat it as real and investigate rather than re-run again.

---

napetrov:
### `mutmut (detector core)`: a pre-existing latent test bug, not a surviving mutant — diagnosis + verified one-line patch

The gate's own message is misleading here, so the headline first: **no mutant survived, and no mutant was ever generated.** mutmut installed fine (`mutmut-3.7.0`) and aborted during its *clean-test baseline* — the unmutated sanity pass it runs before mutating:

```
FAILED tests/test_release_evidence_contract_axis.py::TestReleaseEvidenceContractAxisIsPreserved
       ::test_rank_and_preservation[kwargs3-7-evidence_contract_error]
AssertionError: ExitDecision(code=8, reasons=(REMOVED_REQUIRED_LIBRARY,) ...)  assert 8 == 7
Failed to run clean test

ERROR: --run requested but mutmut produced no output (not installed / could not start).
```

Nothing in this PR touches that test or `resolve_release_exit_decision`.

#### Root cause

`test_rank_and_preservation`'s parametrize cases are module-level dict **literals**, created once at collection — and the test body mutates them:

```python
@pytest.mark.parametrize(("kwargs", "expected_code", "expected_reason"), (
    ...
    ({"removed_required_library": True, "verdict_or_severity_contribution": 4}, 7, "evidence_contract_error"),
    ...
))
def test_rank_and_preservation(self, kwargs, expected_code, expected_reason):
    decision = resolve_release_exit_decision(
        not_comparable=kwargs.pop("not_comparable", False),          # <-- pop
        severity_scheme_active=kwargs.pop("severity_scheme_active", False),
        verdict_or_severity_contribution=kwargs.pop("verdict_or_severity_contribution", 0),
```

One run per process is fine. A **second** run in the same process sees `kwargs3` already stripped: `verdict_or_severity_contribution` is gone, so it defaults to `0`, `removed_required_library=True` is then the active axis, and the decision is `8`/`removed_required_library` instead of `7`. Hence `assert 8 == 7` — on exactly the one case whose expectation depends on that key being present.

mutmut 3.x re-runs the suite in-process, which is why this surfaces there and nowhere else. It is invisible to every ordinary lane, which run the suite once.

#### Reproduced on base, not just here

Running that node twice in one process:

| tree | pass 1 | pass 2 |
|---|---|---|
| this PR (`5d09dabb`) | 5 passed | **2 failed** (`assert 8 == 7`) |
| `main` @ `324815f7` | 5 passed | **2 failed** (identical) |

Identical on base. The bug predates this PR entirely.

#### Verified patch (one line)

```diff
         """Which code decides, and that the axis is reported either way."""
+        kwargs = dict(kwargs)
         decision = resolve_release_exit_decision(
```

Applied to a clean `324815f7` checkout, the same two-pass run gives `5 passed` / `5 passed`. (`_decide()` higher in the file is already safe — its `**kwargs` is a fresh dict per call.)

#### Why I have not pushed it

It is unrelated to this PR's subject, so fixing it here would widen the diff into someone else's area; the guidance I follow says to state the failure with a proposed patch rather than widen. Say the word and I will apply it here, or it can go in as its own one-line PR — it would unblock the mutation lane for any branch that touches a mutated module, not just this one.

Worth noting the class, since this repo keeps a bug-class registry: *a parametrize case that the test body mutates is shared state across runs*. A grep for `kwargs.pop(`/`.pop(` on a parametrized argument would likely find siblings, and none of them can fail in a normal single-pass run.

No re-run: this is deterministic, and re-running would abort at the same clean test.

---

PR #1247, opened 2026-09-12
test: stop `test_rank_and_preservation` mutating its own parametrize cases

napetrov:
## Summary

One line, plus a comment explaining it. Unblocks the `mutmut (detector core)` lane, which currently generates **zero mutants** on any branch whose diff touches a mutated module.

## Motivation

`tests/test_release_evidence_contract_axis.py::TestReleaseEvidenceContractAxisIsPreserved::test_rank_and_preservation` has its cases declared as module-level dict **literals**, created once at collection — and the test body `.pop()`s three keys out of the case it is handed:

```python
@pytest.mark.parametrize(("kwargs", "expected_code", "expected_reason"), (
    ...
    ({"removed_required_library": True, "verdict_or_severity_contribution": 4}, 7, "evidence_contract_error"),
    ...
))
def test_rank_and_preservation(self, kwargs, expected_code, expected_reason):
    decision = resolve_release_exit_decision(
        not_comparable=kwargs.pop("not_comparable", False),
        severity_scheme_active=kwargs.pop("severity_scheme_active", False),
        verdict_or_severity_contribution=kwargs.pop("verdict_or_severity_contribution", 0),
```

One run per process is fine, which is why every ordinary lane is green. A **second** run in the same process sees the already-stripped case: `verdict_or_severity_contribution` is gone, defaults to `0`, `removed_required_library=True` becomes the active axis, and the decision comes back `8`/`removed_required_library` instead of `7`/`evidence_contract_error` — `assert 8 == 7`. It is precisely the one case whose expectation depends on that key still being present.

**mutmut 3.x re-runs the suite in-process**, so it is the only thing that hits this. The failure lands in mutmut's *clean-test baseline* — the unmutated sanity pass before mutation — so the lane produces no mutants at all and reports:

```
Failed to run clean test
ERROR: --run requested but mutmut produced no output (not installed / could not start).
```

That wording is misleading: mutmut installed (`mutmut-3.7.0`) and started fine, and refused to proceed past a red clean test. The gate failing closed here is correct behaviour — it is the underlying test that is wrong.

Found while triaging #1245, whose diff touches neither this test nor `resolve_release_exit_decision`. Reproduced identically on this commit's own parent, so it is not that PR's.

## Changes

```diff
         """Which code decides, and that the axis is reported either way."""
+        # The parametrize cases above are module-level dict *literals*,
+        # created once at collection -- and the calls below `.pop()` from
+        # this argument. One run per process is fine; a second run in the
+        # same process sees an already-stripped case. Copy before popping.
+        kwargs = dict(kwargs)
         decision = resolve_release_exit_decision(
```

`_decide()` higher in the same file is already safe — its `**kwargs` is a fresh dict per call, so nothing else needs changing.

## Verification

Running the node twice in one process (`pytest.main` called twice, module state persisting):

| | pass 1 | pass 2 |
|---|---|---|
| before | 5 passed | **2 failed** (`assert 8 == 7`) |
| after | 5 passed | 5 passed |

Also: the whole file single-pass is `26 passed`; `ruff check` clean; pinned `ruff format --check` reports it already formatted.

No changelog fragment — this touches only `tests/`, not `abicheck/**/*.py`.

## Note on the class

Worth a `tests/regressions/manifest.py` entry if a maintainer agrees the class is real: **a parametrize case that the test body mutates is shared state across runs in a process.** It cannot fail in a normal single-pass run, so no ordinary lane can catch it, and a `grep` for `.pop(` on a parametrized argument may well find siblings. I have not added a registry entry here, to keep this PR to the one line.

## Checklist

- [x] Changelog fragment — n/a, no `abicheck/**/*.py` change
- [x] Docs updated if user-visible behaviour changed — n/a

---
_Generated by [Claude Code](https://claude.ai/code/session_0173q5kUcC63nkVmJao2AGjV)_


napetrov:
### CI status: 5 inherited failures, and why this PR can't demonstrate its own fix in CI

**The mutation lane was skipped here, not run.** `mutmut (detector core)` shows `conclusion: skipped`, zero duration — `resolve` decided `run=false`. That is the path filter working as designed: it fires on a diff touching a mutated module *or that module's own tests*, and this diff touches `tests/test_release_evidence_contract_axis.py`, which is neither. So the one lane this PR exists to unblock never executed on it.

**I considered the `mutation` label and decided against it.** `mutation.yml` is explicit that the label is the documented override for exactly this case — but it also sets `require_baseline=true`:

```
#   - the `mutation` label: it is documented as the complete check,
#     and a label-forced run on a test-only diff outside the globs
#     above would otherwise print "gated nothing" and exit 0 ...
if [ "$MATCHED_TESTS" = "true" ] || [ "$LABELLED" = "true" ]; then
  echo "require_baseline=true" >> "$GITHUB_OUTPUT"
```

There is no `mutation-baseline.json` on `main` (AGENTS.md says as much: *"not recorded yet — until a dispatch run writes it, the drift lanes fail closed rather than passing vacuously"*). A label-forced run would therefore fail on the missing baseline, turning this PR red for a reason that has nothing to do with the change. The label also does not exist in this repository yet, so using it would mean creating it — not my call on someone else's gate.

So the verification for this fix remains the local two-pass reproduction in the PR description: `5 passed / 2 failed` before, `5 passed / 5 passed` after, and the same failure reproduced on the parent commit. The real proof will be the next PR that legitimately triggers the lane — that run should get past mutmut's clean test and actually generate mutants, where today it generates zero.

### The 5 red checks are inherited

| check | root cause | on `main`? |
|---|---|---|
| `unit-tests (windows-latest, 3.13, false)` | WSL-`bash` unavailable + cp1252 decode | yes, identical |
| `integration-tests (macos-latest)` | `case96_hidden_friend_removed` (`expected='API_BREAK' got='BREAKING'`) | yes, identical |
| `Validate examples (clang, debug-headers, shard 1/3)` | same `case96` | yes, identical |
| `Validate examples (gcc, debug-headers, shard 1/3)` | same `case96` | yes, identical |
| `Full example matrix` | downstream aggregate of the two shards above | yes, identical |

I re-verified the clang shard directly against this head rather than assuming: `1 failed, 52 passed, 19 skipped, 4 xfailed`, the single `case96` failure, the same two pre-existing `KINDS_MISMATCH` advisories (`case16_inline_to_non_inline`, `case191_header_only_graph_field_type`), and zero occurrences of anything from the recently-merged #1245 (`case197` is SKIPPED in that lane; `declaration_identity*` appears nowhere).

This PR changes one test file, adding a dict copy. It cannot reach Windows shell behaviour, file encodings, or example fixtures. No fix exists to port for either root cause, and no re-run is warranted — both are deterministic and reproduce on `main`.

---

napetrov:
### Final CI tally (correcting the count above)

CI has finished on `f6d5ab45`: **43 success, 7 failures, 8 skipped**, nothing still running.

My comment above said 5 inherited failures; it was written while two integration lanes were still in flight. The correct number is **7** — the same two root causes, two more lanes:

| check | root cause |
|---|---|
| `unit-tests (windows-latest, 3.13, false)` | WSL-`bash` unavailable + cp1252 decode |
| `integration-tests (macos-latest)` | `case96_hidden_friend_removed` |
| `integration-tests (ubuntu-24.04)` | `case96` + 2 × `test_release_assurance_cli.py` |
| `integration-tests (windows-latest)` | `case96` |
| `Validate examples (clang, debug-headers, shard 1/3)` | `case96` |
| `Validate examples (gcc, debug-headers, shard 1/3)` | `case96` |
| `Full example matrix` | downstream aggregate of the two shards |

All verified identical on `main`. I checked two of these directly against *this* head rather than inheriting the earlier conclusion:

- **`integration-tests (ubuntu-24.04)`** — `3 failed, 540 passed, 103 skipped, 5 xfailed`, exactly the three known node IDs with the same messages. `tests/test_release_evidence_contract_axis.py` — the only file this PR touches — is deselected in this lane and appears nowhere in the failures.
- **`Validate examples (clang, shard 1/3)`** — `1 failed, 52 passed, 19 skipped, 4 xfailed`, the single `case96` failure and the same two pre-existing `KINDS_MISMATCH` advisories.

Both also confirm the freshly-merged #1245 is behaving on `main`: `case197_header_graph_identity_reconciled` is SKIPPED in those lanes, `declaration_identity*` appears in no failure, and `test_declaration_identity_correct_for_every_sibling_shape` **passes on both castxml and clang**.

Codex code review completed with no findings; codecov reports all modified lines covered. `mutmut (detector core)` is `skipped` for the path-filter reason in the previous comment, so nothing has changed there.

---

PR #1248, opened 2026-09-12
fix(action): recognize every Click spelling of --output

napetrov:
## Summary

Follow-up to [#1246](https://github.com/abicheck/abicheck/pull/1246): the clustered attached forms of `--output` (`-voPATH`, `-vvoPATH`) were still unrecognized, so a working run reported `REPORT_UNREADABLE`. Plus two accuracy corrections that #1246's own changes made necessary.

This commit was pushed a few minutes after #1246 merged, so it is a new pull request rather than an addition to that one.

## Motivation

#1246 taught `_effective_output_file` about the bare attached form `-oPATH`. Click also accepts the *clustered* attached form, which `_extra_args_expand_short_clusters` deliberately leaves unexpanded — a cluster ending in an attached value consumes no following token, so "unexpanded" is the right answer for every other consumer. The `-o?*` test added for the bare form therefore never saw `-vo/report.json`, and with `format: json` and no `output-file` the Action judged a file-writing run as the stdout shape and published `REPORT_UNREADABLE` for a successful comparison.

Verified against the installed Click parser rather than inferred from the help text: `-vo/report.json` and `-vvo/report.json` both resolve to verbose plus that output path.

## Changes

- **`_attached_output_value`** replaces the inline `-o?*` glob. It strips leading `v`s one at a time rather than matching `-v*o?*`, because that glob also accepts `-vHofoo` — which Click reads as `-v -Hofoo`, a *header* value whose text merely contains an `o`. Naming the function also puts the "which short options exist" knowledge beside `_extra_args_expand_short_clusters`, which already depends on the same fact.
- **Tests state the whole spelling set** (`--output X`, `--output=X`, `-o X`, `-oX`, `-voX`, `-vvoX`, `-vo X`) across three properties each — the report is found and read, the effective path is published as `report-path`, and a missing destination is still `REPORT_UNREADABLE` — since this is the second spelling to be missed.
- **Three contradictory-report diagnostics** no longer claim the report "claims a schema version": the unversioned `mode: bundle_facts` shape can now be contradictory too, so they state only what is established — an `analysis_assurance` block with the contribution it owes omitted.
- **The bug class's `answerability` axis** now names the two states the seed matrix reaches. `admitted_but_unanswerable` is unreachable on correct code — that is what the invariant asserts — and is reached only under mutation, which the axis comment now records.

## Bug-fix test contract

- Bug class: `report.unestablished_result_reads_as_success` (`tests/regressions/manifest_report.py`) — this instance is its inverse direction: a valid run reported as unreadable.
- Publicly observable failure: `extra-args: -vo/report.json` with `format: json` and no `output-file` publishes `verdict: REPORT_UNREADABLE` and exit 1 for a comparison that completed and wrote a valid report.
- Regression test fails on base: yes — `TestEveryOutputSpellingNamesTheDestination`; reverting `_attached_output_value` to the `-o?*` glob fails 4 cases (`-voX`, `-vvoX` × found-and-read, published-as-report-path).
- Negative control: `test_a_missing_destination_is_still_caught` (all seven spellings) — recognizing a spelling must not stop requiring what it names.
- Public-surface test: yes — every case runs the whole of `action/run.sh` through `tests/_action_run_sh_harness.py` with a real `$GITHUB_OUTPUT`, asserting the published `verdict` and `report-path` outputs.
- Axes covered: `requested_destination` (primary output file, stdout, `extra-args --write json=`) across all seven `--output` spellings; `answerability`.
- General invariant: *every spelling Click accepts for `--output` names the effective destination, and nothing else does.* Exercised over the full seven-form set × three properties (21 cases) plus two adversarial non-destination traps, against the installed Click parser as the oracle — not against the same string-matching the implementation uses.
- Malicious fixture + side-effect absence: `test_a_header_value_containing_o_is_not_an_output` and `test_a_header_value_containing_o_with_no_real_output` execute `-vHofoo` through the real `run.sh` (a PR-controlled `extra-args` value, per `action/AGENTS.md`'s threat model) and assert **no `foo`/`ofoo` file was created** alongside the published verdict and `report-path`. The published value alone cannot see a misread that made `foo` the effective destination, where the freshness check would then fingerprint — and on a rewrite trust — a path the caller never named.

One test-quality note worth recording, because mutation-testing my own work caught it: the first version of the `-vHofoo` trap test was **vacuous**. It placed the trap *before* a real `-o`, and the resolver is last-wins, so the real path overwrote the misread value and the assertion passed whether or not the rule was correct — the loose "strip everything before the first `o`" mutation survived it. The trap now comes after the real `-o`, with a second case carrying no `-o` at all so ordering cannot mask a misread; both fail against that mutation.

## Checklist

- [x] Changelog fragment added (`scriv create`) if this touches `abicheck/**/*.py` — extends the existing `20260912_claude_action_report_reader_truthfulness.md` fragment; no `abicheck/**/*.py` change in this PR
- [x] Docs updated if user-visible behaviour changed — `action/AGENTS.md`
- [x] `pre-commit run --all-files` passes locally — `ruff check` clean, `bash -n` clean, `check_architecture.py` 0 errors, `check_ai_readiness.py` reports only the pre-existing ADR-049 `**Verified:**` receipt error that is red on `main` too

Local verification on `47625787`: `-k action` 3856 passed / 0 failed; `bugfix-test-contract` passes locally with this description via `BUGFIX_CONTRACT_BODY_FILE`.

https://claude.ai/code/session_01Y4iSrvcQHGa8A6spobjNAq

### DIFF
--- a/action/AGENTS.md
+++ b/action/AGENTS.md
@@ -310,6 +310,16 @@ Two predicates close that without weakening anything above:
   earlier one. **Add a consumer of the output path and resolve it through
   `_EFFECTIVE_OUTPUT_FILE`**, never `$OUTPUT_FILE`.
 
+  **Every spelling Click accepts for `--output` names the destination**, and
+  `_attached_output_value` is what recognizes the ones `_extra_args_options`
+  leaves opaque: `-oPATH`, `-voPATH`, `-vvoPATH`. It strips leading `v`s one at a
+  time rather than globbing `-v*o?*`, because that glob also accepts `-vHofoo` —
+  which Click reads as `-v -Hofoo`, a *header* value whose text merely contains
+  an `o`. Confirm a new spelling against the installed Click parser before
+  relying on it; the help text is not the parser. Two of these were separate
+  findings, which is why the tests state the whole spelling set rather than the
+  reported input.
+
   **`-oPATH` counts as naming the destination.** `_extra_args_options` leaves an
   *attached* short option an opaque bare token deliberately — that form consumes
   no following token, so "unexpanded" is the right answer for every other
--- a/action/run.sh
+++ b/action/run.sh
@@ -2216,6 +2216,30 @@ _effective_format() {
   printf '%s' "$_found"
 }
 
+# The attached `-o` value in an opaque bare token, or failure when it holds none.
+#
+# The forms Click accepts and this recognizes: `-oPATH`, and a cluster of
+# leading boolean `-v`s before it (`-voPATH`, `-vvoPATH`). Leading `v`s are
+# stripped one at a time rather than matched with `-v*o?*`, because that glob
+# also accepts `-vHofoo` -- which Click reads as `-v -Hofoo`, a *header* value
+# whose text merely happens to contain `o`. `-v` is the only boolean short
+# option across every command this Action invokes, which is the same fact
+# `_extra_args_expand_short_clusters` relies on.
+#
+# Fails (prints nothing, returns 1) for a bare `-o` or `-vo`, whose value is the
+# *next* token: the cluster expander already turns those into a pending option
+# that `_extra_args_options` pairs with its value, so they never arrive here.
+_attached_output_value() {
+  local _token="${1:-}" _rest
+  _rest="${_token#-}"
+  [[ "$_rest" != "$_token" ]] || return 1
+  while [[ "$_rest" == v* ]]; do
+    _rest="${_rest#v}"
+  done
+  [[ "$_rest" == o?* ]] || return 1
+  printf '%s' "${_rest#o}"
+}
+
 # The real `--output` path `abicheck` writes to, accounting for `extra-args`
 # overriding this script's own `-o "$OUTPUT_FILE"` flag -- the exact sibling of
 # `_effective_format` above, and for the same reason: `CMD` carries this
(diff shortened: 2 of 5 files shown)


napetrov:
Standing down on two failing checks that are not this PR's, re-verified against the **new** base `4ef866d2` (main moved when #1246 and #1249 merged, so the equivalent note on #1246 does not carry over).

**1. `ai-readiness` — ADR-049's `**Verified:**` receipt names an unreachable commit.**

On pristine `origin/main@4ef866d2`, `python scripts/check_ai_readiness.py` reports exactly one error, in a file this PR does not touch:

```
=== adr-status-sync ===
  ERROR: docs/contribute/adr/049-contract-relevance-and-compatibility-configuration.md:
  '**Verified:**' names commit 3930cfce, which is not reachable from origin/main
```

`git merge-base --is-ancestor 3930cfce origin/main` is false, so this is the exact failure mode that gate's own message warns about: a receipt anchored to the branch that added it, which vanishes on merge and then fails a required job on `main` permanently. It is unrelated to this PR's diff (`action/`, `tests/`) and no fix for it exists on any branch I can see, so there is nothing to port.

Proposed patch, for whoever owns that ADR — one line, in its own PR:

```diff
-**Verified:** main@3930cfce on 2026-09-09
+**Verified:** main@4ef866d2 on 2026-09-12
```

…after actually re-reading the Status claim against the code, since the same gate also WARNs that 58 commits have touched the modules that Status names since the recorded sha. Bumping the sha without re-reading would trade a hard error for a false receipt, which is worse. Alternatively, drop the `**Verified:**` line — it is optional — until someone does that re-read.

**2. `ruff format --check` — repo-wide formatting drift on the base branch.**

On the same pristine `origin/main@4ef866d2`:

```
617 files would be reformatted, 1405 files already formatted
```

This PR's own four touched files are formatted (`ruff format --check` is clean on them individually). 617 files is a repo-wide `ruff` version/settings drift that needs its own dedicated PR; folding it in here would bury a ~500-line behavioural change under a ~600-file reformat, and a broad `ruff format tests/` in this tree also trips 47 `check_architecture.py` test-size caps, so it is not a mechanical fix either.

Neither is re-run-able flake: both reproduce deterministically on a clean checkout of the base branch with no involvement from this diff. I have merged `origin/main@4ef866d2` into this branch so CI re-runs against the base that #1249 fixed (`test_release_assurance_cli.py`'s two failures, which I stood down on for #1246, are genuinely fixed there and are no longer expected). I will keep driving anything else that comes back red.

---

PR #1263, opened 2026-09-12
refactor: consolidate the three strict YAML manifest loaders

napetrov:
## Summary

One strict, duplicate-key-checking YAML loader (`abicheck/model/yaml_strict.py`) replaces the three private `yaml.SafeLoader` subclasses that had accumulated in `dump_manifest.py`, `compatibility_evaluation_packs.py` and `impact/use_cases.py`.

## Motivation

The three copies had already **diverged**, and the divergences were real defects rather than style:

- Only two of the three checked key hashability, so a sequence/mapping used as a mapping key (`? [a, b]`) raised a raw `TypeError` out of `dump_manifest._load_yaml_strict` instead of `ManifestValidationError` — which `workflows/bundle_facts_library_overrides.py` then had to catch and translate by hand at its own call site.
- `--dump-manifest` never translated PyYAML's implicit-resolver `ValueError` (an out-of-range timestamp-shaped scalar such as `2023-99-99`), so that escaped its documented error contract too.
- The pack loader's own docstring recorded that it was copied from `dump_manifest.py` purely because that one was private.

The cleanup deletes the duplicated *implementation* while keeping the *protection* — plain `yaml.safe_load` silently accepts a repeated key with last-value-wins semantics, which is exactly what these formats must not do.

## Changes

- **New `abicheck/model/yaml_strict.py`** — `load_strict_yaml(text, error=...)` normalizes every malformed-document failure (syntax error, duplicate key, unhashable key, invalid implicit scalar) into the caller's own exception type, so each manifest format keeps its error vocabulary without keeping its parser. `StrictMappingError` carries a precise one-line message (`duplicate key 'a' in the same mapping (line 3)`) rather than PyYAML's multi-line `ConstructorError` dump. A `RecursionError` is deliberately left untranslated: it is a resource limit, and the one caller that surfaces it to a user has its own message for it. Placed in `model/` because it is a dependency-free leaf that `extract`, `policy`, `workflows` and `impact` all need.
- `dump_manifest.py`, `compatibility_evaluation_packs.py`, `impact/use_cases.py` now supply only their own error vocabulary (net −238 / +84 production lines).
- `workflows/bundle_facts_library_overrides.py` calls the shared loader directly instead of importing `dump_manifest`'s private `_load_yaml_strict` across a layer boundary, and its hand-rolled `TypeError` arm (and the docstring paragraph justifying the cross-layer reach) are gone.
- `suppression_yaml.py` is deliberately **unchanged**, with a docstring note saying why: a suppression document's duplicate key is not an error there. That format resolves YAML merge keys (`<<:`) and follows `safe_load`'s last-value-wins semantics on purpose, because its raw-node pass must agree index for index with the mapping `safe_load` actually built. Switching loaders would change which suppression files load at all — a change to a user-facing format's accepted input, not a diagnostics cleanup.
- `tests/test_yaml_strict.py`, `abicheck/model/AGENTS.md` routing row, changelog fragment.

Remaining copies, deliberately out of scope and named in the commit message: `scripts/pipeline_status_ledger.py` (dependency-free by design, and needs two distinct error types) and the identical pair in `scripts/catalog_subjects.py` / `catalog_classification.py`.

## Behavioral deltas (user-visible)

| Input | Before | After |
|---|---|---|
| Sequence/mapping used as a mapping key, `--dump-manifest` / `--bundle-facts-library-manifest` | raw `TypeError` (translated by hand in one caller only) | that manifest's own error, `unhashable mapping key [...] (line N)` |
| `2023-99-99`-style implicit scalar, `--dump-manifest` | bare `ValueError` | `ManifestValidationError: ... invalid YAML scalar: month must be in 1..12` |
| Same input, `impact-use-cases.yaml` | `invalid scalar value: ...` / `invalid YAML syntax: ...` | one shared wording: `invalid YAML scalar: ...` / `invalid YAML: ...` |

Acceptance is unchanged everywhere: no document that loaded before is rejected now, and none that was rejected now loads.

## Testing

`tests/test_yaml_strict.py` states the primitive's contract as invariants over generated documents rather than fixed strings, per AGENTS.md's "Primitive-level property tests" rule:

- **Acceptance**: 200 Hypothesis-generated documents round-tripped through `yaml.safe_dump`, asserted equal to `yaml.safe_load` — an independent oracle, not a second call into the code under test.
- **Rejection**: duplicates constructed at every nesting depth 0–4 with a paired clean-document control; repetition count and value spelling varied; the whole small domain of unhashable key shapes enumerated; a sibling-mapping repeat asserted *accepted*, so the invariant cannot be satisfied by a loader that rejects too much.
- **Translation**: eight malformed classes each asserted to arrive as the caller's own type, plus a vacuity guard that `StrictMappingError` is not a `yaml.YAMLError` (otherwise the precise-message tests would pass for the wrong reason).
- Two mutations were run against the suite and both fail it: removing the hashability check (7 failures) and disabling the duplicate check (11 failures).

Also run: `ruff check abicheck/ tests/` clean, `mypy abicheck/` clean (799 files), `scripts/check_architecture.py` and `scripts/check_ai_readiness.py` show only the three pre-existing `debt-no-growth` entries and the pre-existing ADR-049 `Verified:` receipt error (all present on `main` — verified by stashing). Targeted suites (dump-manifest, packs, use-cases, bundle-facts overrides, suppression, project-validate, pack application, every `-k yaml` test) pass; the full fast unit lane was still running at push time and the result will be reported on this PR.

## Checklist

- [x] Changelog fragment added (`scriv create`) if this touches `abicheck/**/*.py`
- [x] Docs updated if user-visible behaviour changed (`abicheck/model/AGENTS.md`; error-message deltas are in the changelog fragment)
- [ ] `pre-commit run --all-files` passes locally — ran the equivalent gates individually (see Testing); the repo is not currently `ruff format`-clean on files this PR does not touch

https://claude.ai/code/session_01UJBQQfk7zXgj9X7Uk4o9B8

---
_Generated by [Claude Code](https://claude.ai/code/session_01UJBQQfk7zXgj9X7Uk4o9B8)_

### DIFF
--- a/abicheck/compatibility_evaluation_packs.py
+++ b/abicheck/compatibility_evaluation_packs.py
@@ -70,12 +70,11 @@ from pathlib import Path
 from types import MappingProxyType
 from typing import Any
 
-import yaml
-
 from .change_registry_types import Verdict
 from .compatibility_evaluation_config import ImmutableIdentity
 from .errors import PackManifestError
 from .model.change_catalog.kinds import ChangeKind
+from .model.yaml_strict import load_strict_yaml
 from .policy_file import parse_severity_value
 
 
@@ -97,54 +96,6 @@ _VALID_CHANGE_KIND_SLUGS: frozenset[str] = frozenset(k.value for k in ChangeKind
 #: silently discarding the pack author's actual intent (Codex review).
 _TOP_LEVEL_MANIFEST_FIELDS: frozenset[str] = frozenset(
     {"id", "version", "kind", "assignments"}
-)
-
-
-class _StrictLoader(yaml.SafeLoader):
-    """A ``SafeLoader`` that rejects a duplicate key in the same mapping.
-
-    ``yaml.safe_load`` alone silently accepts ``{func_removed: break,
-    func_removed: ignore}`` with last-value-wins semantics (PyYAML's
-    ``SafeConstructor.construct_mapping`` never checks for a repeat) -- for a
-    hard-load-error manifest format, that would silently drop the earlier,
-    contradictory assignment instead of raising (Codex review). Mirrors
-    ``dump_manifest.py``'s own ``_StrictLoader``/``_construct_mapping`` for
-    the identical ADR-050 D3 gap; kept as a second, independent copy rather
-    than a shared import since that one is private to its own module and this
-    manifest format has no other coupling to ``dump_manifest.py``.
-    """
-
-
-def _construct_strict_mapping(
-    loader: yaml.SafeLoader, node: yaml.MappingNode, deep: bool = False
-) -> dict[Any, Any]:
-    seen: set[Any] = set()
-    mapping: dict[Any, Any] = {}
-    for key_node, value_node in node.value:
-        key = loader.construct_object(key_node, deep=True)
-        # A complex YAML key (e.g. `? [a, b] : value`) constructs to an
-        # unhashable list -- `key in seen`/`seen.add(key)` would otherwise
-        # raise a raw TypeError instead of the documented PackManifestError
-        # (Codex review).
-        try:
-            hash(key)
-        except TypeError as exc:
-            raise PackManifestError(
-                f"unhashable mapping key {key!r} "
-                f"(line {key_node.start_mark.line + 1}): {exc}"
-            ) from exc
-        if key in seen:
-            raise PackManifestError(
-                f"duplicate key {key!r} in the same mapping "
-                f"(line {key_node.start_mark.line + 1})"
-            )
-        seen.add(key)
-        mapping[key] = loader.construct_object(value_node, deep=True)
-    return mapping
-
-
-_StrictLoader.add_constructor(
-    yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, _construct_strict_mapping
 )
 
 
@@ -610,26 +561,13 @@ def load_pack_manifest(path: str | Path) -> LoadedPack:
             f"{manifest_path}: pack manifest is not valid UTF-8 ({exc})"
         ) from exc
 
-    try:
-        # `_StrictLoader` *is* a `yaml.SafeLoader` subclass (it only adds the
-        # duplicate-key check above); bandit's B506 flags any `Loader=` it
-        # cannot name-match against `SafeLoader`/`CSafeLoader`, subclasses
-        # included, so this is the safe-loader path, not an arbitrary-object one.
-        document: Any = yaml.load(raw_text, Loader=_StrictLoader)  # nosec B506
-    except PackManifestError as exc:
-        raise PackManifestError(f"{manifest_path}: {exc}") from exc
-    except yaml.YAMLError as exc:
-        raise PackManifestError(f"{manifest_path}: invalid YAML ({exc})") from exc
-    except ValueError as exc:
-        # PyYAML's implicit timestamp resolver constructs a real
-        # datetime.date/datetime for a timestamp-shaped scalar -- an
-        # out-of-range value (e.g. "2020-99-99") raises a raw ValueError
-        # from that stdlib constructor, not yaml.YAMLError, and would
-        # otherwise escape this loader's documented PackManifestError
-        # contract (Codex review, fresh evidence).
-        raise PackManifestError(
-            f"{manifest_path}: invalid YAML scalar ({exc})"
-        ) from exc
+    # Strict loading (duplicate key, unhashable key, invalid implicit
+    # scalar, syntax error -- all as PackManifestError) is
+    # `abicheck.model.yaml_strict`'s job, shared with every other hard-load-error
+    # manifest format here rather than kept as a second private copy.
+    document: Any = load_strict_yaml(
+        raw_text, error=lambda message: PackManifestError(f"{manifest_path}: {message}")
+    )
 
     if not isinstance(document, dict):
         raise PackManifestError(
--- a/abicheck/dump_manifest.py
+++ b/abicheck/dump_manifest.py
@@ -121,59 +121,20 @@ _SUPPORTED_FRONTEND_CONTEXTS = frozenset({"host", "device"})
 
 
 def _load_yaml_strict(text: str, *, source: str) -> Any:
-    """Parse *text* as YAML, raising :class:`ManifestValidationError` on a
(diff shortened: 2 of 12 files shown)


napetrov:
### `unit-tests (macos-latest, 3.13, false)` is red on a pre-existing `main` failure, not this PR

**Failing check:** `unit-tests (macos-latest, 3.13, false)` — 26 failed, all in `tests/test_action_run_sh_helpers.py::TestIsReleaseStyleOperand` (the 11 `_is_release_style_operand` cases) plus `tests/unit/report/test_build_report_document.py::TestRendererOrderIndependence` (12 cases) and siblings.

**Why it isn't this PR's.** This diff touches only YAML manifest loading (`abicheck/model/yaml_strict.py` and four callers). It changes no shell script, no package-detection code and no report renderer. I reproduced both failing classes locally **on the base commit** `6cd4ce5b` (this PR's merge base), with this branch's changes not applied:

```
$ git checkout HEAD~1   # 6cd4ce5b, i.e. main
$ pytest -q tests/test_action_run_sh_helpers.py -k IsReleaseStyleOperand
11 failed, 7 passed
$ pytest -q tests/unit/report/test_build_report_document.py
12 failed, 75 passed
```

Identical counts and identical test ids on this branch — so the base is red here, and a re-run would reproduce it rather than clear it. (`main`'s own recent CI runs are all `cancelled` by superseding pushes, so there is no green base run to point at; the local base-commit reproduction above is the evidence.)

**Root cause of the `_is_release_style_operand` half, and a proposed patch.** PR #1253's Phase-7n change made that helper ask the *installed* abicheck first, and `abicheck.package.is_package()` is now content-based:

```
$ python -c "from abicheck.package import is_package; from pathlib import Path; \
             open('/tmp/empty.rpm','w').close(); print(is_package(Path('/tmp/empty.rpm')))"
False
```

The probe answers "definitely not a package" (exit 3), so `_is_release_style_operand` returns 1 without ever consulting its suffix table. The test fixtures write **zero-byte** files named `libfoo.rpm`, `libfoo.tar`, … — which under content-based detection genuinely are not packages. The new runtime behaviour looks correct; the tests are stale relative to it.

Proposed fix (a separate PR, since it is unrelated to this diff and would widen it):

- give `TestIsReleaseStyleOperand`'s package cases real content-shaped fixtures — RPM lead magic `ed ab ee db`, an ar `!<arch>\n` header for `.deb`, a genuine `tar`/zip archive for the tarball/`.conda`/`.whl` cases — so the probe answers `0` for the reason the test's name claims;
- keep one explicit case pinning the *new* behaviour: a zero-byte file with a package suffix is **not** a release-style operand;
- the `-k IsReleaseStyleOperand` set then exercises both the probe path and the suffix-table fallback (via `_PY_BIN_HAS_ABICHECK=false`), which is what the current suite silently stopped doing.

I have not applied that here: porting it would mean carrying an unrelated Action/packaging fix into a YAML-loader refactor, and no existing PR carries the fix for me to port. Happy to open it as its own PR if that's wanted.

**This PR's own state:** the full local fast lane (`pytest -m "not integration and not libabigail and not abicc and not slow and not golden"`, 46,173 passed) shows no failure outside those two pre-existing classes; `ruff`, `mypy abicheck/`, `scripts/check_architecture.py` and `scripts/check_ai_readiness.py` are clean apart from the same pre-existing entries present on `main`. I'll keep watching this PR and act on any failure that is genuinely its own.

---

napetrov:
### Correction and completion: the full pre-existing failure set is **six** classes, not two

My comment above said the local full lane showed "no failure outside those two pre-existing classes". That undercounted — the first run's output was captured tail-only. I re-ran the whole fast lane with `-rf` to get the complete list and checked **every** class against the base commit `6cd4ce5b`. Nothing changes about this PR (still no failure attributable to it), but the record should be accurate:

Local full fast lane on this branch (`b220675`): **27 failed, 46,173 passed, 44 skipped, 4 xfailed** in 40m24s. All 27 fall into six classes, and all six reproduce on the base commit with this branch's changes absent:

| Failing test class | Count | Reproduces on base `6cd4ce5b` |
|---|---:|---|
| `tests/unit/report/test_build_report_document.py::TestRendererOrderIndependence::test_envelope_projection_matches_render_output_byte_for_byte` | 12 | yes |
| `tests/test_action_run_sh_helpers.py::TestIsReleaseStyleOperand` (`test_package_extensions_are_release_style` ×10 + `test_package_extension_matched_case_insensitively`) | 11 | yes |
| `tests/test_service_unit.py::TestResolveInput::test_include_dependencies_defaults_true` | 1 | yes |
| `tests/test_python_api_reference.py::test_every_public_name_appears_in_generated_reference` | 1 | yes |
| `tests/test_python_api_reference.py::test_generated_reference_is_in_sync_with_service_all` | 1 | yes |
| `tests/test_ai_readiness.py::test_adr_status_sync_holds` | 1 | yes |

Verified by checking out the base commit and running each class there:

```
$ git checkout HEAD~1   # 6cd4ce5b
$ pytest -q tests/unit/report/test_build_report_document.py            → 12 failed, 75 passed
$ pytest -q tests/test_action_run_sh_helpers.py -k IsReleaseStyleOperand → 11 failed,  7 passed
$ pytest -q tests/test_python_api_reference.py \
          tests/test_ai_readiness.py::test_adr_status_sync_holds        →  3 failed,  4 passed
$ pytest -q tests/test_service_unit.py::TestResolveInput::test_include_dependencies_defaults_true → failed
```

The last three classes look like fallout from the same recent merges as the first two — #1258's typed-API/CLI defaults alignment (`include_dependencies`, the generated Python-API reference) and the ADR-049 `**Verified:**` receipt pointing at a commit not reachable from `main`, which `scripts/check_ai_readiness.py` reports as an error on `main` today. None of them touch YAML manifest loading, and none is introduced or made worse here.

CI on this PR reflects exactly that: only the two `unit-tests` lanes (macOS 3.13, Windows 3.13) are red, both entirely from these classes. Every other completed check is green, including `ai-readiness`, `lint-and-types`, `test-contract`, `changelog-fragment`, `cli-interface-diff`, `e2e`, `CodeQL`, `Security Scan`, the six example-validation shards and the PE/Mach-O native lanes. The proposed fix for the `_is_release_style_operand` class is unchanged from my previous comment, and still belongs in its own PR.

---

PR #1264, opened 2026-09-12
fix(report): align RenderOptions.show_recommendation with render_output

napetrov:
## Summary

Follow-up to [#1258](https://github.com/abicheck/abicheck/pull/1258), which merged before these fixes were pushed. Three things that change landed without: `RenderOptions.show_recommendation` left at `False` while `render_output`'s own default moved to `True`, a `docs/reference/python-api-reference.md` not regenerated for the new `to_stat`/`to_stat_json` exports, and a dependency-scope differential test that the defaults flip quietly made vacuous.

## Motivation

#1258 flipped `render_output`'s `show_recommendation` default to `True` so the function's documented behaviour ("the release recommendation is included in every human-facing format") is true for its own audience — a direct Tier-2 caller. But `render_output` builds a `ReportEnvelope`, and `RenderOptions.show_recommendation` stayed `False`. So the two documented entry points into the same projection disagreed:

- `render_output(fmt, result, old, new)` → Markdown **with** the Release Recommendation section.
- `render_envelope(fmt, build_report_envelope(...))`, option omitted → Markdown **without** it.

That is the same class of defect #1258 was about — two surfaces disagreeing about what *omitting* an option means — reintroduced one layer down. An existing test caught it: `tests/unit/report/test_build_report_document.py::TestRendererOrderIndependence::test_envelope_projection_matches_render_output_byte_for_byte`, which exists precisely to stop these two drifting. 20 parametrizations failed.

## Changes

- `abicheck/report/envelope.py` — `RenderOptions.show_recommendation` now defaults to `True`.
- `abicheck/service_render.py` — `_project_markdown`'s docstring stated the old `False` default as deliberate; it now records the alignment and why the mismatch mattered.
- `docs/reference/python-api-reference.md` — regenerated (`python scripts/gen_python_api_reference.py`). `tests/test_python_api_reference.py` fails on current `main` without this: `to_stat`/`to_stat_json` are in `abicheck.service.__all__` but absent from the reference, which also still recorded `include_dependencies` as defaulting to `True`.
- `tests/test_service_unit.py` — `resolve_input`'s `include_dependencies` default test renamed and flipped to `False`, matching the surface alignment #1258 landed, pointing at `tests/test_front_end_default_parity.py` for the cross-surface invariant.
- `tests/test_dump_write_after_resolve_time_embed.py` — **a second, subtler consequence of the same flip.** `test_write_snapshot_output_accepts_a_resolve_time_embedded_snapshot` asserted `snap.dependency_scope == "full"` before calling `_write_snapshot_output`, and `payload["dependency_scope"] == "filtered"` after; the contrast *is* the test — it is how the test knows `resolve_dependency_scope` ran. Post-flip `execute_dump_request` already returns `filtered`, so both ends read `filtered` and the assertion held no matter what that step did. Changing the expected string would have left a green test proving nothing — the failure mode `AGENTS.md` describes under "a differential test must prove both of its configurations actually ran". The request now asks for the unfiltered surface explicitly, keeping the two states distinct, with the non-default value documented as load-bearing. **Confirmed non-vacuous by deleting the `resolve_dependency_scope` call:** the restored test fails, and passes again when restored; before the fix that same deletion passed.

## Bug-fix test contract

- Bug class: `config.front_end_default_divergence` (`tests/regressions/manifest.py`) — two front ends onto one operation disagreeing about what *omitting* an option means, so nothing is stated anywhere and every propagation test passes.
- Publicly observable failure: `render_envelope("markdown", build_report_envelope(...))` omits the `## Release Recommendation` section that `render_output("markdown", ...)` on the same comparison emits.
- Regression test fails on base: yes — `tests/unit/report/test_build_report_document.py::TestRendererOrderIndependence::test_envelope_projection_matches_render_output_byte_for_byte` fails in 20 parametrizations on the base commit and passes here. No new test was needed; the existing invariant already covered the class.
- Negative control: `show_recommendation=False` still suppresses the section through both entry points (asserted by the same file's existing per-option tests, which pass unchanged).
- Public-surface test: yes — the assertion compares the two public entry points' rendered output byte for byte, not an internal struct.
- Axes covered: all four `_CHANGE_COMBINATIONS` change sets × three option sets (`plain`, `show_only`, `impact`) × every format in `_FORMATS`.
- General invariant: `render_output` is exactly "build one envelope, project it" — asserted as byte equality across that whole product rather than for one option or format, so any future option that one entry point threads and the other ignores fails here too. The oracle is the *other* entry point's real output, not a re-derivation of either's formatting.

## Verification

| check | result |
|---|---|
| `tests/unit/report/`, `test_python_api_reference.py`, `test_service_unit.py`, `test_front_end_default_parity.py` | 515 passed |
| integration lane (`-m integration`, real castxml) | the one failure caused by this change fixed; see below |
| `mypy abicheck/` | clean, 798 files |
| `ruff check abicheck/ tests/` | clean |
| `python scripts/check_architecture.py` | 0 errors, no baseline raised |
| `python scripts/check_ai_readiness.py` | 0 errors |
| `scripts/check_bugfix_test_contract.py` (both halves, real body) | OK |

### CI state: every remaining failure is pre-existing

| lane | result |
|---|---|
| `unit-tests (macos-latest, 3.13)` | 11 failed, **45,329 passed** |
| `unit-tests (windows-latest, 3.13)` | 13 failed, **44,774 passed** |
| everything else in the run | success or skipped |

All 24 belong to the `action/run.sh` release-operand breakage introduced by `65473b26` (#1260), which made `_is_release_style_operand` consult the installed `abicheck.package.is_package()` first and treat a definite "no" as final — while the fixtures are zero-byte files that content-based detection correctly reports as non-packages. Diagnosis, a verified patch sketch, and the open semantics question are in the comments; I have not fixed it here, because it is another PR's CLI-surface migration and the "empty suffix-named file vs. content decides" call is its author's.

Two scope notes worth carrying: the Windows leg fails on **strictly more** than the Linux leg (13 vs 11 — two extra in `tests/test_action_release_operand_detection.py`, which pass on Linux), so a Linux-only fix will leave `windows-latest` red. And a `Baseline regression (PR vs base)` failure on an earlier sha was investigated and is a **measurement artifact**: head and base are timed sequentially on one runner, and local measurement reproduces CI's baselines to within 1–2% while showing head equal to base — details in the comments.

## Checklist

- [x] Tests added / updated for new behaviour — the existing invariant already covered the report class; two tests updated, one of them to restore a differential the flip had made vacuous
- [x] Docs updated if user-visible behaviour changed — generated Python API reference
- [ ] `pre-commit run --all-files` passes locally — individual gates run directly, see table above

https://claude.ai/code/session_018BYuG83u8jFkGVn5JXwvMD

### DIFF
--- a/abicheck/report/envelope.py
+++ b/abicheck/report/envelope.py
@@ -105,7 +105,7 @@ class RenderOptions:
     show_impact: bool = False
     demangle: bool = False
     follow_deps: bool = False
-    show_recommendation: bool = False
+    show_recommendation: bool = True
     require_complete_analysis: bool = False
     contract_evaluation: bool = False
 
--- a/abicheck/service_render.py
+++ b/abicheck/service_render.py
@@ -306,10 +306,14 @@ def _project_review(envelope: ReportEnvelope) -> str:
 def _project_markdown(envelope: ReportEnvelope) -> str:
     """The full Markdown report (and its ``md`` alias).
 
-    ``show_recommendation`` stays this projection's own presentation option
-    with its pre-removal Tier-2 default of ``False`` (CLI cleanup phase two,
-    PR 1) -- ``cli._render_output`` passes ``True`` explicitly, which is why
-    the CLI's output is unconditional without this default changing.
+    ``show_recommendation`` stays this projection's own presentation option,
+    read off the envelope. It defaults to ``True`` in both
+    :class:`~abicheck.report.envelope.RenderOptions` and ``render_output``
+    above, so building an envelope directly and calling ``render_output``
+    with the option omitted produce byte-identical output -- the two used to
+    disagree (``False`` here, ``True`` there), which meant a direct Tier-2
+    envelope caller silently lost the Release Recommendation section that
+    every real consumer gets.
 
     Gap-C disposition for Markdown's own remaining facts: ``severity_groups``'
     headed-section grouping is **presentation** over already-classified
--- a/changelog.d/20260912_225403_noreply_hopeful_volta_h46a24.md
+++ b/changelog.d/20260912_225403_noreply_hopeful_volta_h46a24.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- `RenderOptions.show_recommendation` now defaults to `True`, matching
+  `render_output`'s own default. The two disagreed, so building a
+  `ReportEnvelope` directly and projecting it silently dropped the Release
+  Recommendation section that every consumer going through `render_output`
+  gets.
--- a/docs/reference/python-api-reference.md
+++ b/docs/reference/python-api-reference.md
@@ -136,7 +136,7 @@ One side of a comparison: a binary/snapshot path plus its build context.
 | `version` | `str` | `''` |
 | `pdb` | `Path \| None` | `None` |
 | `debug_roots` | `tuple[Path, ...]` | `()` |
-| `include_dependencies` | `bool` | `True` |
+| `include_dependencies` | `bool` | `False` |
 | `sources` | `Path \| None` | `None` |
 | `build_info` | `Path \| None` | `None` |
 | `build_targets` | `tuple[str, ...]` | `()` |
@@ -360,8 +360,7 @@ Render comparison result in the requested output format.
 | `severity_config` | `SeverityConfig \| None` | `None` |
 | `demangle` | `bool` | `False` |
 | `contract_evaluation` | `bool` | `False` |
-| `stat` | `bool` | `False` |
-| `show_recommendation` | `bool` | `False` |
+| `show_recommendation` | `bool` | `True` |
 | `require_complete_analysis` | `bool` | `False` |
 
 **Returns:** `str`
@@ -409,7 +408,7 @@ Auto-detect input type and return an ABI snapshot.
 | `notify` | `Callable[[str], None] \| None` | `None` |
 | `include_labels` | `dict[Path, str] \| None` | `None` |
 | `dump_manifest` | `DumpManifest \| None` | `None` |
-| `include_dependencies` | `bool` | `True` |
+| `include_dependencies` | `bool` | `False` |
 | `public_include_search_dirs` | `list[Path] \| None` | `None` |
 
 **Returns:** `AbiSnapshot`
@@ -457,7 +456,7 @@ Compare two ABI inputs and return the classified diff result.
 | `debuginfod_url` | `str \| None` | `None` |
 | `diagnostic_comparison` | `bool` | `False` |
 | `contract_evaluation` | `bool` | `False` |
-| `include_dependencies` | `bool` | `True` |
+| `include_dependencies` | `bool` | `False` |
 | `contract_mode` | `str \| None` | `None` |
 | `pack_policy_overrides` | `dict[Any, Any] \| None` | `None` |
 | `pack_internal_namespaces` | `tuple[str, ...] \| None` | `None` |
@@ -512,7 +511,7 @@ Extract an ABI snapshot from a native binary (ELF, PE, or Mach-O).
 | `include_labels` | `dict[Path, str] \| None` | `None` |
 | `dump_manifest` | `DumpManifest \| None` | `None` |
 | `public_include_search_dirs` | `list[Path] \| None` | `None` |
-| `include_dependencies` | `bool` | `True` |
+| `include_dependencies` | `bool` | `False` |
 
 **Returns:** `AbiSnapshot`
 
@@ -535,5 +534,33 @@ Read a small header chunk and return ``'json'``, ``'perl'``, ``'symvers'``, or `
 | Parameter | Type | Default |
 |---|---|---|
 | `path` | `Path` | *(required)* |
+
+**Returns:** `str`
+
+## `to_stat`
+
+One-line summary for CI gates.
+
+| Parameter | Type | Default |
+|---|---|---|
+| `result` | `DiffResult` | *(required)* |
+| *(keyword-only below)* | | |
+| `severity_config` | `SeverityConfig \| None` | `None` |
+
+**Returns:** `str`
+
+## `to_stat_json`
+
+JSON output for --stat mode: summary only, no changes array.
+
+| Parameter | Type | Default |
+|---|---|---|
+| `result` | `DiffResult` | *(required)* |
+| `indent` | `int` | `2` |
+| *(keyword-only below)* | | |
+| `severity_config` | `SeverityConfig \| None` | `None` |
+| `require_complete_analysis` | `bool` | `False` |
+| `show_only` | `str \| None` | `None` |
+| `contract_evaluation` | `bool` | `False` |
 
 **Returns:** `str`
--- a/tests/test_dump_write_after_resolve_time_embed.py
+++ b/tests/test_dump_write_after_resolve_time_embed.py
@@ -176,6 +176,17 @@ def test_write_snapshot_output_accepts_a_resolve_time_embedded_snapshot(
             sources=tmp_path,
             build_info=compile_db,
             compile=CompileContext(frontend="clang"),
+            # Deliberately the NON-default value. Every front end now
+            # defaults `include_dependencies` to False (PR #1258), so a
+            # request that omits it reaches `_write_snapshot_output` already
+            # `filtered` -- and then the "did `resolve_dependency_scope`
+            # run?" contrast below compares `filtered` with `filtered` and
+            # holds no matter what that step does, which is exactly the
+            # vacuous differential AGENTS.md's "a differential test must
+            # prove both of its configurations actually ran" warns about.
+            # Asking for the unfiltered surface here keeps the two states
+            # genuinely distinct, so the flip is still real evidence.
+            include_dependencies=True,
         ),
         depth="source",
     )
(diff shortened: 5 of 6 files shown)


napetrov:
Heads-up before the unit lanes report here: `unit-tests` will fail on this PR with 11 failures in `tests/test_action_run_sh_helpers.py::TestIsReleaseStyleOperand`, and **they are not this PR's**. This diff touches `abicheck/report/envelope.py`, `abicheck/service_render.py`, a generated doc and one test — nothing in `action/`.

**Established, not assumed.** They reproduce identically on clean `origin/main` (`git checkout origin/main && pytest tests/test_action_run_sh_helpers.py -q` → `11 failed, 128 passed`), so this is a red base branch rather than a flake — no re-run spent, because the local reproduction on the unmodified base is the stronger evidence a re-run would be standing in for.

**Root cause.** `65473b26` ("derive the release-operand test from the installed abicheck", #1260) made `_is_release_style_operand` ask the installed `abicheck.package.is_package()` *first*, and treat probe exit 3 as a definite "not a package" that returns before the suffix `case` table is reached. The test's fixtures are zero-byte files (`f.write_text("", encoding="utf-8")`), and content-based detection correctly reports an empty `libfoo.rpm` as not a package:

```
empty .rpm is_package -> False
empty .tar is_package -> False
```

So the predicate answers `1` for every zero-byte fixture, and the suffix table that used to answer `0` is now unreachable. The production behaviour change is intentional; the fixtures are what went stale.

**No fix exists to port**, so here is the proposed patch rather than a widening of this PR. The fix belongs in the test, not in `run.sh` — making the predicate fall back to the suffix table on a definite `False` would undo exactly what #1260 set out to fix (a tar/conda/wheel staged under a nonconventional name). Give each fixture real content so the probe answers what the test means to assert. Verified locally:

```
rpm w/ 0xedabeedb lead magic -> True
real tar (tarfile.open(..., "w")) -> True
real whl (zipfile) -> False   # still needs wheel-shaped members
```

`.whl` needs more than a bare zip, so the parametrization likely wants a per-suffix content builder rather than one shared empty write. Worth confirming with #1260's author which side they intended: the split between "a suffix-named empty file is a release operand" and "content decides" is a real behavioural question, and the answer determines whether the two `test_package_extension_matched_case_insensitively` cases want real content too.

I'm not pushing this from here — it's outside this PR's scope and belongs with the change that introduced it.

---

napetrov:
## `Baseline regression (PR vs base)` — failed on a stale sha, and the failure is a measurement artifact

The failing run measured `44baab4b`, which is **#1258's old head**, not this PR's `0a578fa7`. It reports against this branch because the branch name is shared. Recording the investigation anyway, because the same check will run here and this is a methodology finding rather than a one-off.

It reported five failures against a 15% tolerance:

| scenario | CI head | CI baseline | CI delta |
|---|---|---|---|
| `union_churn` @ 2000 | 1.293s | 1.077s | +20% |
| `internal_leak` @ 1000 | 1.071s | 0.882s | +21% |
| `internal_leak` @ 2000 | 2.326s | 1.865s | +25% |
| `vtable_churn` @ 1000 | 0.595s | 0.484s | +23% |
| `vtable_churn` @ 2000 | 1.230s | 1.011s | +22% |

**No mechanism exists.** All five scenarios build `AbiSnapshot` pairs in memory and time `abicheck.checker.compare` directly — they never call `run_dump`, `InputSpec`, `resolve_input`, or any renderer. `44baab4b`'s diff against its base touches 24 files, none of them `checker.py` or any `diff_*` module, and the two report-layer files it does touch (`reporter.py`, `service_render.py`) are not on this path at all. The head↔base diff is also entirely #1258's own: the merge commit in it merged `origin/main` at the moment main *was* the base, so it contributed nothing.

**Measured rather than argued.** Checking out each sha and running the harness locally:

| scenario | base `ae27d9f6` | head `44baab4b` | CI's baseline |
|---|---|---|---|
| `internal_leak` @ 1000 | 0.885s (cv 2.1%) | 0.879s (cv 3.6%) | 0.882s |
| `union_churn` @ 2000 | 1.067s (cv 3.2%) | 1.079s (cv 2.1%) | 1.077s |
| `vtable_churn` @ 1000 | 0.473s (cv 2.2%) | 0.463s (cv 3.3%) | 0.484s |

Head equals base locally in all three, within noise. The decisive detail is the third column: **my base numbers reproduce CI's baselines to within 1–2%**, so the harness is stable and the runner is comparable. It is only CI's *head* half that is ~20% high, uniformly across every failing scenario — including one where the local head number came out marginally *faster* than base.

**So the artifact is in the job, not the code.** The job checks out both revisions (`/head` and `/base`) and measures them in sequence on one runner, which makes the second half pay for whatever the first half did to the machine — cache and thermal state on a shared runner. A uniform one-directional skew across five independent scenarios, with per-scenario `cv` of 0.4–6% *within* each half, is the signature of that rather than of five separate regressions. Note the same log shows `cv` up to 30% for `report_sarif`, so the harness is not uniformly tight at these sizes either.

I'm not pushing anything for this: there is nothing in this PR to fix, and changing the gate's measurement order (interleave the halves, or measure each half in its own job) is a change to the performance workflow that belongs in its own PR with its own discussion — not folded into a report-defaults fix. Flagging it as the more useful outcome, since an order-biased gate will keep producing ~20% phantom regressions on unrelated PRs.

No re-run spent — the check runs again on `0a578fa7` on its own, which is the confirmation.

---

napetrov:
## `integration-tests (ubuntu-24.04)` — one of the eight was mine; fixed in 10e9254f

Reproduced the whole lane locally (castxml available): **8 failed, 554 passed**. All eight also fail on clean `origin/main`, but that is *not* evidence they are someone else's — #1258 is already merged into main, so "pre-existing on main" and "mine" are not mutually exclusive here. Splitting them by actual cause rather than by which branch shows them:

### Mine (1) — fixed

`tests/test_dump_write_after_resolve_time_embed.py::test_write_snapshot_output_accepts_a_resolve_time_embedded_snapshot`

```
assert snap.dependency_scope == "full"
E   AssertionError: assert 'filtered' == 'full'
```

Direct consequence of #1258's defaults flip. The important part is that **changing the expected string would have been the wrong fix.** That assertion is the test's *ground truth that `resolve_dependency_scope` had not yet run*; a later `assert payload["dependency_scope"] == "filtered"` is what proves it did. With the flip, `execute_dump_request` already returns `filtered`, so both ends of the contrast read `filtered` and the test would have passed green while proving nothing about the step it exists to guard — the precise failure mode `AGENTS.md` describes under "a differential test must prove both of its configurations actually ran". The test's own comment even says so: *"nothing before this call flips `dependency_scope` off `full`"*. That premise stopped being true.

So the request now asks for the unfiltered surface explicitly (`include_dependencies=True` on the `InputSpec`), keeping the two states genuinely distinct, with a comment recording why the non-default value is load-bearing rather than incidental.

**Confirmed non-vacuous, not merely green:** deleting the `resolve_dependency_scope` call from `_write_snapshot_output` makes the restored test fail, and it passes again when restored. Before the fix that same deletion passed.

### Not mine (7) — the `-o FORMAT=DESTINATION` rewrite

`tests/test_cli_compare_bundle_facts.py` (2) and `tests/test_release_assurance_cli.py` (5). Two distinct symptoms, neither reachable from this PR's diff:

- Error-wording drift — tests assert `'collide' in output` / `'same path' in output`, while the CLI now says *"the second would silently overwrite the first. Give each export its own destination."*
- `assert 64 == 1` and `assert 64 == 4` — exit **64** is `_EXIT_USAGE_ERROR`, i.e. these invocations are now rejected at parse time rather than running, so the assertions compare a usage error against an expected verdict exit.

Both are the signature of the export-surface unification (`-o FORMAT=DESTINATION` replacing `--format`/`--output-dir`): the tests still pass the retired spelling and assert the old message text. They belong with that change, and I have not touched them — fixing another PR's CLI-surface migration from inside a report-defaults follow-up is the widening these rules warn against.

Worth flagging plainly: **`main`'s integration lane is red on those 7 independently of anything here**, alongside the 11 unit-lane failures in `tests/test_action_run_sh_helpers.py` I reported earlier.

---

napetrov:
### Scope correction on the release-operand breakage: it is wider than I reported, and Windows-only in part

My earlier comment named `tests/test_action_run_sh_helpers.py::TestIsReleaseStyleOperand` (11 cases). The CI results on this PR's head `10e9254f` show that was incomplete — **not wrong, but under-counted**, because I could only reproduce on Linux:

| lane | result |
|---|---|
| `unit-tests (macos-latest, 3.13)` | **11 failed**, 45,329 passed — exactly the set I named |
| `unit-tests (windows-latest, 3.13)` | **13 failed**, 44,774 passed — those 11 **plus 2 more** |

The two extra are in a *different* file, `tests/test_action_release_operand_detection.py`:

- `TestAgreementWithTheRealPredicate::test_every_package_shape_agrees_under_a_nonconventional_name`
- `TestThePreInstallFallback::test_the_fallback_is_what_the_probe_improves_on` — `assert False`, where `False = _ask(WindowsPath('.../operand'), abicheck_available=True, ...)`

Same predicate, same root cause family as the other 11 (`65473b26`'s content-based `is_package()` probe versus fixtures that predate it), and equally unrelated to this PR: that file arrived with #1260, and greps clean for every symbol this PR touches (`show_recommendation`, `RenderOptions`, `render_output`, `include_dependencies`, `to_stat` — zero hits).

**They pass on Linux** (`5 passed` locally on this branch), which is why my local full-lane and integration runs could not see them, and why my first report undercounted. Anyone fixing the release-operand fixtures should know the Windows leg fails on strictly more than the Linux leg — fixing only what reproduces on Linux will leave `windows-latest` red.

No second comment on the shared root cause; the proposed patch and the open semantics question are in my earlier comment and unchanged. Recording this only because the scope I gave there was narrower than the truth.

Worth noting from the same two runs: **45,329 and 44,774 tests pass** on macOS and Windows respectively, covering this PR's entire diff on both platforms. Every failure on this head belongs to the pre-existing breakage.

---

PR #1267, opened 2026-09-13
fix: close five review findings on the L2 perf harness (include-pass scaling, service classification, export parity, set cache lifecycle, base-baseline gating)

napetrov:
## Summary

Follow-up to #1262 (merged): five review findings against that work, each of which let a *faster, wrong* run pass — or let a correct run fail.

## Motivation

#1262 merged while two Codex findings on its head were still open, and three more arrived against the merged head. Each is a case of the defect class that PR kept hitting: a check that accepts one favourable observation, or an expectation narrower than the thing it claims to verify. They are defects in `main` now, so they are fixed here rather than left.

## Changes

- **Include-pass floor now scales with the header count.** It required one include/dependency pass per live *side* (2), while the extended `simple-h8`/`simple-h32` arms really perform **16** and **64** (measured: h1→2, h8→16, h32→64). A regression processing only the first top-level header of each side dropped to 2, stayed under the floor, kept `effective_depth=headers`, and still found the break in `part0.h` — so every other check agreed while most of the include-graph work had stopped. The floor is derived from the measured argv's own `--header`/`-H` arguments, which cannot drift from what the run was told to do and makes the `one_side` case fall out for free (a stored operand names no header).
- **The perf path filter classifies the whole `abicheck/service*.py` family.** It named `service.py`, which only re-exports `_attach_header_graph` from `service_header_graph_attach.py` for import stability — so a PR changing the real attach implementation, code both the in-process benchmark and every full-CLI live dump execute, skipped every performance job. Swept by glob per the `buildsource/**` precedent; the four now-redundant per-file entries are gone.
- **Both exports of the two-format scenario are validated against the same findings.** The JSON half was checked only for "did this reach L2" and the rendered half only for the word `BREAKING`, so dropping the removal/layout findings while keeping the verdict metadata read as a pure speedup. Both now check the shared `EXPECTED_BREAK_KIND_FAMILIES` table (the renderer names each kind slug verbatim), deliberately shared rather than restated so the two expectations cannot disagree.
- **The multi-library set is one cache lifecycle.** Every member used the default per-step `cold` mode, so each library's comparison started from an empty cache and cross-library reuse — the only thing separating the shared-context arm from the distinct-context one — was unobservable by construction. Now one `shared_workload` sequence, reset once before the set and still beginning cold. **This makes the measurement say something it previously could not:** at `--repeat 2` every member of both arms performs 4 header extractions and 2 include passes, identical in both arms, so **no cross-library reuse happens today**. Recorded in `docs/contribute/performance.md` as a product observation — this change alters no caching strategy.
- **Whether to gate is asked of the gate's own rule.** The workflow decided by testing the base receipt for non-emptiness; a base run that failed every scenario still writes a full diagnostic receipt, so that passed and the head run then failed for zero overlap, turning an unmeasurable base into a failed PR. It now asks `--count-gateable-points`. Requiring `base_status == 0` would be wrong in the other direction — it would discard six good baselines because a seventh scenario failed, which `gated_points` already drops on its own.

Two structural consequences, both splits rather than trims: `scripts/l2_cli_observation.py` (the harness crossed the 2000-line cap a fifth time; every function there takes a `CommandRun` plus a `Step` and reaches no fixture, argv, filesystem or scenario) and `tests/test_l2_cli_perf_cache.py` / `tests/test_l2_cli_perf_lane.py` (the contracts test file crossed the 1200-line test cap; thematic siblings, not a debt entry).

## Bug-fix test contract

- Bug class: a correctness check whose expectation is narrower than the work it claims to verify, so a run that performed less of that work passes it (the include-pass floor, the single-export finding check) — plus its mirror, a usability check whose proxy is weaker than the question it stands for (file size for "is this a usable baseline").
- Publicly observable failure: a PR that stops running the include-graph pass over all but the first header per side, or that drops every finding from one export, is reported as a faster, passing perf run; and a base measurement that failed every scenario fails the PR rather than falling back to report-only.
- Regression test fails on base: `TestIncludePassesScaleWithTheHeaderCount::test_the_per_side_floor_alone_no_longer_passes_a_multi_header_run` (the old floor accepted `include_pass=2` for 8 and 32 headers); `TestBothExportsAreValidatedAgainstTheSameFindings::test_the_verdict_alone_does_not_satisfy_it`; `TestTheMultiLibrarySetSharesOneCacheLifecycle::test_the_cache_is_reset_once_for_the_set_not_per_member`; `TestTheWorkflowAsksTheHarnessWhetherToGate::test_the_l2_job_no_longer_decides_on_file_size`.
- Negative control: `test_one_pass_per_named_header_is_accepted` (a correct count must still pass, so the floor is not simply always-failing); `test_an_unobserved_counter_is_still_unknown_not_a_failure`; `test_a_render_naming_every_family_passes`; `test_the_glob_is_not_a_blanket_over_abicheck` (a pattern matching all of `abicheck/` would pass the coverage test); `test_the_body_is_really_just_this_job` (a workflow reader returning the whole file would pass the assertions above vacuously).
- Public-surface test: `--suite pr --repeat 1 --require-toolchain` and `--suite extended --scenario h8 --scenario h32 --scenario multi_library` run the real CLI as a subprocess against real compiled fixtures; `main(["--count-gateable-points", ...])` is exercised through the harness's real entry point.
- Axes covered: header count 1/2/8/32 × 1/2 live sides (including the one-short boundary for each), both `--header` and `-H` spellings, an argv naming no header at all, an unobserved counter; every alternative in every break family by cartesian product; library counts 2 and 5 in shared and distinct arms; a base receipt that is wholly failed, partially failed, malformed, and absent.
- General invariant: *a check must fail whenever less of the promised work was done, and an absent observation is unknown rather than satisfied.* Exercised with generated inputs beyond the reported case (the header-count sweep and the `itertools.product` over break-family alternatives) against oracles derived independently of the implementation — the expected include-pass count comes from counting argv flags, not from the implementation's own helper, and the break-family expectation comes from the product's own shared table.

## Checklist

- [x] `pre-commit run --all-files` passes locally

### Local verification

592 tests passed across the nine perf suites; `--suite pr --repeat 1 --require-toolchain` green end to end; `--suite extended` green on the h8/h32 and multi-library arms at `--repeat 2`; `check_ai_readiness.py` and `check_architecture.py` both 0 errors; `ruff check scripts/ tests/` clean.

Three environment caveats, stated so they are not read as results: three tests fail in this container only because another process deleted the global ssh commit-signing helper `/tmp/code-sign` (all three pass with `commit.gpgsign=false`); `ruff format --check` reports three `scripts/*.py` files this branch never touches (reproduced on a clean stash — a ruff-version difference here); and `unit-tests (macos-latest)` fails identically on clean `main` via `TestIsReleaseStyleOperand` (commit `65473b26`).

https://claude.ai/code/session_01Vee2ztjM4Utc7E3fVNjsEN

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vee2ztjM4Utc7E3fVNjsEN)_


napetrov:
**CodeFactor on `1ae956bb`: "1 issue fixed. 4 issues found."** — not fixing these, and here is the reasoning rather than a dismissal. The check run publishes no detail and no inline comments, so I reproduced its analyzers locally over every file this branch touches.

What the analyzers report, and why each stands:

- **`C0413` wrong-import-position** (`scripts/l2_cli_observation.py`, and the pre-existing block in `check_l2_cli_perf.py`). These modules must insert `_SCRIPTS_DIR` on `sys.path` *before* importing their siblings, so that each resolves whether it is loaded directly as a script or as a `scripts.` submodule by a test that never imported the parent. That guard is the documented pattern across this whole family (`l2_cli_argv.py`, `l2_cli_model.py`, `l2_cli_validation.py`, `l2_cli_gating.py`), and the imports carry `# noqa: E402` for exactly this reason. Moving them above the guard breaks the import.
- **`W0212` protected-access** (the cache test file). These tests exist *specifically* to pin the contracts of private helpers — `_warm_cache_problems`, `_needs_cold_cache`, `_classify_cache_service`, `_SEQUENCE_CACHE_MODES`. That is the repo's own "primitive-level property tests" guidance in `AGENTS.md`: test the reusable primitive directly, not only through its highest-level caller. Routing them through a public wrapper purely to satisfy this check would make them test less.
- **`C0302` too-many-lines** (`check_l2_cli_perf.py`, 1867/1000). This repo's own cap is 2000 (`scripts/check_ai_readiness.py`'s `file-size` gate), and that gate is green. The file is under its real cap *because* this PR split `l2_cli_observation.py` out of it; shrinking it toward pylint's default 1000 would mean more splits chosen by a line count rather than by responsibility, which the repo explicitly warns against ("the way to shrink an entry is to move responsibility out to a properly-owned module").

Complexity specifically: `radon cc` reports **no D or E grade** anywhere in the diff. The highest are C(18) `_validate_snapshot`, C(15) `_print_table`, C(14) `classify_perf_paths.main` — all pre-existing shapes, and `pylint --enable=R0912,R0913,R0914,R0915,R0902,R0904,R1702` is clean on every file here.

This is the same position I reached on #1262 after reproducing its findings the same way; I am restating it here because this is a separate PR and a stand-down on one does not carry to another. If a maintainer would rather I satisfy any of these three classes, say which and I will — the `W0212` one is the only one I could address without losing something, and it would cost real coverage.

The rest of CI is still queued on this head; I am watching the PR and will act on whatever it reports.

---

napetrov:
P1 fixed in `8c0ce7ee` — and the finding is right in the way that stings: it cites the rule from `AGENTS.md` that I wrote there myself after an earlier `performance.yml` assertion hit exactly this. My new reader returned the job's raw text, so commenting out the `--count-gateable-points` invocation or the positive-point conditional left the asserted substring present as a comment and both wiring tests passed while the workflow no longer gated.

`_strip_comments` follows `test_classify_perf_paths.py`'s existing rule (cut at the first `#`, drop blank remainders) rather than adding a second, subtly-different stripper — that file already had the helper for the same reason.

It is proven by mutation rather than asserted in prose, since prose is what failed here: `_workflow_l2_body` now takes optional text, so `TestTheWorkflowReaderIgnoresComments` pushes a deliberately-broken copy of the **real** workflow through the **real** reader, comments out each depended-on line, and requires the assertion to fail. Two guards around it: one asserting the mutation really leaves the text present (otherwise the mutation test would pass because the text was removed, not because stripping caught it), and one requiring the real workflow to still satisfy both assertions, so stripping cannot be over-aggressive in the other direction.

Verification on this head: 597 tests passed across the nine perf suites, readiness and architecture 0 errors, `ruff check scripts/ tests/` clean.

On the `unit-tests (macos-latest)` failure that arrived for this PR: it is on the old sha `2ca79b5e`, and it is the `TestIsReleaseStyleOperand` failure that reproduces identically on clean `main` (`action/run.sh` delegating to `is_package`, commit `65473b26`) — already established, not this PR's, and noted in the description. Nothing ported because no fix for it exists yet.

---

napetrov:
**Not this PR's, but it needs saying loudly: `main` is red, and I previously mis-diagnosed the cause.**

The CI events arriving for this PR are from run `34729579282` on sha `2ca79b5e` — #1262's own head, now merged into `main`. Five jobs failed there, and two of them are not the known platform ones:

- `unit-tests (ubuntu-latest, 3.12)`: **2 failed, 20,867 errors**
- `unit-tests (ubuntu-latest, 3.14)`: **2 failed, 29,164 errors**

Every error is the same: `FileNotFoundError: .../pytest-of-runner/pytest-0/popen-gwN/snapshot-caches-XXXXXXXX` out of the autouse `_isolate_snapshot_cache` fixture.

**I need to correct myself here.** I hit this exact signature locally earlier today (287 setup errors), attributed it to another process in my container deleting `/tmp/pytest-of-root`, worked around it with `--basetemp`, and said so in a comment on #1262 and in this PR's description. The local deletion was real — but calling it *only* external interference was wrong, because there is no second agent on a GitHub runner and the same failure appears there at 20k scale. The fragility is in the fixture.

Root cause, reproduced rather than reasoned about. `tests/conftest.py`'s `_snapshot_cache_bucket` does:

```python
bucket = _SNAPSHOT_CACHE_BUCKETS.get(basetemp)
if bucket is None or not bucket.is_dir():
    bucket = Path(tempfile.mkdtemp(prefix="snapshot-caches-", dir=basetemp))
```

The `is_dir()` guard handles the bucket disappearing, but not its **parent**. If the worker's basetemp goes away, `mkdtemp(dir=basetemp)` raises `FileNotFoundError` — and because the fixture is autouse and nothing recreates the parent, *every remaining test in that worker errors identically*, each reporting a different freshly-generated bucket name. That is exactly what the logs show, and it is why the count equals "all remaining tests" rather than a handful. I reproduced the transition directly against the allocator's own logic:

```
allocated: snapshot-caches-5sr2e_uo
FileNotFoundError: .../snapshot-caches-5vpz5m8d     # a NEW name — it is the parent that is missing
```

So the fixture converts a *transient* directory deletion, from whatever source, into a whole-session failure. Proposed patch, which fixes the class rather than the deleter:

```python
basetemp = tmp_path_factory.getbasetemp()
bucket = _SNAPSHOT_CACHE_BUCKETS.get(basetemp)
if bucket is None or not bucket.is_dir():
    basetemp.mkdir(parents=True, exist_ok=True)   # the parent may have gone too
    bucket = Path(tempfile.mkdtemp(prefix="snapshot-caches-", dir=basetemp))
    _SNAPSHOT_CACHE_BUCKETS[basetemp] = bucket
```

with a regression test that removes the basetemp mid-session and asserts the next allocation succeeds (the invariant: the allocator recovers from its directory tree vanishing, rather than a test pinned to one deleter).

I am **not** pushing that here: it is unrelated to this PR's diff, it is red on the base branch rather than caused by this change, and widening a perf-harness PR with a repo-wide test-infrastructure fix is the wrong shape. It also shouldn't wait on this PR — `main` is failing its two Linux unit lanes right now. Happy to put it up as its own change the moment someone says go.

This PR's own status is unaffected: the nine perf suites pass locally on `8c0ce7ee` (597 tests), and the tests failing above are in `tests/unit/storage/` and `tests/test_action_report_*`, which this branch does not touch.

---

PR #1272, opened 2026-09-13
fix(tests): recover the autouse cache allocator when its temp tree vanishes

napetrov:
## Summary

`main`'s two Linux unit lanes are failing with tens of thousands of errors from a single root cause in `tests/conftest.py`. This fixes it.

## Motivation

On run [34729579282](https://github.com/abicheck/abicheck/actions/runs/34729579282): `unit-tests (ubuntu-latest, 3.12)` reported **2 failed, 20,867 errors** and 3.14 reported **29,164**, every one of them

```
FileNotFoundError: .../pytest-of-runner/pytest-0/popen-gwN/snapshot-caches-XXXXXXXX
```

out of the autouse `_isolate_snapshot_cache` fixture.

`_snapshot_cache_bucket` guards against its *bucket* disappearing, but not against the bucket's **parent**. Once a worker's basetemp was gone, `mkdtemp(dir=basetemp)` raised, nothing recreated the tree, and because the fixture is autouse every remaining test in that worker failed on entry — each reporting a *different* freshly-generated bucket name, which is what disguised one root cause as a flood of unrelated ones and explains why the error count equals "all remaining tests".

I found this while watching #1267's CI and [wrote up the diagnosis there](https://github.com/abicheck/abicheck/pull/1267#issuecomment-5650081083) rather than widening that perf PR with it. Worth recording honestly: I had previously hit the same signature locally, attributed it solely to an unrelated process pruning `/tmp` in my container, and said so — the deletion there was real, but calling it only external interference was wrong, since a GitHub runner has no such process and fails identically.

## Changes

- `_snapshot_cache_bucket` recreates the parent (`mkdir(parents=True, exist_ok=True)`) before allocating. That makes recovery independent of **what** removed the tree — pytest's own retention policy, a tmp reaper, a sandbox cleanup and a stray `rmtree` all arrive here identically, so keying the fix to any one deleter would leave the rest open. `exist_ok` rather than a prior existence check, since check-then-create reintroduces the window being closed.
- `TestTheAllocatorSurvivesItsDirectoryTreeVanishing` in `tests/test_conftest_cache_isolation.py`.
- New bug class `test_infra.autouse_allocator_cannot_recover` in `tests/regressions/manifest_guards.py` — a distinct root cause from the neighbouring shared-state class, which is about a differential test being served its own cached result.

## Bug-fix test contract

- Bug class: `test_infra.autouse_allocator_cannot_recover` — a fixture every test enters converts a transient failure of its own environment into a failure of the whole session.
- Publicly observable failure: tens of thousands of `FileNotFoundError` errors across a full unit lane, one per remaining test, each naming a different directory, from one lost directory.
- Regression test fails on base: `TestTheAllocatorSurvivesItsDirectoryTreeVanishing::test_allocation_recovers_after_a_deletion[basetemp]` and `[ancestor]` — verified by stashing the `conftest.py` change and re-running: exactly those two fail, 33 pass.
- Negative control: `test_repeated_allocation_without_deletion_reuses_one_bucket` — a heal-by-always-allocating version would pass the recovery cases while silently discarding the per-process bucket reuse the allocator exists for (the reason it was introduced: keeping pytest's numbered allocator off a directory with thousands of siblings).
- Public-surface test: none, and the registry entry claims none — the seed test inspects a pytest fixture's own allocator directly.
- Axes covered: which level of the tree was removed (the bucket, the basetemp, an ancestor above it); recovery properties (fresh directory, empty, reused when intact).
- General invariant: *allocation must succeed again after any part of the tree above it is removed, and must still reuse one bucket when the tree is intact.* Exercised across the three removal levels rather than the one observed deletion, with an oracle that is not the implementation: `test_the_deletion_really_breaks_the_unfixed_allocator` reproduces the pre-fix allocator verbatim and requires it to raise, so the recovery cases cannot pass because a deletion happened to be harmless.

## Checklist

- [ ] Docs updated if user-visible behaviour changed — not applicable, test infrastructure only
- [x] `pre-commit run --all-files` passes locally

### Local verification

**2,212 tests pass in the exact lanes CI failed** (`tests/unit/storage`, `tests/test_action_report_query.py`, `tests/test_action_report_destinations.py`), plus 35 in the isolation and regression-manifest suites. `check_ai_readiness.py` and `check_architecture.py` both 0 errors; `ruff check tests/ scripts/` clean. The fix was verified in both directions — the two new cases fail without the `conftest.py` change and pass with it.

https://claude.ai/code/session_01Vee2ztjM4Utc7E3fVNjsEN

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vee2ztjM4Utc7E3fVNjsEN)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resilience of test runs when temporary directories are unexpectedly removed.
  - Automatically restores missing temporary-directory levels while preserving privacy, ownership, symlink, and permission safeguards.
  - Prevents unrelated system directories from being modified during recovery.
  - Reuses the existing temporary allocation after recovery instead of creating unnecessary replacements.

- **Tests**
  - Added coverage for recovery after removing the cache directory, its parent, or the surrounding temporary tree.
  - Added validation for permissions, ownership, symlink handling, and restrictive umask scenarios.

### DIFF
--- a/changelog.d/20260913_claude_autouse_cache_allocator_recovery.md
+++ b/changelog.d/20260913_claude_autouse_cache_allocator_recovery.md
@@ -13,3 +13,7 @@
   -- since these paths are predictable and sit on a temp root shared between
   users, so a `parents=True` recreate under the process umask would leave the
   hierarchy traversable and accept a path planted in the deletion window.
+  Those checks apply only within pytest's own `pytest-of-<user>` root: asserting
+  ownership or repairing modes on ancestors above it would error every test on a
+  runner that does not own the shared temp directory, and as root would strip that
+  directory's world-writable mode machine-wide.
--- a/changelog.d/20260913_claude_autouse_cache_allocator_recovery.md
+++ b/changelog.d/20260913_claude_autouse_cache_allocator_recovery.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- `tests/conftest.py`'s autouse snapshot-cache fixture now recreates pytest's
+  basetemp before allocating, so a transient deletion anywhere in the temp tree
+  no longer fails every remaining test in the worker. The bucket's own
+  existence check covered the bucket disappearing but not its parent, so once a
+  worker's basetemp was gone every subsequent test errored with a
+  `FileNotFoundError` naming a different freshly-generated bucket — two Linux
+  unit lanes reported 20,867 and 29,164 such errors from one root cause. Bug
+  class: `test_infra.autouse_allocator_cannot_recover`.
--- a/changelog.d/20260913_claude_autouse_cache_allocator_recovery.md
+++ b/changelog.d/20260913_claude_autouse_cache_allocator_recovery.md
@@ -8,3 +8,8 @@
   `FileNotFoundError` naming a different freshly-generated bucket — two Linux
   unit lanes reported 20,867 and 29,164 such errors from one root cause. Bug
   class: `test_infra.autouse_allocator_cannot_recover`.
+  The recovery recreates each level with pytest's own guarantees -- mode `0o700`,
+  a refusal on a symlinked or foreign-owned level, and the same loose-mode fixup
+  -- since these paths are predictable and sit on a temp root shared between
+  users, so a `parents=True` recreate under the process umask would leave the
+  hierarchy traversable and accept a path planted in the deletion window.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import os
 import shutil
+import stat
 import subprocess
 import sys
 import tempfile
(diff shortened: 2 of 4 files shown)


napetrov:
P2 fixed in `abe5da28`, and it was a good catch on a fix that was itself a weakening.

`mkdir(parents=True, exist_ok=True)` was the wrong tool for precisely the reason given: these are *predictable* paths (`/tmp/pytest-of-<user>/pytest-N/popen-gwM`) on a temp root shared between users, and pytest's own `getbasetemp` creates each level `0o700`, raises on a symlinked root, raises on a foreign-owned one, and clears group/other bits on a directory it owns (I read that source rather than assuming it). A `parents=True` recreate under the process umask leaves the hierarchy world-traversable, and `exist_ok` accepts whatever was planted in the window between the deletion and the recovery — so the recovery would have been the vulnerability.

The recovery now reproduces those checks: each level created individually (no `parents=True`, so the mode applies to all of them), `chmod` after `mkdir` because `mkdir` masks its mode with the umask and `chmod` does not, pre-existing ancestors validated rather than adjusted into place — except the one group/other fixup pytest also performs — and the uid/symlink checks skipped only where the platform cannot express them, as pytest skips them.

Five tests, one per property, because a single "recovery works" test passes while any one is absent:

| property | test |
|---|---|
| every recreated level is private | `test_every_recreated_level_is_private` (checks all three levels — an open ancestor makes the leaf reachable) |
| a symlinked level is refused | `test_a_symlinked_level_is_refused` |
| a foreign-owned level is refused | `test_a_foreign_owned_level_is_refused` (needs root to create; skipped otherwise) |
| a loose existing mode is tightened | `test_a_loose_mode_on_an_existing_level_is_tightened` |
| the umask cannot loosen a recreated level | `test_the_umask_cannot_loosen_a_recreated_level` — sets `umask(0)` explicitly, since under an ordinary umask a mode-only `mkdir` passes and the chmod's necessity would go unproven |

**All five fail against the `parents=True` version and pass with this one**, verified by substituting the old line back in: 5 failed, 35 passed. The bug class's axes gained `privacy` and `hostile_environment` to record what the seed test now covers.

Verification on this head: 2,212 tests pass in the lanes CI failed (`tests/unit/storage`, `tests/test_action_report_*`), 40 in the isolation and manifest suites, both gates 0 errors, ruff clean.

On the CI failures arriving for this PR: all of them are on `8c0ce7ee`, the pre-fix sha (now `main`) — the Linux unit lanes are the very bug this PR fixes, and the macOS/Windows ones are the `TestIsReleaseStyleOperand` family that fails identically on clean `main` (`action/run.sh` delegating to `is_package`, commit `65473b26`). Nothing from them is actionable here; what matters is whether those two Linux lanes come back green on this head, which is the CI-level proof of the fix.

One environment note so the commits are not misread: they show as Unverified. The committer email is correct (`noreply@anthropic.com`); signing is impossible in this container because the configured `gpg.ssh.program` (`/tmp/code-sign`) was deleted by the same unrelated process that prunes `/tmp`, and the signing pubkey is a 0-byte file — a test commit with `commit.gpgsign=true` fails outright. Every commit on this branch's predecessors was identically unsigned and merged fine through #1262 and #1267.

---

napetrov:
P1 fixed in `935b889b` — and it was not theoretical. `abe5da28` broke a real macOS job within minutes of being pushed, with exactly the predicted error:

```
OSError: refusing to use /private/var/folders/36: it is owned by another user
```

(`Native PE/Mach-O compare workflows (macos-latest)`, 11 errors — every test in the job, since the failure is in an autouse fixture.) So my hardening commit reintroduced the same shape as the bug this PR exists to fix: one bad decision in a fixture every test enters takes the whole job with it. Worth saying plainly — that is two rounds in a row where the *fix* was the defect.

`_pytest_owned_levels` now bounds the region to pytest's own `<temproot>/pytest-of-<user>` root and everything under it — exactly what pytest creates, owns and privacy-checks itself. Levels above it are created plainly when missing and never validated or chmod'ed. On the failing path that means `/private/var/folders/36` is no longer touched at all, while `pytest-of-runner` and below still get the full treatment. The marker is found by name rather than by comparing against `tempfile.gettempdir()`, because `PYTEST_DEBUG_TEMPROOT`, a `TMPDIR` changed mid-session and a resolved symlink (`/var` → `/private/var`, as here) all move the temp root without moving the marker. With `--basetemp` there is no marker and pytest creates exactly one directory, so the owned region is that path alone.

Six tests, with both halves of the finding stated in their real shapes:

| case | test |
|---|---|
| a foreign-owned ancestor must **not** raise (the macOS failure above) | `test_a_foreign_owned_system_ancestor_does_not_raise` |
| a `01777` sticky ancestor keeps its mode **and** sticky bit (the destructive half) | `test_a_sticky_world_writable_ancestor_keeps_its_sticky_bit` |
| a system ancestor's mode is untouched | `test_a_system_ancestors_mode_is_left_alone` |
| the marker directory itself is still validated (boundary inclusive) | `test_the_pytest_root_itself_is_still_validated` |
| `--basetemp` owns only that directory | `test_with_an_explicit_basetemp_only_that_directory_is_owned` |
| the boundary as a pure function | `test_the_owned_region_is_the_marker_and_everything_under_it` |

Four of them fail against the unbounded version. I also checked my own container for damage and could prove there was none rather than hope: the fixup only clears group/other bits, so `01777` would become `01700` with the sticky bit intact, and `/tmp` here is `755` *without* a sticky bit — a mode that code cannot produce. `/tmp` is still `755` after a full default-basetemp run.

One trap recorded in the test, because it caught me: the `--basetemp` case built under `tmp_path` is never actually marker-free, since by default `tmp_path` lives inside `/tmp/pytest-of-<user>/pytest-N/`. It passed only while I was running with `--basetemp` pointed outside `/tmp`, and failed the moment the same test ran under pytest's default root. It now allocates its own marker-free root and asserts that premise before relying on it.

**The original fix is confirmed at CI level:** the check suite on `b09c7147` completed with nothing failing, i.e. the two Linux unit lanes that produced 20,867 and 29,164 errors on `8c0ce7ee` are green with the allocator recovery in place.

Local verification on this head: 46 tests in the isolation suite pass under **both** the default and an explicit `--basetemp`; the regression-manifest suite passes; gates 0 errors; ruff clean.

---

PR #1282, opened 2026-09-13
refactor: subtractive cleanup batch 2 — one owner per duplicated helper

napetrov:
## Summary

Follow-on to #1279. That batch removed the delegation-only *facades*; this removes duplicated *implementations* underneath them. **−282 net lines of package implementation**, of which −116 is executable structure and the rest is five copies of one Mach-O explanation collapsing into one.

## Motivation

Batch 1's survey left a question open: after the facades went, was there a pool of genuinely dead or duplicated code left? I answered it mechanically rather than by guessing — hashing every function body under `abicheck/` and every module-level definition's reference count across the whole repo.

The answer is that the dead-code well is nearly dry (four real finds), but **duplicated implementations are not**: 16 groups of byte-identical bodies. Per this repo's own rule that "a duplicate test body is a review queue, not a deletion list", I read all 16 rather than deleting on the hash — which is what caught the three that must not be merged.

## Changes

**Seven helpers deduplicated onto one owner**, every caller importing it.

The largest had five copies: the Mach-O Itanium leading-underscore strip in `buildsource/call_graph.py`, `type_graph.py`, `macro_graph.py` (with `override_graph.py` importing one of them). `model/mangled_name.py` already declared itself *"the single canonical home for this specific structural check — previously duplicated, independently, by … so a Mach-O-target identity fix made in one no longer needs to be separately rediscovered and reapplied in the other two."* The consolidation was declared but never finished. The copies' stated reason — *"kept as a local copy … both sit near the AI-readiness line-count cap"* — was void: importing the owner makes both files smaller, not larger.

Their real knowledge is preserved, not deleted with the fifth copy: the `asm("__Zfake")` gap (clang reports that literal spelling on any platform, so the strip corrupts a hand-labelled symbol's identity) is carried into the canonical docstring, where it now applies to every caller instead of two of five.

The other six: `layer_payload_empty`, `_change_covers_symbol`, `_freeze`, `_type_identifiers`, `_split_top_level_commas`, `_resolve_under_root`, `_frozen_tuple`.

**Four functions deleted as unreferenced:** `internal_leak._typenode_is_indirection_wrapper`, `frontends.cli.runtime._validate_show_only`, `report.render_markdown.render_severity_sections`, and `post_processing._safe_index`. The last was found by the dedup, not the dead-code scan — it was never called in its own module and had a single test as its only consumer, which now exercises the live `diff_filtering` implementation it duplicated.

## What was deliberately *not* deduplicated

This is the part that needed reading rather than hashing:

- **`_deadline_bound_worker`** — identical bodies, **different type signatures** (one typed for `SourceAbiTu`, one for type-graph results). Deduping needs a `TypeVar`, which adds structure rather than removing it. I applied it, mypy caught it, I reverted it.
- **`policy/selectors._strip_template_args`** — its copy documents a real reason: importing `internal_leak` would pull `checker_types`/`buildsource` in transitively.
- **`_is_intel_sycl_driver` / `_is_dpcpp_family_binary`** — same body, different names, and the docstring already says they are deliberately two functions over one name set for two distinct call sites.
- **`_node_file`** — would create a brand-new cross-package edge for 14 lines.
- **`_linker_identity` / `_canonical_decl_key`** — a one-line body (`return mangled or name`) whose two docstrings explain different call-site concerns. A one-liner cannot meaningfully drift.

Where a pair was mutually dependent (`appcompat` ↔ `appcompat_consumer_impact`, `storage/dto` ↔ `types_section_codec`), ownership went to the module the other already imports. My first attempt picked the wrong direction in both cases and created import cycles; both were caught by the test suite and reversed, so **no new import edge or cycle is introduced**.

## Line accounting

| Area | + | − | Net |
|---|---:|---:|---:|
| Package implementation (`abicheck/**/*.py`) | 49 | 331 | **−282** |
| Tests | 6 | 6 | 0 |
| Changelog fragment | 20 | 0 | +20 |
| **Total** | **75** | **337** | **−262** |

Of the −282: **−116 is executable structure**; the remainder is comment and docstring text that existed only because the same explanation was written five times. Three oversized files shrank (`appcompat.py` 1714→1701, `internal_leak.py` 1652→1620, `type_graph.py` 1995→1958 — the last within 5 lines of the hard cap).

## Validation

Against recorded base `b0be72f3`.

- `verify.py --profile pr --only lint,fmt-check,typecheck,ai-readiness,architecture,docs-contract,repo-facts,schema-sync` — **8 passed, 0 failed**.
- `mypy abicheck/` — Success, 792 files, baseline 0 held. (This is what caught the `_deadline_bound_worker` signature mismatch.)
- `check_architecture.py` — 0 errors, including no new cycle after both direction reversals.
- Full unit lane: **47,523 passed, 46 failed**. 43 reproduce on the recorded base; two are the `test_action_report_destinations` / `test_action_run_sh_compare_pr_json_write` modules that fail only under `-n 4`, with a different parametrization each run (established in #1279, pass serially); one was mine (`_safe_index`) and is fixed in this branch.
- Real compiled C fixture re-run — unchanged pair exit 0, breaking pair exit 4 on the same three change kinds. The Mach-O dedup touches mangled-name normalization, so this is behavioural coverage and not just a type check.

## Remaining candidates

Unchanged from #1279's list, minus the strict-YAML item (now known blocked), plus one new:

- **Export convergence (`expand_export_kwargs`)** — `secondary_writes` spans 9 files including the refusal/error output paths, where a missing secondary artifact is a correctness bug. Needs its own slice with export-on-refusal tests.
- **`checker_policy`/`contract_gating`/`reclassify`** — blocked on `model`-owned `DiffResult` reaching policy. Note most of what `checker_types` imports through them are *enums* (shapes, arguably `model`'s), but `apply_policy_file_overrides`/`policy_kind_sets` are real policy, so moving the enums alone does not unblock it.
- **Storage convergence** — assessed again this batch and still not started. `storage/import_v1.py` is on the live write path, not dead; replacing it means a direct `AbiSnapshot`→sections encoder plus regenerated baselines, with persisted user artifacts at stake. Not a one-session change.
- **Schema mirror** (`abicheck/schemas/` ↔ `docs/reference/schemas/v1/`) — all 5 files byte-identical, 4,353 tracked duplicate lines. Removable by generating the docs copy at build time, but that is docs-workflow surgery on the publish path, and #1279 already showed this repo has CI jobs whose constraints are invisible locally. Wants its own PR. Hygiene, not runtime code.

## Checklist

- [x] Docs updated (n/a — no user-visible surface changed)
- [x] Lint / format / mypy / architecture / ai-readiness all clean

https://claude.ai/code/session_01KTGETYEXyHHGnEep1CrUqg

---
_Generated by [Claude Code](https://claude.ai/code/session_01KTGETYEXyHHGnEep1CrUqg)_

## Summary by CodeRabbit

* **Refactor**
  * Consolidated shared helper logic to improve consistency across analysis, build, reporting, and storage workflows.
  * Removed several unused internal helpers without changing supported public interfaces.
  * Centralized symbol normalization and type-identity handling for more consistent results across build analysis.
* **Documentation**
  * Clarified limitations around Mach-O symbol normalization.
* **Tests**
  * Updated tests to use the consolidated helper implementations.

### DIFF
--- a/abicheck/appcompat.py
+++ b/abicheck/appcompat.py
@@ -31,6 +31,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from .appcompat_consumer_impact import (
+    _change_covers_symbol,
     attach_consumer_impact,
     consumer_impact_explanations,
     enrich_covered_changes,
@@ -584,20 +585,6 @@ def _is_relevant_to_app(change: Change, app: AppRequirements) -> bool:
             return True
 
     return False
-
-
-def _change_covers_symbol(change: Change, symbol: str) -> bool:
-    """Does *change* already account for *symbol* (exact, demangled, or via
-    ``affected_symbols``)? Mirrors :func:`_is_relevant_to_app`'s matching in
-    reverse -- symbol-name lookup, not app-requirements lookup."""
-    if change.symbol == symbol:
-        return True
-    from .demangle import demangle as _demangle_symbol
-
-    plain = _demangle_symbol(change.symbol)
-    if plain and plain == symbol:
-        return True
-    return bool(change.affected_symbols and symbol in change.affected_symbols)
 
 
 def uncovered_missing_symbols(
--- a/abicheck/buildsource/baseline_publish.py
+++ b/abicheck/buildsource/baseline_publish.py
@@ -46,30 +46,12 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
-from .build_output import BuildOutput, BuildOutputTarget
+from .build_output import BuildOutput, BuildOutputTarget, _resolve_under_root
 
 #: Default ``key_prefix`` (ADR-047 section 10) when a baseline channel's
 #: ``.abicheck.yml`` entry doesn't set one -- matches the ADR's own example
 #: (``key_prefix: "abicheck-baseline-main"``).
 DEFAULT_ACCEPTED_MAIN_KEY_PREFIX = "abicheck-baseline-main"
-
-
-def _resolve_under_root(root: Path, rel: str) -> Path | None:
-    """Resolve *rel* under *root*, refusing an absolute path or an escape.
-
-    A local copy of :func:`~.build_output._resolve_under_root`'s identical
-    guard -- mirrors :mod:`~.baseline_set`'s own precedent of keeping this
-    small safety check duplicated per-module (see
-    ``_resolve_under_baseline_dir``'s docstring there) rather than importing
-    a leading-underscore helper across a module boundary.
-    """
-    if Path(rel).is_absolute():
-        return None
-    candidate = (root / rel).resolve()
-    root_resolved = root.resolve()
-    if candidate != root_resolved and not candidate.is_relative_to(root_resolved):
-        return None
-    return root / rel
 
 
 @dataclass
--- a/abicheck/buildsource/call_graph.py
+++ b/abicheck/buildsource/call_graph.py
@@ -55,6 +55,7 @@ from ..model.graph_facts import (
     GraphEdge,
     GraphNode,
 )
+from ..model.mangled_name import strip_macho_itanium_decoration
 from ..model.source_graph import function_decl_identity
 from .adapters.base import source_from_argv
 from .clang_ast_run import run_clang_ast_dump
@@ -177,46 +178,9 @@ class CallEdge:
 def _identity(node: dict[str, Any]) -> str:
     """Stable callee/caller identity: the mangled name when clang emits one
     (encodes the full signature, keeps overloads distinct), else the name."""
-    return _normalize_mangled(str(node.get("mangledName") or node.get("name") or ""))
-
-
-def _normalize_mangled(mangled: str) -> str:
-    """Strip a spurious macOS Mach-O ABI leading underscore from an Itanium
-    mangled name clang reports (``__ZN...`` -> ``_ZN...``).
-
-    On Darwin, clang's own AST dump reports a C++ decl's ``mangledName`` with
-    the platform's extra linker-symbol-table underscore still attached (the
-    same decoration ``macho_metadata.py`` strips when parsing the *binary*
-    export table) -- but the ``Function``/``Variable`` objects this identity
-    must join against (``header_graph._decl_identity``, seeded from the flat
-    ``AbiSnapshot``) carry the already-stripped, one-underscore form, since
-    that normalization already happened upstream in ``macho_metadata.py``/
-    ``dumper._dump_macho``. Left unstripped, a header-only graph's
-    call/type-graph node for a public *function* (as opposed to a type --
-    types don't get this decoration) never joins its ``SOURCE_DECLARES``-
-    seeded, provenance-tagged counterpart, so it can never be recognized as
-    a public graph entry (``is_public_dependency_node``) -- silently
-    dropping every ``PUBLIC_API_INTERNAL_DEPENDENCY_ADDED`` finding rooted
-    at a function on macOS. ``__Z`` is an unambiguous, platform-independent
-    marker (a real Itanium mangled name always starts with ``_Z``; a literal
-    C++ identifier starting with two underscores is reserved and never
-    emitted here), so this is a no-op on Linux/Windows, where clang's
-    ``mangledName`` is already the bare ``_Z...`` form.
-
-    **Known gap, not fixed here** (self-review round, fresh evidence): the
-    "no-op on Linux/Windows" claim above does not hold for an explicit GNU
-    ``asm("__Zfake")`` label -- clang reports that literal spelling
-    verbatim on *any* platform, confirmed empirically. Called
-    unconditionally (unlike :mod:`template_graph`'s own
-    ``_normalize_mangled``, whose join now tries the exact spelling first
-    and only falls back to this strip -- see that module's
-    ``_resolve_emitted_symbol``), this corrupts such a decl's identity here
-    too, silently failing (or mis-joining) the same way. Porting the same
-    guarded-fallback fix to this module's own join
-    (:func:`augment_graph_with_calls`) needs its own scoped change, not a
-    drive-by extension of this docstring.
-    """
-    return mangled[1:] if mangled.startswith("__Z") else mangled
+    return strip_macho_itanium_decoration(
+        str(node.get("mangledName") or node.get("name") or "")
+    )
 
 
 def _function_identity(node: dict[str, Any], scope: list[str]) -> str:
@@ -240,7 +204,7 @@ def _function_identity(node: dict[str, Any], scope: list[str]) -> str:
     type_obj = node.get("type")
     type_qual = str(type_obj.get("qualType", "")) if isinstance(type_obj, dict) else ""
     return function_decl_identity(
-        _normalize_mangled(str(node.get("mangledName") or "")),
+        strip_macho_itanium_decoration(str(node.get("mangledName") or "")),
         name,
         qualified_name,
         type_qual,
--- a/abicheck/buildsource/macro_graph.py
+++ b/abicheck/buildsource/macro_graph.py
@@ -203,6 +203,7 @@ from functools import partial
 from typing import TYPE_CHECKING, Any
 
 from .. import deadline
+from ..model.mangled_name import strip_macho_itanium_decoration
 from .clang_ast_run import run_clang_ast_dump
 from .graph_facts import CONF_HIGH, CONF_REDUCED, GraphEdge
 from .preprocessor_facts import _DEFINE_RE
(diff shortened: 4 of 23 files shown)


napetrov:
**`unit-tests (ubuntu-latest, 3.12)` is red on `main`, not from this PR — fix ported in `0ec5fd7c`.**

The failure event arrived on this PR's subscription but names `head_sha 634986ec`, which is #1279's head, already merged. Two of its eight failures are `tests/test_backend_capability_matrix.py::test_generated_matrix_is_in_sync` and `::test_check_mode_passes_on_the_committed_file`, and I verified both reproduce on a **clean `origin/main` checkout** — they were also in the base-failure set I recorded before #1279 was written, so neither batch introduced them.

**What it is:** one stale row in a generated file. `Function.is_inline`'s note was extended in `scripts/gen_backend_capability_matrix.py` without `docs/reference/header-backend-capabilities.md` being regenerated, so the sync test fails. Nothing to do with the helper dedup in this PR.

**Why it's ported here rather than left:** a base-branch failure with an existing fix gets carried into the PR so this branch's own unit lanes can go green; it no-ops once main carries the same regeneration. The change is `python scripts/gen_backend_capability_matrix.py`'s own output byte for byte — the remedy `docs/AGENTS.md` documents under "Regenerating generated docs" — not a hand edit. `test_backend_capability_matrix.py` now passes 21/21 locally, with `ai-readiness` and `docs-contract` still clean.

**Worth flagging separately:** `main`'s 3.12 lane reported **8** failures and I have only accounted for these 2. The other 6 are not visible in the log tail I could retrieve, and I have not chased them since they are not this PR's and I have no evidence they are related. If that lane is expected to be green, main needs its own look.

---

napetrov:
**Follow-up to my earlier comment: `main`'s integration lane is red too, and I can now name the cause. Not this PR's, and not ported — it needs a decision I shouldn't make unilaterally.**

The `integration-tests (ubuntu-24.04)` failure event also names `head_sha 634986ec` (#1279's merged head), and its log is 1,740 bytes of post-job cleanup — the job died before any test body ran. So I ran the integration lane locally instead, with real `castxml` + `gcc`.

**Result: 7 failed, 577 passed** on this branch — and **the same 7, by identical test ID, on a clean `origin/main` worktree**. Batch 2 introduces zero integration regressions.

**Root cause — a retired flag with 7 tests left behind.** `compare --output-dir` no longer exists (`abicheck compare --help` has no such option; it was folded into the `-o FORMAT=DESTINATION` export grammar). These still pass it:

- `test_release_assurance_cli.py` ×5 — invoke `--output-dir` and assert exit 4; they now get exit 64, `No such option '--output-dir'. Did you mean '--output'?`
- `test_cli_compare_bundle_facts.py` ×2 — assert the collision error contains `"collide"`; the current message ends `…nesting the two means one export's document lands on the other's. Give each its own location.`

**Proposed patch, for whoever retired the flag:** migrate the five release-assurance invocations to the export grammar (`-o json=<dir>/` or whichever destination shape replaced the per-component directory), and decide the two collision cases deliberately — either restore `collide` to the message or update the assertion to the current wording. I'm not pushing that here: picking between "the message regressed" and "the test is stale" is a judgement about someone else's change, and the alternative — relaxing the assertion to make it pass — is exactly what a test like this exists to prevent.

**Net picture of `main` right now**, closing the loop on the "8 failures, 2 accounted for" I left open earlier:

| Lane | Failures | Status |
|---|---|---|
| `unit-tests (ubuntu, 3.12)` | 8 | 2 = stale generated matrix, **fixed and ported here** (`0ec5fd7c`). 6 unidentified. |
| `integration-tests (ubuntu-24.04)` | 7 | All 7 = the retired `--output-dir`, diagnosed above. Not ported. |

Both predate this PR and #1279. This branch's own diff is clean against them.

---

napetrov:
**Quantified against the correct base: this branch is strictly better than `main` on every unit lane that has reported.**

The remaining red checks here are `main`'s, not this PR's. Measured by test counts rather than inferred:

| Lane | Base `b0be72f3` (current `main`) | This PR `0ec5fd7c` | Δ |
|---|---|---|---|
| `unit-tests (macos-latest, 3.13)` | 8 failed / 46,700 passed | **6 failed / 46,702 passed** | **−2 failed, +2 passed** |
| `unit-tests (windows-latest, 3.13)` | 27 failed / 45,982 passed | **25 failed / 45,984 passed** | **−2 failed, +2 passed** |

Both deltas are exactly the two `test_backend_capability_matrix` tests the regeneration in `0ec5fd7c` fixed. No lane gained a failure.

For the record, `main`'s unit lanes have been red since before either cleanup batch: the CI run on `44f48dda` — the commit #1279 branched from — already shows `failure` on all five (`ubuntu 3.12/3.13/3.14`, `macos`, `windows`).

The six remaining macOS failures are `test_model_package_surface` ×2, `test_real_world_false_positives`, `test_subprocess_bash_is_resolved`, and `test_debug_evidence_presence` ×2. I confirmed the first four reproduce on a clean `origin/main` worktree; the last two I attribute to the base by the count arithmetic above rather than by reading them in the base log, and this branch changes no debug-info detection or Mach-O parsing.

One methodological note in case anyone repeats this check: comparing against `44f48dda` instead of `b0be72f3` makes Windows look like a **+6 regression**, because `main` gained ~120 tests across #1277 and #1279 in between. The base to diff against is the PR's actual merge base.

---

napetrov:
**`mutmut (detector core)` is red, and unlike the other lanes this one *is* this PR's — but it needs a maintainer action I can't take. Not a surviving mutant.**

Unlike the unit/integration lanes, mutmut doesn't run on `main` pushes (it's PR-triggered by a diff touching an `only_mutate` module), so there's no base run to compare against. I traced it instead.

**Why it triggered:** exactly one file in this diff is in `[tool.mutmut].only_mutate` — `abicheck/surface_graph.py`. Its entire change is two import lines:

```diff
+from .compare.surface_graph import _type_identifiers
-from .surface import _type_identifiers
```

forced by `_type_identifiers` moving to its owner. I checked by AST: **zero functions** in that file have a changed body or signature.

**Why it fails:** that's a *module-scope* edit, which `check_mutation_score.py` documents as having "no mutant of its own for `check_diff_scoped()` to attribute — by design, it is scored by the per-module drift gate instead." That gate needs `mutation-baseline.json`, which is **not in the repo** (`AGENTS.md`: "not recorded yet; until a dispatch run writes it, the drift lanes fail closed rather than passing vacuously"). So: `check_diff_scoped` has nothing to attribute → `check_per_module` has no baseline → the gate correctly refuses to exit 0.

It is working as designed. This PR is just the first to hit the combination of *touches a mutated module* and *module-scope-only edit*.

**What unblocks it — a maintainer call, not mine:** a workflow dispatch of the mutation lane with `write_baseline: true`, which records `mutation-baseline.json`. `AGENTS.md` is explicit that this is deliberately manual: "accepting the current survivor set is a review decision, not something a cron does silently." I'm not going to make that decision by proxy, and the alternatives are all worse — weakening or bypassing the gate is out, and the import rewrite can't be avoided without abandoning the dedup that motivates this PR.

I have a re-run still unspent, but I'm not spending it: the failure is deterministic, and a second run reaches the identical state.

---

napetrov:
Merged `main` (`2cb1c857`) in. One conflict, in `abicheck/report/render_markdown.py`: this branch deletes `render_severity_sections` as unreferenced, and main meanwhile added a `rollups` field to `ChangeGroup` plus a `render_kind_rollups(group.rollups)` line *inside that same function*.

I re-checked the deletion against the new main rather than just re-applying it, and it still holds: on `2cb1c857` nothing calls `render_severity_sections`. `reporter_markdown.compute_severity_sections` is called only by `report/render_markdown_document.py`, which projects each group itself (`heading`/`oneline`/`note_lines`/`rows`) and renders from that dict — so the function was already dead on main too. Resolved by keeping the deletion, and dropped the now-unused `render_kind_rollups` import (`KindRollup` is still used by the `rollups` field).

Worth flagging separately, because the conflict is what surfaced it and a reviewer could otherwise read it as caused by this PR: **`rollups` is computed on main but never rendered.** `reporter_markdown.compute_severity_sections` fills `risk_rollups`/`hygiene_rollups`/`quality_rollups`, and the only place `render_kind_rollups` was ever called from is the dead function above — the live document path drops the field. That is a pre-existing gap on `main`, present before this merge and unchanged by it; this PR neither creates nor fixes it, and fixing it is a behaviour change that belongs in its own PR, not in a subtractive cleanup.

Validated on the merge commit (`b3a49d1c`): `verify.py --profile pr --only lint,fmt-check,typecheck,architecture` 4 passed / 0 failed, `mypy abicheck/` clean over 806 files, `check_architecture.py` 0 errors, `ai-readiness` 0 errors, and the report/markdown/rollup test modules (110 tests) pass.

---

PR #1285, opened 2026-09-13
fix(tests): resolve three new bare-bash sites, and land #1255's last review round

napetrov:
## Summary

`tests/test_subprocess_bash_is_resolved.py` is failing on `main`, on every unit lane. Two separate causes, both fixed here.

## The gate is working

Three bare `["bash", …]` argv sites landed after #1255 merged, and the gate named them:

- `tests/test_l2_real_profile_status.py:298, 1032`
- `tests/test_l2_real_profiles.py:614`

This is the class #1255 exists to close — new modules written without the convention — caught mechanically instead of by a red Windows lane nobody attributes correctly. Resolved through `bash_executable()` and guarded with `require_bash()` in the established shape.

## #1255's final review round never reached main

Worth stating plainly, since I reported otherwise: **#1255 merged at `4963307f7`, two minutes before `58ba691f2` was pushed.** That last commit — the three CodeRabbit fixes — stayed on the branch. It is cherry-picked here:

- **`_dominating_guard_line` read only `func.id`** while `_names_called` accepted `func.attr`, so a function guarding with `_workflow_exec.require_bash()` and resolving with `_workflow_exec.bash_executable()` was reported unguarded.
- **An over-eager guard** in `test_the_two_public_functions_agree_on_this_machine`, inserted by the migration sweep into a test that never shells out.
- **Unnamed parse failures**: a malformed module raised `SyntaxError` against `"<unknown>"`, failing all three rules without saying which file.

## Two further findings from review of this PR

- **The de-guarded assertion was itself wrong.** Removing the guard (correct as far as it went) exposed that `have_bash() is not is_wsl_launcher_stub(bash_executable())` holds on only two of three states: on a genuinely bash-less host `have_bash()` is `False` while the resolver returns its documented `"bash"` fallback, which is not a stub. De-guarding would have turned a skip into a *failure* on exactly the host the de-guarding was for. Restated over all three states.
- **The guard's receiver was unchecked**, so an unrelated `helper.require_bash()` could stand in for the real one.

## Not addressed here

`test_backend_capability_matrix`'s two failures on the same lanes are a stale generated file from an unrelated change — `python scripts/gen_backend_capability_matrix.py` regenerates it.

## Bug-fix test contract

- Bug class: `guard.platform_convention_without_a_gate` (registered in `tests/regressions/manifest_guards.py` by #1255) — a platform-portability rule enforced by a shared helper other modules must remember to import, rather than by a scan for the banned shape. This PR is the class's first live instance: new violations appeared within hours of the gate landing, and the gate named them.
- Publicly observable failure: `unit-tests` red on every lane (Linux, macOS, Windows), with `test_no_test_module_passes_a_bare_bash_program` naming three modules. Before the gate existed the same defect surfaced only as an unattributed red `windows-latest` lane, where a bare `"bash"` resolves to the WSL launcher stub and the whole calling module fails at once with no bash-level diagnostic.
- Regression test fails on base: yes, and that is how this PR was found rather than a claim made afterwards — `test_no_test_module_passes_a_bare_bash_program` fails on `b0be72f3e` (main's tip), naming `test_l2_real_profile_status.py` [298, 1032] and `test_l2_real_profiles.py` [614]. The two review findings are pinned the same way: `wrong_receiver` returns `[]` under the old `func.attr`-only check, and the old two-state assertion evaluates `False is True` when `_real_bash()` is stubbed to `None` (verified by simulating that state, not by reasoning about it).
- Negative control: the non-firing half of each rule, asserted alongside the firing one — `shutil.which("bash")`, prose and comments naming the banned identifier, `"bash"` in a non-program argv position, `args=[exe, "bash"]`, `cmd = [exe, "bash"]` traced but not the program, a name bound from a parameter or call (genuinely untraceable), a nested helper carrying its own guard, and a correctly guarded function. Without these, each rule could "pass" by rejecting everything — for the guard rule that would mean demanding a guard in every test that merely mentions the resolver, which is exactly the over-eager insertion this PR removes.
- Public-surface test: none, deliberately. This is test-suite infrastructure reaching no shipped CLI, typed-API or report surface, which is why the registered bug class carries `public_surfaces=()`. The equivalent rigor here is that all three rules run against the real `tests/` tree rather than fixtures alone, and that whole-suite collection (48598 tests, zero errors) is the check matching a change to a module's exported surface — the check whose absence let #1255 ship two modules dead at collection.
- Axes covered: host (`posix`, `nt`, and the third state — no real bash at all); argv shape (inline literal, `args=` keyword, name traced from `ast.Assign`, name traced from `ast.AnnAssign`, untraceable name); guard shape (bare call, qualified by `_workflow_exec`, wrong receiver, placed after the call, inside `if`/`try`/loop, in a nested closure); scope (module body, function, nested function). Not covered: a real `windows-latest` runner, which no environment available here can provide — CI on this PR is the only source of that.
- General invariant: three rules over every module under `tests/`, each exercised with generated shapes rather than the reported inputs — (1) no bare `"bash"` reaches a subprocess program position, by any argv spelling; (2) the retired `_bash_executable` name appears nowhere, as definition, import, call or attribute; (3) a function that resolves *and* shells out is dominated by a real `require_bash()`. Plus the resolver's own three-state contract: `have_bash()` true ⇒ `bash_executable()` returns a real, runnable, non-stub program; false ⇒ it returns the documented fallback and claims nothing. The oracle is independent of the implementation in each case — the scans are AST walks written against Python's grammar, not against the resolver, and the three-state assertion checks `shutil.which`/`Path.is_file` rather than re-deriving `_real_bash`'s own answer.

## Verification

Gate at 35 tests, `test_l2_real_profile_status.py` 116 passed, `test_workflow_exec_harness.py` and the workflow-exec consumers green. Pinned-`ruff` clean. Whole-suite collection: **48598 tests, zero collection errors**.

As with #1255, the Windows lane itself can only be verified by CI here.

## Checklist

- [x] Changelog fragment — n/a, no `abicheck/**/*.py` change
- [x] Docs updated — n/a

### DIFF
--- a/tests/test_l2_real_profile_status.py
+++ b/tests/test_l2_real_profile_status.py
@@ -37,6 +37,7 @@ from _l2_real_profiles_support import (
     measurement_output,
     profiles,
 )
+from _workflow_exec import bash_executable, require_bash
 
 
 class TestReadinessIsNotAMeasurement:
@@ -288,6 +289,7 @@ class TestSideAcquisitionIsDeclared:
         string a text assertion would look for while passing for an empty
         directory.
         """
+        require_bash()
         script = profiles.prepare_script(profiles.SVS)
         guard = [
             line
@@ -296,7 +298,7 @@ class TestSideAcquisitionIsDeclared:
         ]
         assert guard, "no guard was generated for the supplied side"
         return subprocess.run(
-            ["bash", "-c", "set -u\n" + "\n".join(guard)],
+            [bash_executable(), "-c", "set -u\n" + "\n".join(guard)],
             cwd=tmp_path,
             capture_output=True,
             text=True,
@@ -1014,6 +1016,7 @@ class TestASuppliedDistributionMustProveItsIdentity:
         guard line failed to parse, and a text assertion would have reported the
         expected strings present throughout. Only running it caught that.
         """
+        require_bash()
         script = profiles.prepare_script(profiles.SVS)
         guard = "\n".join(
             line
@@ -1030,7 +1033,7 @@ class TestASuppliedDistributionMustProveItsIdentity:
                 f'#define SVS_RUNTIME_VERSION_STRING "{version}"\n'
             )
         result = subprocess.run(
-            ["bash", "-c", "set -u\n" + guard],
+            [bash_executable(), "-c", "set -u\n" + guard],
             cwd=tmp_path,
             capture_output=True,
             text=True,
--- a/tests/test_l2_real_profiles.py
+++ b/tests/test_l2_real_profiles.py
@@ -40,6 +40,7 @@ from _l2_real_profiles_support import (
     materialize_operands,
     profiles,
 )
+from _workflow_exec import bash_executable, require_bash
 
 
 class TestProfileDefinitions:
@@ -610,9 +611,14 @@ class TestPrepareScriptIsolatesEachCommand:
     def test_the_rendered_script_is_valid_shell(self, profile_id):
         # Parse-checked rather than eyeballed: a script nobody can run is not
         # reproduction instructions. `bash -n` needs no network and builds nothing.
+        require_bash()
         script = profiles.prepare_script(profiles.PROFILES[profile_id])
         proc = subprocess.run(
-            ["bash", "-n"], input=script, capture_output=True, text=True, check=False
+            [bash_executable(), "-n"],
+            input=script,
+            capture_output=True,
+            text=True,
+            check=False,
         )
         assert proc.returncode == 0, proc.stderr
 
--- a/tests/test_subprocess_bash_is_resolved.py
+++ b/tests/test_subprocess_bash_is_resolved.py
@@ -53,7 +53,9 @@ call site, and only the *first* element of an argv list is the program.
 from __future__ import annotations
 
 import ast
+from collections.abc import Callable
 from pathlib import Path
+from typing import TypeVar
 
 import _workflow_exec
 import pytest
@@ -63,6 +65,8 @@ TESTS_DIR = Path(__file__).resolve().parent
 
 #: The retired private clone. Banned as a name, not merely as a definition.
 _CLONE_NAME = "_bash_executable"
+
+T = TypeVar("T")
 
 
 #: Callables whose first positional argument is an argv sequence.
@@ -83,6 +87,31 @@ def _test_modules() -> list[Path]:
     module that defines them.
     """
     return sorted(TESTS_DIR.rglob("*.py"))
+
+
+def _scan(rule: Callable[[str], list[T]], path: Path) -> list[T]:
+    """Apply one scan to one file, naming the file if it will not parse.
+
+    `ast.parse` reports a `SyntaxError` against `"<unknown>"`, so a malformed
+    module added under `tests/` would fail all three rules at once without
+    saying which file it was (CodeRabbit review). Re-raised rather than
+    skipped: an unparseable module is a real problem, and one this scan is
+    well placed to notice.
+    """
+    try:
+        return rule(path.read_text(encoding="utf-8"))
+    except SyntaxError as exc:
+        raise AssertionError(
+            f"{_display(path)} does not parse, so it cannot be scanned: {exc}"
+        ) from exc
+
+
+def _display(path: Path) -> str:
+    """`path` relative to `tests/` when it lives there, else in full."""
+    try:
+        return path.relative_to(TESTS_DIR).as_posix()
+    except ValueError:
+        return str(path)
 
 
 def _argv_program(node: ast.Call) -> ast.expr | None:
@@ -241,11 +270,20 @@ def _unguarded_resolutions(source: str) -> list[str]:
         hoist a guard it already has.
         """
         for stmt in getattr(fn, "body", []):
-            if (
-                isinstance(stmt, ast.Expr)
-                and isinstance(stmt.value, ast.Call)
-                and getattr(stmt.value.func, "id", None) == "require_bash"
-            ):
+            if not (isinstance(stmt, ast.Expr) and isinstance(stmt.value, ast.Call)):
+                continue
+            # Both call forms, matching how `_names_called` reads a
+            # resolution: reading only `func.id` here while accepting
+            # `func.attr` there made the two halves disagree, so a function
+            # guarding with `_workflow_exec.require_bash()` was reported
+            # unguarded (CodeRabbit review).
+            func = stmt.value.func
+            name = (
+                func.attr
+                if isinstance(func, ast.Attribute)
+                else getattr(func, "id", None)
+            )
+            if name == "require_bash":
                 return stmt.lineno
         return None
 
@@ -313,7 +351,7 @@ class TestNoModuleSpellsBashItself:
     def test_no_test_module_passes_a_bare_bash_program(self) -> None:
         offenders = {}
         for path in _test_modules():
-            hits = _bare_bash_call_sites(path.read_text(encoding="utf-8"))
+            hits = _scan(_bare_bash_call_sites, path)
             if hits:
                 offenders[path.relative_to(TESTS_DIR).as_posix()] = hits
         assert offenders == {}, (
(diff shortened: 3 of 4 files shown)


napetrov:
Both P2s addressed in `b1652c2f0`. The first is a correction to the previous round, not an addition to it, so it's worth stating in full.

**"Distinguish an absent bash from a non-stub executable."** Correct, and I verified it by simulating the state rather than reasoning about it:

```
_real_bash() -> None:  have_bash() False | bash_executable() 'bash' | not is_wsl_launcher_stub('bash') True
```

So `have_bash() is (not is_wsl_launcher_stub(bash_executable()))` is `False is True` — the de-guarded test *fails* on a genuinely bash-less host. CodeRabbit was right that the inserted guard retired the assertion on the machines it exists for; what neither of us noticed is that the assertion underneath it was a two-state claim in a three-state world, so removing the guard converted a skip into a failure on that same host.

Restated over what the two functions actually promise, which holds in all three states and skips in none: when `have_bash()` reports a bash, `bash_executable()` hands back a real, runnable, non-stub one; when it does not, the resolver hands back its documented `"bash"` fallback and claims nothing.

**"Verify the guard belongs to `_workflow_exec`."** Also right, and the asymmetry is the point: `_dominating_guard_line` accepted any `*.require_bash()`, so an unrelated `helper.require_bash()` could stand in for the real guard. The two directions fail differently — a stricter *resolution* rule only asks for a guard that is already present, while a looser *guard* rule permits the subprocess — so the receiver is checked on the guard side only. A bare name stays trusted (the module-level import binds it, and nothing here can see past a deliberately shadowed import without a symbol table). Wrong-receiver case pinned.

Gate at 35 tests; whole-suite collection 48598, zero errors.

---

napetrov:
## CI status on `b1652c2` — `unit-tests (macos-latest, 3.13)` is red, and it is not this PR's

Compared against the same job on main's tip (`b0be72f3e`, run 34743117784), which is the check that settles ownership:

| | main | this PR |
|---|---|---|
| failures | **8** | **7** |
| `test_subprocess_bash_is_resolved.py::test_no_test_module_passes_a_bare_bash_program` | ❌ | ✅ |
| `test_backend_capability_matrix.py` ×2 | ❌ | ❌ |
| `test_debug_evidence_presence.py` ×2 | ❌ | ❌ |
| `test_model_package_surface.py` ×2 | ❌ | ❌ |
| `test_real_world_false_positives.py` ×1 | ❌ | ❌ |

The set is identical **minus the one this PR fixes**. So the change does what it claims and adds nothing: the gate now passes on macOS, having named the three new bare-bash sites on main.

The seven that remain are all pre-existing on main and none are in code this PR touches:

- `test_backend_capability_matrix` ×2 — a stale generated file; `python scripts/gen_backend_capability_matrix.py` regenerates it.
- `test_debug_evidence_presence` ×2, `test_model_package_surface` ×2, `test_real_world_false_positives` ×1 — unrelated changes in the model/evidence area.

No fix exists for any of them that I could port into this PR, and manufacturing one here would widen a test-infrastructure change into someone else's area. I have not spent the one re-run: these reproduce identically on main, so a re-run would tell us nothing.

`unit-tests (windows-latest, 3.13)` is still in progress. That is the lane this whole line of work is about, and it has never yet reported on any revision carrying the fix — #1255 merged before its run completed, and my own pushes cancelled the two runs after it. I will report what it says either way.

---

napetrov:
## `unit-tests (windows-latest, 3.13)` — the verification #1255 never got

**main: 27 failed · this PR: 15 failed.** Same run configuration, same job, base `b0be72f3e` vs head `b1652c2f0`.

The twelve that disappear are exactly the ones the bash fix targets:

| gone on this PR | count |
|---|---|
| `test_l2_real_profile_status.py` (`TestASuppliedDistributionMustProveItsIdentity`, `TestSideAcquisitionIsDeclared`) | 7 |
| `test_l2_real_profiles.py::test_the_rendered_script_is_valid_shell[onedal, pvxs, svs, svs_pr_base]` | 4 |
| `test_subprocess_bash_is_resolved.py::test_no_test_module_passes_a_bare_bash_program` | 1 |

Those eleven `test_l2_*` failures were the WSL launcher stub doing precisely what #1255 describes: three bare `["bash", …]` argv sites at `test_l2_real_profile_status.py:298,1032` and `test_l2_real_profiles.py:614`, each resolving to `%SystemRoot%\System32\bash.exe`, printing UTF-16LE text and exiting 1 — so every test in those two classes failed at once with no bash-level diagnostic. Resolving the three call sites removes all eleven.

That is the first end-to-end confirmation on a real `windows-latest` runner that this whole line of work does what it claims. #1255 merged before its own CI reported and my subsequent pushes cancelled two more runs, so until now the Windows behaviour had never actually been measured on any revision carrying the fix.

The remaining 15 are a strict subset of main's 27 — this PR introduces no failure of its own. All are pre-existing and none touch code it changes:

- `test_backend_capability_matrix` ×2 — stale generated file; `python scripts/gen_backend_capability_matrix.py`
- `test_conftest_cache_isolation` ×6 — Windows POSIX-permission semantics
- `test_action_release_operand_detection` ×2, `test_debug_evidence_presence` ×2, `test_model_package_surface` ×2, `test_real_world_false_positives` ×1

Same picture on macOS (8 → 7, detailed in the comment above): the failure set shrinks by exactly what this PR fixes and gains nothing.

---

napetrov:
## Every lane, base vs head

| lane | main `b0be72f3e` | this PR `b1652c2f0` | the gate |
|---|---|---|---|
| unit-tests (windows-latest, 3.13) | 27 failed | **15** | ❌ → ✅ |
| unit-tests (ubuntu-latest, 3.14) | 23 failed | **21** | ❌ → ✅ |
| unit-tests (ubuntu-latest, 3.13) | 8 failed | **5** | ❌ → ✅ |
| unit-tests (ubuntu-latest, 3.12) | — | **7** | (absent) ✅ |
| unit-tests (macos-latest, 3.13) | 8 failed | **7** | ❌ → ✅ |

Every lane is lower, and `test_no_test_module_passes_a_bare_bash_program` — red on every main lane — passes on every one here.

### One PR-only failure, and why it is not this PR's

`test_action_coverage_verdict.py::TestCompareTellsTheTwoAxesApart::test_the_default_legacy_scheme_has_no_severity_block_to_read` fails on ubuntu-3.14 here and not on main. That module is one this PR migrated, so it deserves a real answer rather than a flake label:

- The change to it is `["bash", str(RUN_SH)]` → `[bash_executable(), str(RUN_SH)]` plus a `require_bash()`. On Linux `bash_executable()` returns `"bash"` and `require_bash()` is a no-op, so the argv and the control flow are **identical**. Nothing there can turn a `COVERAGE_INCOMPLETE` verdict into `SEVERITY_ERROR`.
- The module fails on *both* base and head, with **a different test each run**:
  - main 3.14 → `TestADemotedBreakStaysVisible::test_the_step_still_passes_because_the_policy_said_so[compare]`
  - main 3.12 → `TestADemotedBreakStaysVisible::test_a_compatible_text_verdict_is_not_read_either`
  - this PR 3.12 → `test_the_step_still_passes_because_the_policy_said_so[compare]`
  - this PR 3.13 → none at all

A rotating membership across runs on both sides of the diff is nondeterminism inside that module, not a consequence of this change. The same rotation appears in the neighbouring Action/workflow harness modules (`test_action_report_destinations`, `test_action_release_topology_config`, `test_reusable_workflows_*`) — all shell out through `action/run.sh`, all fail with a different member each run, on main as well.

That looks like a pre-existing flakiness worth its own investigation, and I have deliberately not folded it into this PR: it is unrelated to bash resolution and would widen a test-infrastructure change into the Action harness. I have not spent the one re-run on it either — re-running would only re-roll the same dice, and the cross-lane evidence above is stronger than a re-run would be.

Stable across every lane and both sides: `test_model_package_surface` ×2, `test_real_world_false_positives` ×1, `test_backend_capability_matrix` ×2 (stale generated file). None are in code this PR touches.

---

napetrov:
### `integration-tests` (macos-latest, windows-latest) — identical on both sides

Both are red here and both fail with a **byte-identical** set to the same job on main `b0be72f3e`:

- macOS: `test_cli_compare_added_public_header_live.py::test_reported_invocation_reports_the_added_api`, `test_debug_evidence_presence.py::…[-g-True]`
- Windows: `test_debug_evidence_presence.py::…[-g-True]`, `…[-g0-False]`

Same tests, same parameters, base and head. Neither touches bash resolution or anything this PR changes, and no fix exists that I could port. Not re-run: an identical reproduction on main is already the evidence a re-run would be looking for.

That completes the CI picture — every failing check on this PR is now accounted for as pre-existing on main, and the unit lanes are strictly better than main on all five.

---

PR #1286, opened 2026-09-13
fix: four defects in the merged MKL-evaluation work, found after it landed

napetrov:
## Summary

Defects in the work merged as #1283, none of which the merge carried a fix for, plus the review rounds since. Two came from a review that landed on the branch three minutes before the merge; two were in a commit the merge did not pick up; the rest are this PR's own review rounds. All of them are on `main`, so this is a fresh change rather than a continuation of the merged PR.

## Motivation

#1283 merged at head `d2f7d2d10`. My last push, `90586aa88`, was not in it, and a further Codex round arrived on `d2f7d2d10` three minutes before the merge. This branch was restarted from `main` with that unmerged commit rebased on.

## Changes

**1. A weak import that becomes strong was treated as pre-existing.** `import_existed_in_old` matched on library and version only, so a *weak* OLD import counted as history. It is not: the loader resolves an unresolved weak symbol to `0`/`NULL`, so OLD having loaded says nothing about anything outside the bundle providing that symbol. A weak-to-strong transition is a genuinely new requirement and was being suppressed on the vacuously-satisfied outward-edge rule — a candidate that now fails to load could come back `NO_CHANGE`.

**2. `<gcc_options>` was parsed with `str.split`.** A shell-valid quoted argument (`-I"/opt/Program Files/inc"`) became fragments with quote characters still attached. It was also the one place that did not agree with `join_gcc_options`, the emitter those same values pass through — two grammars for one string. Both now share `split_gcc_options`.

**3. A symbol that changes ELF type is not an addition.** The undeclared-export detector subtracted OLD's exports *within each symbol class*, so a name going `STT_OBJECT` → `STT_FUNC` read as newly gained: `func_added_elf_only` alongside the `symbol_type_changed` that already described the real change, corrupting the addition count and `recommend_release`'s MINOR bump.

**4. Descriptor header skips are recorded, with the rule they were matched by.** `<skip_headers>`/`<skip_including>` narrow the parsed surface like `--exclude-header` but were filtered without being recorded, so the comparability gate saw two empty exclusion sets and let an asymmetric pair through. Recording the patterns alone then proved insufficient twice over: the same text is not the same scope under both rules (`include/foo.h` excludes `/pkg/include/foo.h` natively via the `*/<pattern>` branch, and is kept under a descriptor's exact membership). The matching rule is now recorded as the fact it is (`excluded_header_matching`, schema v48).

**5. The exclusion gate and its warning were two different comparisons.** The gate compared sets, the warning compared tuples; a native run's CLI order against a sorted record was accepted as comparable and then reported as having differing exclusions. Both now call one predicate.

## Bug-fix test contract

- **Bug class:** three recurrences of *two implementations of one rule, drifted* (descriptor parser vs. its emitter; the comparability gate vs. its warning; exact vs. glob header matching), and two of *an equality that omits the field carrying the semantics* (`import_existed_in_old` ignoring `weak`; the export diff subtracting within one partition when the question spans all of them).
- **Publicly observable failure:** a bundle whose consumer's weak import became strong reports `NO_CHANGE` though it no longer loads; a descriptor with a quoted include path compiles against nonexistent paths; a variable turned into a function of the same name is counted as an addition; a descriptor-narrowed dump compares against a full baseline and reports removals; a comparable pair is told its findings may be scope artefacts.
- **Regression test fails on base:** yes. #3 reproduced on real `gcc -shared` binaries (both `symbol_type_changed` and `func_added_elf_only`, now only the former); #4 reproduced end to end (`compat dump` recorded `()`, now `('b.h',)` with `matching="exact"`, and the pair is refused rather than scored); #2 and #5 by running the parser and the warning on the reported inputs.
- **Negative control:** each fix has one. `test_a_strong_old_import_is_history` (a predicate returning `False` always would satisfy the weak-import claim completely); `test_a_genuinely_new_name_is_still_reported`; `test_unquoted_whitespace_still_tokenises` (the behaviour the shell grammar exists for); `test_two_descriptor_runs_still_compare` (refusing everything would satisfy the cross-rule refusal claim); `test_a_real_difference_is_still_reported` (a warning path that never claimed a difference would pass the reordering test).
- **Public-surface test:** yes, for every fix that has a CLI surface. #2, #4 and the matching-rule recording are driven through the real `compat dump` command with `CliRunner`, asserting what reaches `dump()` and what lands in the written snapshot on disk — not the helpers alone; #4's consequence is then asserted through the real `check_contracts_comparable` gate. #3 was reproduced through `dumper.dump` + `checker.compare` on real compiled binaries. #1 is the stated exception: it is exercised at the predicate, because `test_bundle._meta` has no route to a weak undefined symbol (see "Known unsupported cases").
- **Axes covered:** ELF only, on Linux/CPython 3.13, at the header-aware (L2) and binary (L0) evidence tiers. Schema axis: snapshot v47 → v48, with the pre-v48 default asserted to load as `"glob"` — correct for every existing snapshot, since the native path was the only producer that existed. Both matching rules (`glob`, `exact`) are covered, in both directions, including the same-rule and no-pattern cases. PE/Mach-O were not exercised: no toolchain here, and the shapes reproduced (zero-`DT_NEEDED`, ELF symbol types) are ELF-specific. The `compat` XML/HTML projections of `library` are asserted against the real renderers rather than a golden file.
- **General invariant:** the ELF-type test parametrizes *every* transition direction, since the defect is symmetric in the two symbol classes and one example forecloses half of it; `test_the_rule_is_exactly_old_weakness` states the weak rule as a parametrized equivalence; `test_the_two_never_disagree` states gate-refuses ⟺ warning-claims-difference over symmetric and asymmetric inputs; `test_join_and_split_are_inverses` states the round trip over several shapes rather than the one reported path.
- **False-positive removed / real break preserved:** both. The FP-rate gate stays at 0 FP / 0 FN and the per-tier accuracy gate stays top-tier-correct and under-call monotonic.
- **Known unsupported cases:** the weak/strong transition is exercised at the predicate, not through a compiled fixture, because the ELF fixture helper has no route to a weak undefined symbol; widening it is a separate change. Whether real ABICC globs in `<skip_headers>` is unverified and recorded in `docs/contribute/known-gaps.md` — the alternative fix (make descriptor skips use `fnmatch`) would change drop-in behaviour on a guess about ABICC's semantics, which needs evidence from the real tool.

## Notes

`codecov/patch` reports a stale figure for the same reason it did on #1283: `ci.yml`'s Codecov upload step has no `if: always()`, so a red canonical lane skips the upload and Codecov keeps serving an older report. The one-line fix is deliberately not included here — it is shared CI and not this PR's subject.

Two latent test-oracle bugs on `main` were found while verifying this work and are **not** fixed here (proposed patches in the PR comments): Python 3.14 changed `PurePath.suffix` for leading-dot names, breaking `test_cache_header_walk.py`'s oracle; and `test_dedup_key.py`'s "a value keys the same as an equal copy" is false for `nan`. Happy to open both as one test-only PR.

## Checklist

- [x] Changelog fragment updated
- [x] `ruff`, `ruff format`, `mypy`, `check_architecture.py` (0 errors) and the FP-rate gate all clean locally

https://claude.ai/code/session_01A2Qt7wTDEcViRqGjGwjery


napetrov:
Merged current `main` in (it had moved four merges ahead) and pushed `e1dd87237`. Two things worth recording, one actionable by me and one not.

**Fixed: my own test was over-pinned.** `test_library_is_declared_in_the_report_schema` asserted `REPORT_SCHEMA_VERSION == "4.6"`, which is not what the test is about — the claim is that the emitted `library` key is *declared* and *optional*. Main reached 5.0 and the pin failed on a bump that has nothing to do with this field. Removed; the two assertions carrying the claim are unchanged. A version pin belongs with whatever change owns the bump.

The merge also picks up #1285, which fixes the bare-`bash` sites I had been reporting as red-on-`main`, so that one is off the list.

**Not this PR's: `unit-tests (ubuntu-latest, 3.14)` is red on `main` for a Python 3.14 pathlib change.** This woke me as a failure on #1286, but the run was against `d2f7d2d10` — the already-merged head, attributed here only because the branch pointer moved. 20 failures, and the bulk are one root cause:

Python 3.14 changed `PurePath.suffix` for leading-dot names — `PurePath("..h").suffix` is `".h"` on 3.13 (measured) and `""` on 3.14. `tests/test_cache_header_walk.py` uses the *live* `PurePath.suffix` as its oracle (line 79 builds the expectation with `p.suffix`; line 224 asserts `PurePath("..h").suffix == ".h"` outright), so the oracle moved under it while `extract/cache_header_scan._path_suffix` kept the behaviour PR #1275 deliberately gave it.

Both files come from `dd3a77a62` and this branch touches neither, so it is not mine to fix inside this PR, and no open PR carries a fix to port — #1286 is currently the only one open. Proposed patch, for whoever owns it:

> The behaviour is right and the oracle is wrong. The cache key *must* treat `..h` as a `.h` header — that was #1275's fix, and a stdlib revision does not change what the cache key needs. What has to change is the test: state the intended rule directly (last dot at an index > 0 and not trailing) instead of delegating to a stdlib function whose semantics are version-dependent. That is this repo's own "the oracle must be an independent derivation, not the same helper" principle, applied to a helper that happens to live in the stdlib. `_path_suffix`'s docstring also needs its claim about `splitext` vs `PurePath` narrowed to "≤ 3.13", since on 3.14 they agree.

Happy to open that as its own PR if you'd like — it is a two-file change and clearly separable from this one.

CI on this PR's real head is still settling; I have not re-run anything, and the one re-run allowance is unspent.

---

napetrov:
P1 verified and fixed in `4ecfbf13f`. You're right, and it is sharper than the summary suggests — this is the same failure the field was added to prevent, arriving from the opposite direction. Last round it was *a request recorded that was never achieved*; this is *two different achievements recorded identically*.

Confirmed in the code: `_resolve_headers_from_list` filters with `h.name not in skip_headers and str(h) not in skip_headers` — exact membership — while `extract.header_exclusions` uses `fnmatch` against three spellings. So `*.h` excludes every header natively and none through a descriptor, and both wrote `("*.h",)`.

Only what the matching rule can actually achieve is recorded now. The part I want to flag, because it is where a careless fix would go wrong: **this must not over-refuse.** A pattern with no metacharacter means the same thing under both rules, so a descriptor and a native run naming `b.h` are genuinely comparable and must stay so — blanket-tagging descriptor skips as a distinct namespace would have refused that pair for no reason. Only a metacharacter-bearing pattern is dropped, because only that one narrowed nothing.

Verified end to end on real inputs, both directions:

| | descriptor `<skip_headers>` | native `--exclude-header` |
|---|---|---|
| `*.h` | records `()`, excludes neither header | records `("*.h",)`, excludes both |
| `b.h` | records `("b.h",)`, excludes `b.h` | records `("b.h",)` |

The `*.h` pair is now refused by the real gate; the `b.h` pair still compares equal. Both asserted in tests, the second as the explicit anti-over-refusal control.

Also: silently doing nothing with a rule the user wrote is its own failure, and is what made this look like working configuration, so both commands now say so. Recording and reporting live in one function — what gets recorded and what must be reported are exact complements, and splitting them is how a third caller ends up doing one and not the other, which is this area's recurring bug shape.

**One thing I deliberately did not do,** and I'd rather flag it than bury it: your first suggested remedy — make descriptor skips use `fnmatch` too — would make both paths agree and may well be what real ABICC does. I didn't take it, because it changes ABICC drop-in behaviour on a *guess* about ABICC's own semantics. That's a parity claim needing evidence from the real tool (an `abicc`-marked comparison), not a comparability fix smuggled into a review round. It's recorded in `docs/contribute/known-gaps.md` with what would settle it. If you know ABICC globs there, say so and I'll make the change properly with a parity test.

Unit lane: 48,657 passed, 8 failed, all verified not this branch's. The CI failures that woke me in between were all against `d2f7d2d10` — the already-merged head, attributed here only because the branch pointer moved — and the 3.14 one is covered in my previous comment.

---

napetrov:
P2 verified and fixed in `e64f4ebc9`, plus one latent `main` flake I found while checking it.

**The gate and the warning were two different comparisons.** Reproduced exactly: `('z.h','a.h')` vs `('a.h','z.h')` — the gate returns `None` (comparable) and the warning says the exclusions differ and the findings may be scope artefacts. The tool contradicting itself on the same field.

Both now call one predicate, `model.header_exclusion_record.exclusions_are_symmetric`. Sharing rather than just correcting the second copy is deliberate: this is the **third** time in this workstream that two implementations of one rule drifted — the descriptor parser vs. its emitter, and now the gate vs. this warning — and a second *correct* copy is still a second copy.

Two details worth stating, since a careless fix gets both wrong. Reordering now produces the **symmetric** disclosure ("headers matching … were excluded, on either side") rather than silence — both sides genuinely are narrower than their operands name, so saying nothing would be the opposite error. And the message renders sorted now, for the same reason the comparison ignores order. The test states the invariant over symmetric *and* asymmetric inputs (gate-refuses ⟺ warning-claims-difference), with a negative control, since a warning path that never claimed a difference would satisfy the reordering case completely.

---

**Separately — a latent flake on `main`, not this PR's.** While verifying the above I hit `tests/unit/compare/test_dedup_key.py::TestEqualValuesKeyEqually::test_a_value_keys_the_same_as_an_equal_copy`:

```
assert nan == nan
 +  where nan = hashable_value(nan)
 +  and   nan = hashable_value(deepcopy(nan))
Failing test case: value=nan
```

Hypothesis found `nan`, which is never equal to itself — so the invariant "a value keys the same as an equal copy" is simply false for it. `hashable_value` is behaving correctly; the *property* is stated too strongly.

This branch touches neither `tests/unit/compare/` nor `abicheck/compare/dedup_key.py`. It passed once on a fresh `origin/main` worktree, but that proves nothing here: Hypothesis only finds `nan` sometimes, and it is sticky in my checkout solely because `.hypothesis/` is a gitignored local database. On CI this is a latent random failure of a `main` test that will fire whenever the search happens to reach `nan`.

Proposed patch, for whoever owns it:

> Exclude non-reflexive values from the strategy, or state the invariant in the form that is actually true — `hashable_value(v)` equals `hashable_value(deepcopy(v))` *whenever `v == v`*. The second is better: it keeps NaN in the search space and documents that the dedup key inherits Python's own equality semantics rather than pretending to repair them.

Happy to open this together with the Python-3.14 `PurePath.suffix` one from my earlier comment as a single small "two latent test-oracle bugs on main" PR — both are test-only, both are oracles asserting something stronger than what is true.

Unit lane: 48,665 passed, 10 failed, all verified not this branch's.

---

napetrov:
P1 confirmed and fixed in `eb77d139a`, and you were right that my previous fix was still a guess.

Reproduced your counterexample exactly before changing anything:

| | `--exclude-header include/foo.h` | `<skip_headers>include/foo.h` |
|---|---|---|
| `/tmp/pkg/include/foo.h` | **excluded** (the `*/{pat}` branch) | **kept** |

No metacharacter anywhere, two different scopes, one recorded text. So "no metacharacter ⇒ same meaning under both rules" is simply false for any pattern containing a path separator.

That is the **second** falsification of a rule inferred from the pattern text — first recording the raw text for both, then dropping only metacharacter-bearing patterns. This repo has an explicit discipline for exactly that situation ("attempted twice, reverted twice": stop proposing a third heuristic and record the limitation or the fact instead), so the third attempt is not another text rule. The matching mode is recorded as the fact it is — `AbiSnapshot.excluded_header_matching`, schema v48 — and two sides are comparable only if they narrowed by the same patterns *under the same rule*. A pre-v48 snapshot loads as `"glob"`, which is correct for every one of them rather than a default, because the native path was the only producer that existed when v47 shipped.

**The cost, stated plainly:** a descriptor `b.h` and a native `b.h` do achieve the same thing and are now refused anyway. I took that deliberately — nothing in the patterns alone proves which pairs are equivalent, which is precisely what your counterexample demonstrates, and refusing a comparable pair costs a run while accepting an incomparable one manufactures findings. `test_two_descriptor_runs_still_compare` is the control that keeps this from degenerating into refusing everything. The mismatch reason now names both rules, since with identical pattern text "the scopes differ" leaves a user nothing to act on.

**Separately, `test-contract` was failing on my own PR body** — a real check I'd been overlooking while reading the CI failures on stale heads. Two required rows, "Public-surface test" and "Axes covered", were genuinely unanswered: I had written that section with my own row names. Now answered properly, including the honest exception (the weak-import fix is exercised at the predicate, because the ELF fixture helper has no route to a weak undefined symbol) and the axis limits (ELF/Linux/3.13; PE and Mach-O not exercised, no toolchain here). The check passes against the new body.

Schema-bump fallout handled: `docs/reference/snapshot-format.md` and the G20 catalog fixtures regenerated.

Unit lane: 48,669 passed, 9 failed — all verified not this branch's, including the two latent `main` oracle bugs from my earlier comments.

---

PR #1289, opened 2026-09-13
docs(catalog): refresh the clang validation-status row after #1283

napetrov:
## Summary

Fix the remaining red workflow on `main`: `Examples Validation`, failing on a stale `catalog/README.md` status row.

## Motivation

`Examples Validation` has been failing on `main` since #1283 — that was the first run after it that wasn't cancelled, so the breakage sat unnoticed behind a string of cancelled runs. `check_examples_validation_status_sync.py` reports:

```
was: gcc: 164 PASS / 5 XFAIL / 39 SKIP; clang: 165 PASS / 5 XFAIL / 38 SKIP
now: gcc: 164 PASS / 5 XFAIL / 39 SKIP; clang: 164 PASS / 6 XFAIL / 38 SKIP
```

## Changes

- `catalog/README.md`: the default/debug verdict row's clang figures, plus a note in that row recording why the number moved.

## Why not just regenerate the table

A PASS → XFAIL is a verdict **mismatch** that a pre-existing `known_gap` tolerates, so refreshing the number blindly can record a genuine regression as the new normal. I identified the case first.

Bisecting the merges since the last passing run (`b0be72f3`) over the known-gap case set: the flip is `case180_symbol_binding_lost_unique` under clang, and it enters at `2cb1c857` (#1283). Locally, `b0be72f3` gives clang PASS=1/XFAIL=5 over that set and current `main` gives PASS=0/XFAIL=6, with `case180` the single difference — `expected=COMPATIBLE_WITH_RISK, got=NO_CHANGE`.

**It is not a detection regression.** Under clang there is no `STB_GNU_UNIQUE` signal for this case to detect at all — the binding is a GCC ELF extension, and the case's own `known_gap` says precisely that. So the `COMPATIBLE_WITH_RISK` it used to produce cannot have come from the symbol-binding path; it came from a persistent cross-source hygiene finding, which #1283 deliberately stopped counting toward the verdict ("Pre-existing cross-source hygiene debt no longer downgrades the verdict", its own changelog fragment). The case was passing for an unrelated reason, `NO_CHANGE` is the honest answer under clang, and the gap annotation had already anticipated exactly this outcome.

## Verification

`check_ai_readiness.py` clean (0 errors), and the four catalog/validation-status tests pass. The status-sync check itself is CI-only — it consumes the full-matrix artifacts from the sharded jobs — so the authoritative confirmation is this PR's own `Examples Validation` run; the figures written here are the ones CI itself reported as current.

## Checklist

- [ ] Tests added / updated for new behaviour — none; this corrects a recorded observation, no behaviour changes
- [ ] Changelog fragment — not required, no `abicheck/**/*.py` change

https://claude.ai/code/session_01R5qaFHsKcS2rvXzakcEJQB

---
_Generated by [Claude Code](https://claude.ai/code/session_01R5qaFHsKcS2rvXzakcEJQB)_


napetrov:
**`unit-tests (ubuntu-latest, 3.13)` failed on `3aa66a08`** (the previous head of this branch, already merged as #1287). Recording what it was, since it is not this PR's and not a new break.

Two failures, both `FileNotFoundError`, in tests unrelated to either diff:

- `test_action_coverage_verdict.py::TestCoverageGatedTrustsAReadableZero::test_an_unreadable_report_is_not_attributed_to_this_axis`
- `test_reusable_workflows_require_complete_analysis.py::TestAssuranceOverlayPreservesUsableDiscoveredCompileDb::test_relative_sources_root_still_resolves`

The coverage floor was met on that same run (96.29% ≥ 95%), so the gate itself is fine.

Both run `action/run.sh` as a subprocess with `cwd=tmp_path`, and `FileNotFoundError` there means the directory went away mid-run. That signature is already documented in this repo: `TestTheAllocatorSurvivesItsDirectoryTreeVanishing` in `tests/test_conftest_cache_isolation.py` was written for exactly it, after "a single transient deletion under pytest's temp tree" produced 20,867 errors on one Linux lane and 29,164 on another (run 34729579282). Its own docstring states the deleter is not identifiable and deliberately does not pin one — "pytest's retention policy, a tmp reaper, a sandbox cleanup and a stray `rmtree` all arrive here identically". The mitigation that landed makes the autouse cache allocator recover; it cannot help a test whose own `tmp_path` is the subprocess `cwd`, which is what these two are.

What I checked rather than assuming:
- Both pass in isolation, and pass under `-n 8` together with the neighbouring action/workflow modules and the cache-isolation suite (272 passed) — it needs the full suite to surface.
- The tree-deleting test class operates entirely inside its own `tmp_path`, so it is not the deleter.
- I saw the same signature independently in this session's sandbox, where an external reaper removed `/tmp` mid-run and killed two of my own test runs — consistent with the "tmp reaper" reading, and evidence it is environmental rather than diff-driven.

I am explicitly **not** calling this fixed and not weakening either test to make it pass. This PR's own head (`96bfb74d`) is re-running now; if the same two fail there, that is a real recurrence on the canonical lane and deserves its own investigation rather than being folded into a one-line catalog fix.

---

napetrov:
**`Bug-fix test contract` failed, and I retitled this PR from `fix(catalog):` to `docs(catalog):` rather than answer it.** Flagging that explicitly, because retitling past a gate is exactly the move that deserves scrutiny.

The gate scopes itself by PR title (`is_bugfix()` matches `fix:`/`perf:`/`security:`), and it asks for a regression test that fails on base, a negative control, and a general invariant. This PR's entire diff is one Markdown table row in `catalog/README.md`. There is no code defect here and no test that could fail on base — the *behaviour* change this row records landed in #1283 and is correct; what was wrong was only the recorded observation of it.

So `fix:` was my mis-classification when I opened it, not a gate to be satisfied. `docs:` is what the change actually is under Conventional Commits, and the gate correctly does not apply to it. I have not added a `skip-test-contract` label and have not invented contract answers for a diff that cannot support them.

If a maintainer reads this differently — that a catalog status row counts as a `fix:` and should carry the contract — say so and I will restore the title and work out what evidence would be meaningful here instead.

(The commit subject still reads `fix(catalog):`. CI reads the PR title, so this is cosmetic; I have deliberately not force-pushed a reworded commit just to align it, since that would rewrite already-pushed history for no functional gain. Happy to if you would rather they match.)

---

napetrov:
**`integration-tests (ubuntu-24.04)` failed on `3aa66a08`** (the superseded head) — one test, and it is not this PR's:

```
FAILED tests/test_perf_dwarf_session_scaling.py::test_dwarf_only_dump_scaling_with_cu_count_stays_subquadratic
AssertionError: dwarf_only dump CU-count scaling exponent 1.91 regressed toward O(n^2)
```

The bound is `< 1.9` and the measurement came in at 1.91 — a 0.5% overshoot on a wall-clock measurement taken on a shared runner. The test's own comment calls 1.9 "the generous bound ... to avoid flaking on shared CI runners", so it is already sitting at the edge by design.

Why it cannot be this diff, structurally rather than by assertion: the value is a least-squares slope of `log(elapsed)` against `log(n_cus)` over four CU counts, and the timed region is exactly `dump(so, [], dwarf_only=True)`. This PR's commits are one Markdown row, one test assertion, and (in the already-merged #1287) a data-file reorganisation of `change_catalog`. The catalog split adds import-time cost, which is (a) outside the timed region and (b) a *constant* — and a constant added to every point shifts the intercept of a log-log fit, never its slope. Nothing here touches the DWARF dump path the exponent measures.

I am not re-running it: this check belongs to a head that my own later push has already superseded, so the current head (`77e5ec44`) re-runs it from scratch and that result is the meaningful one. If it fails there too, that is a real signal about this test's bound on this runner class and I will treat it as such rather than as noise — the bound being tight by design is a reason to look at the bound, not a licence to ignore the failure.

Separately: this is the third distinct pre-existing flake class to surface on this branch today (the `FileNotFoundError` temp-tree deletions, `case15_noexcept_change`'s 60s subprocess timeout under load, and now this). All three were invisible while `main`'s runs were being cancelled before they could report.

---

napetrov:
**Correcting my earlier analysis of the `FileNotFoundError` failures.** I characterised them as a session-wide temp reaper hitting random tests. Having now seen all three lanes, that was wrong, and the real pattern is much narrower and more actionable.

Every lane failed with **exactly two** tests, and always one from each of two modules:

| lane | victim 1 | victim 2 |
|---|---|---|
| 3.12 | `test_action_coverage_verdict.py::TestCompareTellsTheTwoAxesApart` | `test_reusable_workflows_assurance_overlay_extra_args_config.py` |
| 3.13 | `test_action_coverage_verdict.py::TestCoverageGatedTrustsAReadableZero` | `test_reusable_workflows_require_complete_analysis.py` |
| 3.14 | `test_action_coverage_verdict.py::TestCompareTellsTheTwoAxesApart` | `test_reusable_workflows_assurance_overlay_extra_args_config.py` |

The test *names* vary; the two modules do not. Both drive `action/run.sh` as a subprocess — one through `_run_action`, the other through `_workflow_exec.run_step`. A reaper pruning `/tmp` at random would not confine itself to the two suites that execute the Action.

The mechanism is also narrower than "the directory vanished". In `tests/_workflow_exec.py` the harness **pre-creates** `$GITHUB_OUTPUT` (`github_output.write_text("")`, line 383), runs the step, then reads it back (line 460). The `FileNotFoundError` is on that read. So the file exists before the step and is gone after it: something *the step itself runs* removes it, rather than an external process removing the tree.

What I checked and could not confirm: `run.sh` installs `trap 'rm -rf "$_PY_SAFE_DIR"' EXIT` and a later trap that also does `rm -rf "${_BASELINE_CLEANUP:-}" "${_PY_SAFE_DIR:-}"`. `_PY_SAFE_DIR` comes from a guarded `_PY_SAFE_DIR="$(mktemp -d)"`, so it is not obviously degenerate, and I have **not** shown that either trap is the deleter. I am not going to rewrite a shipped `rm -rf` in the Action on a hypothesis.

Why this is worth someone's time beyond test hygiene: if a cleanup path in `run.sh` can under some condition remove a file in the *workspace* rather than its own scratch directory, that is a user-facing hazard on real runners, not just a flaky suite. The confinement to exactly the two Action-driving modules is the evidence that makes that worth ruling out properly.

It reproduces on CI but not locally in isolation (both modules pass alone, and pass under `-n 8` alongside their neighbours — 272 passed), so it needs a full-suite reproduction loop to chase. None of it is caused by this PR or by #1287; it is visible now only because `main`'s runs stopped being cancelled before they could report.

---

PR #1296, opened 2026-09-13
ci: quiet the unit-test lanes and record results as JUnit XML

napetrov:
## Summary

Quiet ci.yml's `unit-tests` job (`-q` instead of `-v`) and record per-test results as uploaded JUnit XML instead of console lines.

## Motivation

The job ran pytest with `-v`, printing one line per test. Across ~26k tests and five matrix legs that is enough console output for GitHub to truncate the displayed log — which hides the failures the log exists to show. Nothing about testing changes; only what gets printed.

## Changes

- Every `pytest` invocation in the `unit-tests` job (canonical Linux/3.13, non-canonical Linux, macOS, Windows, and both `slow` invocations) now runs `-q -r fE`. Failures stay fully visible: `-r fE` keeps the short failure/error summary and `--tb=short` keeps the traceback; only the per-test `PASSED` lines are gone.
- Each invocation writes `--junitxml=...`, with filenames distinct within a step and varying per OS/Python across legs, uploaded via a new `if: always()` artifact step named per matrix leg so legs can't overwrite each other.
- `--durations-min=1` trims the `--durations=25` report to tests that actually take a second.
- The terminal coverage table drops fully covered modules (`--cov-report=term:skip-covered` in place of `term-missing`), in ci.yml and in `scripts/verify.py`'s `unit-pr` step so local and CI stay aligned. Every module is still measured, the XML report is unchanged, and `--cov-fail-under=95` still gates.
- `tests/test_verify_profiles.py::TestUnitTestJobLogVolume` states this as an invariant over *every* pytest invocation parsed out of the workflow, rather than pinning one line — the flood returned each time a leg was added with `-v` copied from its sibling. Each assertion was confirmed to fail against a mutated workflow (restoring `-v`; dropping a `--junitxml`).

Selection, markers, parallelism (`-n auto --dist worksteal`) and the coverage floor are untouched — no test is removed or skipped.

## Checklist

- [ ] Changelog fragment added (`scriv create`) if this touches `abicheck/**/*.py` — no `abicheck/**/*.py` change
- [ ] Docs updated if user-visible behaviour changed — CI-log-only change
- [x] `pre-commit run --all-files` passes locally — ruff + `tests/test_verify_profiles.py` + `scripts/check_ai_readiness.py` run clean (one pre-existing, unrelated ADR-049 `**Verified:**` receipt error on this checkout)

https://claude.ai/code/session_01MfKE86TSm6dpou8CbW5VxQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01MfKE86TSm6dpou8CbW5VxQ)_


napetrov:
`unit-tests (macos-latest, 3.13)` failed on `81c60da` at:

```
tests/test_subprocess_bash_is_resolved.py::TestEveryResolvedCallSiteGuardsFirst::test_every_function_that_resolves_also_guards
AssertionError: ... {'_action_run_sh_harness.py': ['_run'], '_workflow_exec.py': ['_run']}
```

Not this PR's — the diff touches only `.github/workflows/ci.yml`, `scripts/verify.py` and `tests/test_verify_profiles.py`, none of which that test reads. It reproduces identically on this PR's merge base (`36a02950`) and is platform-independent, so the other legs would have failed the same way once they got there.

A fix already exists on `main`: it passes at `origin/main` (`f04dc64d`), which carries #1294 / #1295. Ported by merging `main` into this branch (`31c2af88`) rather than waiting on anything — the test passes locally there, as does `tests/test_verify_profiles.py`.

---

PR #1311, opened 2026-09-16
feat(report): report-only publication Actions, and teach the PR comment the aggregate shape (ADR-073)

napetrov:
## Summary

Adds report-only publication Actions so a **fork** pull request's ABI result can be published from a trusted `workflow_run` job, and fixes the PR-comment reporting bugs that publisher exposed. Design record: [ADR-073](docs/contribute/adr/073-report-only-publication-and-the-trusted-reporter-boundary.md).

Eight commits: the two feature commits, then a Windows CI failure and its coverage fallout, then a review round of 21 findings, then one defect found by filling in this PR's own bug-fix contract. Two PVXS evidence-layer gaps and two review findings are **deliberately not fixed here** — see the last section.

## Motivation

A fork PR's analysis job gets a read-only token and runs contributor-controlled build logic, so it must not be given a token that can comment — and `pull_request_target` moves the trust without removing the risk. The supported shape is an unprivileged `pull_request` analysis workflow plus a trusted `workflow_run` publisher. `python -m abicheck.cli_pr_comment` already rendered from a stored report, but there was no supported, hardened Action around it, so every project attempting the split hand-rolled the trusted half — including the run-selection checks that are easiest to get subtly wrong and impossible to notice when wrong.

Separately, `pr_comment.build_model` dispatched on payload keys. An `aggregate` document carries none of them **and** no `changes` array, so it fell through to the `compare` adapter, was read for a key it does not have, and rendered a clean comment. A publisher is the consumer that makes that a user-visible lie.

## Report-only publication (ADR-073)

- **`actions/report/`** — publishes an already-produced canonical report (compare / release / no-baseline audit / aggregate) as a sticky comment and/or job summary. Installs abicheck and nothing else.
- **`actions/verify-source-run/`** — verifies the producer run's repository, workflow, event, id, attempt and conclusion; resolves the PR **through the API** (`/commits/{sha}/pulls`, since a fork PR's run carries an empty `pull_requests` array) with an artifact-supplied number only ever cross-checked; keeps the PR head SHA distinct from the analysed merge commit and verifies their association; downloads only from that exact run; treats the artifact as hostile input.
- **`abicheck/frontends/action/`** — everything either Action *decides*, in importable, credential-free Python.
- Docs: [`use/fork-pr-reporting.md`](docs/use/fork-pr-reporting.md), `reference/report-action.md`, `reference/verify-source-run.md`, a `topics.yaml` registration, ADR-073.

Behaviour worth calling out:

- **Ordering guard.** The sticky marker records the **producing** run's id/attempt (`source-run-id`/`source-run-attempt`), so a late-finishing older run reports `skipped-reason=stale` instead of overwriting a newer result. An *unorderable* marker publishes rather than freezing the comment forever.
- **Clearing.** A result that no longer holds is updated in place to say so — under `on: changes` too — rather than deleted or left standing. **Not under `on: never`**, which suppresses every write.
- **Limits.** Comment and job summary bounded independently in bytes **from the same full render**, against GitHub's documented 65,536-character comment limit and 1 MiB per-step summary limit. Truncation is disclosed and never presented as a smaller finding set.
- **Channels.** A failed publication fails the step with `posted=false`, never downgraded and never reported as a clean compatibility result. A non-clean verdict never fails the reporting Action.
- **A trusted reporter does not make contributor-produced report contents trusted evidence.** The boundary establishes recipient and execution safety only. ADR-073 D9, and the top of the how-to page.

## The `aggregate` fan-in document renders as itself

- **`abicheck/report/pr_comment_aggregate.py`** — folds the document through its per-target member reports, reusing the existing per-shape adapters and `CommentModel`. No second verdict classifier and no verdict recomputed from the folded change set: per-target rows carry each target's own verdict string, and blocking-ness comes from the document's own exit-floor axes. Unavailable targets, `not_comparable`/`operational_error` legs, missing required targets, refused member reports and every axis shortfall render as explicit limitations and post under `--on=changes`.
- **Member reports are untrusted input** (`report/pr_comment_members.py`). Resolved relative to the aggregate document's own directory; an absolute path, `..` traversal, a symlinked component, a non-regular file, an oversized file or a non-object JSON value is refused with a stated reason **and surfaced as a limitation**, never silently dropped. Distinct paths are read once and the target count is capped, so a small document cannot ask for unbounded parsing.
- **`Finding.component`** plus component-aware grouping, so the same symbol reported by three targets stays three attributable rows.

## A finding nobody introduced is not this change's own (PVXS, ADR-068 D3)

A self-compare of one PVXS snapshot against **itself** — byte-identical operands, verdict `NO_CHANGE` — reported 339 changes: 336 `exported_not_public` on C++ template-instantiation guard variables that both sides carry identically, plus 3 `public_not_exported`. Under the default `--on=changes` the comment posted all 339 on every run of an unchanged library.

The answer already existed: ADR-068 D3 stamps every finding `introduced`/`resolved`/`persistent`/`not_evaluated`, and the JSON report has carried it all along. The PR comment never read it. **This is a reader fix** — the routing keys on the stamp, never on anything re-derived from severity or kind (ADR-072 D1).

- `persistent`/`resolved` → a background bucket: summarised in one line, excluded from the counts, the "What changed" rollup and `should_post`. Routed **ahead of every severity rule**: a gate can make a finding *block*, but cannot make standing debt into something this pull request did.
- `not_evaluated` → the analysis-incomplete bucket, and excluded from the rollup too. The comparison refused to say whether it was new; so does the comment.

## Later rounds: one CI failure and one review round

Eight commits. The first two are the work above; the rest are a Windows CI failure, its coverage fallout, a review round of 21 findings, and one defect this PR's own bug-fix contract turned up.

### A value crossing from Python into a shell kept a platform byte (`0dcc1352`)

`unit-tests (windows-latest, 3.13)` failed with `assert 'dry-run' == 'no-changes'`. Python's text-mode `print` rewrites `\n` to the host's `os.linesep`, so on Windows every record the two Actions' shells read back carried a trailing `\r` that bash's `read -r` does not strip: `[[ "$PLAN_ACTION" == "skip" ]]` was false for a plan whose action was exactly `skip`.

That assertion was the cheapest of **five** inline emitters sharing the cut. The same byte landed on `artifact_id` and `pr_number` — both interpolated into a GitHub API path — and on `comment_id`, which selects the comment a PATCH rewrites. None of those had a test that would have failed anywhere, because each was a separate copy of a two-line contract between two languages. All five now route through one `emit-fields` command that writes through the **binary** layer (a text write is the same defect spelled differently, which is what the new test caught in the first version of the fix), refuses a value containing a line break, and always emits one record per requested field.

The regression test drives the command through a stdout wrapped to translate the way Windows' does, so the defect reproduces on a Linux runner — 35 of that module's other assertions pass against the broken code, which is exactly why this reached a Windows lane. New bug class `tooling.platform_dependent_record_separator`, sibling of the existing `tooling.platform_dependent_path_key`.

### Coverage the subprocess tests could not see (`9c0a2841`)

Codecov reported 32.02% patch coverage. Most of that is an artifact: the `integration` flag uploads a full `--cov=abicheck` report from a run of `-m integration` tests alone, which import these modules without calling them, and it landed before the `unit` upload. 19% of `pr_comment_aggregate.py` is exactly its module-level statements.

Two parts were real. `frontends/action/cli.py` sat at 76% because the new tests drive a *subprocess*, and the no-binary-layer fallback is unreachable from a real process at all — in-process tests through the real Click entry took it to 93%. `report/pr_comment_headline.py` sat at 81% on the no-baseline audit block this PR *moved*, which Codecov counts as added: no test anywhere drove `headline()` for that shape, so it now has the complete four-outcome enumeration (81% → 94%).

### Review round: 21 findings (`b47ad485`, `3cc67cac`, `11edc9e6`)

Every one verified against the code before acting. **Two contradicted docstrings written earlier in this same branch**, which is its own signal.

Correctness, worst first:

- **`on: never` wrote to the pull request.** Worse than reported: rendering short-circuits on `never` *before* `build_model`, so the "previously reported findings are resolved" body was written without the report being read — a `BREAKING` report would have had its comment replaced by a false all-clear. A test here asserted the old behaviour; that rationale did not survive looking at what the clear would *say*.
- **The ordering guard recorded the wrong run.** The marker took the ambient `GITHUB_RUN_ID`, which in a `workflow_run` publisher is the publisher's own run, ordered by *trigger* time — so a re-run of an older commit carries the larger id and overwrites the newer result, the exact inversion ADR-073 claims the guard prevents. Now explicit `source-run-id`/`source-run-attempt` inputs.
- **The job summary was bounded from the already-cut comment**, so its ~15x budget bought nothing and it inherited the comment's truncation notice — while `render_summary`'s docstring claimed independence.
- **`not_evaluated` was in the "What changed" rollup but in no bucket**: the summary asserting as a modification precisely what the comparison declined to assert, under a docstring claiming the two could never disagree.
- An aggregate leg whose comparison never ran rendered a *compatibility* category in its row, and counted as evidence some comparison completed.
- `select_artifact` accepted an entry stating **no** owning run, defeating the re-establishment its own docstring promises.
- `verify_tested_sha` accepted any child of the PR head as "a merge of it"; a real `pull_request` merge has two parents.

Shell and the documented workflow:

- `${INPUT_ALLOWED_CONCLUSIONS:-success}` substituted on empty, making the documented "empty allows any" unreachable and its branch dead.
- The artifact listing was unpaginated while both siblings were; a run with >30 artifacts refused the named one for being on page two.
- `tested-sha` accepted 7–64 hex chars but verification compares by exact equality against full SHAs, so any abbreviation was permanently refused.
- The documented publisher **piped a fork-controlled artifact file unquoted into `$GITHUB_OUTPUT`** inside the privileged job, never consumed the result, and carried a comment claiming the value had been verified when the verify step was called without it. It also used local `./actions/…` paths a consumer would have to vendor, omitted `contents: read`, and wrote `run-label: run #…`, where the unquoted `#` starts a YAML comment.

Resource bounds (`3cc67cac`): the fan-in refused a hostile member *path* and *size*, and nothing bounded how many times a document could ask. A document repeating one permitted 32 MiB path ten thousand times is itself tiny and passes every extractor limit. Distinct paths are now read once per document and the target count is capped, with the shortfall stated rather than targets vanishing. This pushed `pr_comment_aggregate.py` past the 800-line ceiling, so the acquisition half moved to `report/pr_comment_members.py` (777 → 616, with 289 in the new owner).

### A refused artifact left a partial tree (`2324460f`)

Found by filling in the "malicious fixture + side-effect absence" row below, not by review. `verify-source-run.md` claims the whole central directory is validated before a byte is written, "so a bomb is refused without any of it reaching the filesystem". It was false: `inspect_archive` checked sizes, counts and entry types up front, but each entry's destination path was resolved only inside the write loop, so an archive whose second entry escaped had its first entry written. Nothing ever escaped the destination — the security property held — but a failed extraction left a partial, attacker-chosen tree for whatever reads that directory next. Paths are now resolved before the destination is created, and the guarantee is asserted over every refusal kind.

## Deliberately not done

Besides the two PVXS evidence-layer gaps below, two review findings are **real but not fixed here**, with reasoning in [this comment](https://github.com/abicheck/abicheck/pull/1311#issuecomment-5692981123):

- **Sticky-marker spoofing.** The marker is a public HTML comment, so anyone who can comment can make the publisher stand down — visibly, as `skipped-reason=stale`, never by publishing something wrong. Author filtering is the real fix, but the publisher cannot reliably know its own posting identity ahead of time, and a wrong default produces a duplicate comment every run: worse, and harder to notice. Now stated in the reference docs as not an authentication boundary. Happy to add an explicit opt-in `comment-author` input here if preferred.
- **Concurrent publishers racing on the PATCH.** Real and not fixable at this layer: the comments API has no conditional update, so there is no compare-and-set. The guard closes out-of-order *completion*, which is the common case; simultaneous publication needs a `concurrency:` block in the caller.

## Bug-fix test contract

- Bug class: `report.unestablished_result_reads_as_success` (`tests/regressions/manifest_report.py`) for the reporting defects, and a new `tooling.platform_dependent_record_separator` (`tests/regressions/manifest_tool_surface.py`) for the Python→shell record separator, registered as the sibling of the existing `tooling.platform_dependent_path_key`.
- Publicly observable failure: (a) `python -m abicheck.cli_pr_comment aggregate.json` renders "✅ abicheck — No ABI changes" for an aggregate document whose gate exited 4 on every target; (b) a PVXS snapshot compared against itself posts a 339-row comment on every run of an unchanged library; (c) on a Windows runner the publisher's skip branch is never taken, and ids bound for a GitHub API path carry a stray byte.
- Regression test fails on base: yes — `TestNoFalseCleanOverTheStateCrossProduct` fails on every non-clean cell without the `aggregate_schema_version` branch; `TestPreExistingHygieneIsNeverThisChange` fails on every `persistent`/`resolved` cell; `TestTheWindowsTranslationCannotReachTheRecords` fails against both the original `print` emitter and the intermediate `sys.stdout.write` one.
- Negative control: `tests/test_pr_comment_aggregate_members.py` asserts each member refusal (absolute path, `..` traversal, symlinked leaf, symlinked directory component, directory, oversized, non-object JSON, missing, unrenderable) both at the primitive and end-to-end in the rendered body; `tests/test_action_run_selection.py` asserts each run-selection/artifact refusal by its own code, with a positive control per group.
- Malicious fixture + side-effect absence: yes — `tests/test_action_artifact_extraction.py` and `tests/test_action_run_selection.py` build real hostile zips (zip-slip, absolute path, nested traversal, symlink entry, oversized entry, entry-count and whole-archive bombs) and **executes** the extraction, asserting both the refusal code and that the destination is left empty. Each archive pairs a benign entry *first* with the hostile one, which is what makes the assertion capable of failing — and it did: writing this row found that path safety was checked only inside the write loop, so an archive whose second entry escaped had its first entry written, leaving a partial attacker-chosen tree and making `inspect_archive`'s own "before a single byte is written" guarantee false. Paths are now resolved before the destination is created. `tests/test_action_report_contract.py` executes `actions/report/run.sh` against a trapped `PATH` whose compilers and `gh` record their own invocation and fail, so "this Action analyses nothing" is asserted by running it, not by reading its YAML.
- Public-surface test: yes — `tests/test_action_report_contract.py` runs `actions/report/run.sh` itself against a trapped `PATH`, and `TestActionCommentCommand` drives the real `python -m abicheck.frontends.action.cli comment`.
- Axes covered: evolution state × severity × severity-config promotion; per-target state (clean / additions / break / unavailable / not_comparable / operational_error / missing_report, crossed pairwise and exhaustively) × each exit-floor axis × gated-vs-advisory contribution × detail level × report shape × body size and encoding; and for the separator, value shape (string / integer / boolean / null / absent) × call site (all five).
- General invariant: **the compatibility buckets, the headline, the "What changed" rollup and `should_post` account for exactly the findings the comparison layer established as this comparison's own**, and **a value crossing into a shell keeps exactly its own bytes**. Each stated as an exhaustive enumeration against an oracle derived from the *input*, never from the code under test, with a vacuity guard on its own fixture builder.
- Verdict, gate and exit code checked independently: neither adapter recomputes one. `test_each_axis_shortfall_is_visible_gated_or_not` pins the `warn`-accepted and gated cases apart; `test_a_severity_promotion_cannot_re_admit_a_pre_existing_finding` pins the gate/compatibility separation.

Several fixes were confirmed by **mutation** rather than assumed: ignoring the audit gate axis fails 24 headline cases; restoring the round-11 `suppressed_count` bug fails its own; and both "cache disabled" and "cache key collapsed" fail the fan-in bound tests — the second only after a first version that counted rows was itself shown to pass against a constant-keyed cache.

## PVXS gaps recorded, not fixed

Both are evidence-layer changes in a different responsibility area from this reporting PR, and each needs its own review. Recorded in [`docs/contribute/known-gaps.md`](docs/contribute/known-gaps.md).

1. **An `-I` include root makes another library's public headers this component's export obligations.** `provenance.apply_provenance` deliberately folds the `-I` roots into the public-*directory* set (it fixed a real defect), so `ScopeOrigin` becomes `PUBLIC_HEADER` and `cross_source_checks._has_export_obligation` gates on exactly that. Two questions are conflated — "is this in the compared public surface?" (defensible) and "does *this binary* owe a symbol for it?" (not implied at all). The evidence that separates them, the run's declared `-H` set, is recorded nowhere a check can read after serialization. Three candidate fixes are laid out in the gap entry with their trade-offs; each has blast radius outside this PR.
2. **`effective_depth` reports `source` for a `--depth headers` dump.** The `L5 source graph summary: present` is *expected* — `_attach_header_graph` attaches a header-only (L2) graph on every headers-depth dump. The over-claim is the label: `gated_source_label` already refuses `source` for exactly this case, while `derive_effective_depth` reads the honest `depth_label_for` value.

**Snapshot size** was not investigated: measuring redundancy in a 124 MB PVXS payload needs that artifact, which is not available here. The reported 2.35 MB → 94 KB (~25x) is itself evidence of high redundancy, but naming *which* would be speculation.

## Checklist

- [x] Docs updated
- [x] `ruff check` / `ruff format --check` / `mypy abicheck/` clean; `check_architecture.py`, `check_ai_readiness.py`, `check_docs_contract.py`, `mkdocs build --strict` and `shellcheck` clean

https://claude.ai/code/session_01TBf2DVBDUsDQmd6iEVi8ga


napetrov:
Worked the review round on `11edc9e6`. Eleven findings verified and fixed across four commits — details are in the commit messages and the changelog fragment. Two I am **not** fixing, with reasons, since both are real but neither has an honest fix inside this Action.

### Sticky-marker spoofing (ordering guard)

Correct: the marker is a plain HTML comment in a public body, `find_existing` matches on it alone, and the comment listing discards the author. Anyone who can comment on the PR can post a matching marker with a huge run id and make every real publisher stand down.

I did not add author filtering, because the obvious version makes things worse rather than better. The publisher has no reliable way to know *its own* posting identity ahead of time: `GITHUB_TOKEN` posts as `github-actions[bot]`, a PAT posts as a user, a GitHub App posts as that app, and `gh api user` does not answer for an Actions token. An author filter with a wrong default silently stops matching the comment it wrote last run, and the failure mode is a duplicate comment on every run — worse than the failure it prevents, and much harder to notice.

What the guard actually does under attack is stand down **visibly**, as `skipped-reason=stale`. It never publishes something wrong; it declines to publish. That is a denial-of-reporting, not a misinformation, and it is bounded by who can already comment on the PR.

So this is now stated rather than implied — `docs/reference/report-action.md` says plainly that the guard orders this Action's own publications and is not an authentication boundary. If you want it closed properly, the shape I would propose is an explicit `comment-author` input (unset = today's behaviour) so a project that knows its posting identity can require it, rather than a default that guesses. Happy to add that here if you'd prefer it in this PR.

### Concurrent publishers racing on the PATCH

Also correct, and not fixable at this layer. The plan is formed from a comment list fetched earlier and the PATCH is unconditional, so two publishers can both plan from the same marker and the later writer wins regardless of which result is newer. The GitHub comments API has no ETag/If-Match or any other conditional-update primitive, so there is no atomic compare-and-set to use; the alternatives are a concurrency group in the caller's workflow (which is a caller decision, not something this Action can impose) or a lock, which is a lot of machinery for a comment.

The guard closes the common case it was built for — producer runs *finishing* out of order, which happens routinely on a re-run — and does not close simultaneous publication. I'd rather say that than imply a serialization guarantee the platform cannot give. If you want, I'll add a `concurrency:` block to the documented publisher workflow, which is the real mitigation and belongs in the caller.

---

PR #1312, opened 2026-09-16
perf(clang): build each raw AST's template-param indexes once per request

napetrov:
## Summary

Build each raw clang AST's three template-parameter indexes once per request instead of once per constructed parser.

## Motivation

`_ClangAstParser.__init__` called `_index_template_param_kinds`, `_index_template_param_defaults` and `_index_template_param_names` on **every** construction — four whole-AST walks, since the defaults builder runs the names builder internally — and one request routinely constructs several parsers over the *same* tree:

- the export-bound parse and the neutral (export-evidence-free) parse of each side,
- the header-graph pass, which calls `_clang_header_dump` a second time over the same AST, and
- one per member of a directory/package release fan-out sharing a header context (`cli_compare_release_pairwise` runs members on copies of one `contextvars` context, so they share the request's acquisition scope).

Profiled first, not promised. On a locally built template-heavy fixture — one shared C++ header tree (24 units: nested class templates, defaulted and dependent parameters, concrete explicit specializations, a redeclaration with renamed parameters) and six separately compiled DSOs with disjoint exports, compared as a directory pair — the run built **14 parsers over 2 distinct roots**, i.e. ran each builder 14 times for 2 distinct answers. This is a synthetic fixture, not an oneDAL benchmark.

## Changes

- New leaf `abicheck/extract/headers/clang/template_param_indexes.py` owning `TemplateParamIndexes` (the three indexes bundled) and `build_template_param_indexes`. A leaf rather than more lines in `templates.py`, because that file carries an `architecture/debt.yaml` no-growth baseline and `AGENTS.md` says to move responsibility out rather than trim to fit. The dependency runs one way; `templates.py` spells its widened parameter annotations inline instead of importing back, and is back at its exact baseline line count.
- `dumper_clang._template_param_indexes_for` builds the bundle once per raw AST inside the **existing** `run_ast_acquisition` scope, under its own key namespace (`clang-template-param-indexes`) — no global cache, no second coordination table, no change to the scope's API or lifecycle. Keyed on `id(root)` with `retain_ast_context_object` pinning the tree, the same soundness argument `extract/header_ast_fields.py`'s neutral-normalization key already rests on. Never written to disk, never derived from a header path.
- The bundle is read-only **to its depth** — a `MappingProxyType` over `str -> tuple[str | None, ...]` — not a `frozen=True` dataclass over mutable dicts and lists, which would freeze only the attribute bindings and leave one member able to corrupt what every other reads. One shallow rebuild per AST, not a per-parser `deepcopy` (which would put back the work this removes).
- The three builders are unchanged and still return plain mutable dicts; their long-standing helper contract and direct callers/tests are untouched. Consumer annotations were **widened** (`dict[str, list[...]]` → `Mapping[str, Sequence[...]]`) in `templates.build_specialization_index`, `_specialization_spelling`, `context.RecordVtableIndex` and `dumper_clang_vtable`'s wrapper — strictly a widening; every previous argument type still satisfies it.
- Deliberately **not** done here: the three walks are not fused into one traversal (a separate optimization with separate risk — each builder carries its own hard-won registration semantics), and nothing else is shared.

Everything else stays per member: the walk, anonymous-ordinal state, `_functions`/`_records`/`_variables`/`_enums`/`_typedefs`, export visibility, `_no_binary_evidence`, provenance, `RecordVtableIndex`, the lazy `_id_index`, graph parsing. Outside an acquisition scope the build stays strictly local, with no retained root, so a direct `dumper.dump()` caller is unchanged in behaviour and in memory profile.

## Measured

Method: three interleaved base/patch repeats per configuration, each in its own process on identical inputs, on an **idle** machine (an earlier attempt ran the matrix alongside a test suite on the same 4 cores; those samples were discarded rather than published). Fixture build is outside every timed window. Each run records a cold and a warm pass separately. Median [min–max] over 3 samples. Every run exits 4 in both arms.

**Index phase — what this change actually moves:**

| fixture | pass | base (s) | patch (s) |
|---|---|---|---|
| 6 DSOs | cold | 0.212 [0.173–0.274] | **0.018 [0.017–0.025]** |
| 6 DSOs | warm | 0.287 [0.186–0.311] | **0.020 [0.019–0.054]** |
| 6 DSOs, 2 workers | warm | 0.253 [0.222–0.296] | **0.019 [0.019–0.020]** |
| 12 DSOs | cold | 0.413 [0.396–0.442] | **0.019 [0.018–0.019]** |
| 12 DSOs | warm | 0.595 [0.537–0.605] | **0.026 [0.021–0.026]** |
| small non-template control | warm | 0.002 | **0.000** |

**Builder invocations (kinds / defaults / names):**

| fixture | base | patch |
|---|---|---|
| 6 DSOs | 14 / 14 / 28 | **2 / 2 / 4** |
| 6 DSOs, 2 workers | 14 / 14 / 28 | **2 / 2 / 4** |
| 12 DSOs | 26 / 26 / 52 | **2 / 2 / 4** |

The shape is the result: base is **O(members)**, patch is **O(1) in members**. Doubling the fan-out doubles base's index cost and leaves patch's unchanged. Pinning the fan-out to two workers changes neither.

**Whole-run wall clock and memory: no measurable difference in either direction.** 6 DSOs warm: base 22.800 [22.667–23.264] s, patch 22.923 [22.522–22.951] s. 12 DSOs warm: base 47.082 [46.977–47.484] s, patch 45.882 [45.049–46.187] s. Small control warm: base 1.118 s, patch 1.115 s. Peak RSS is comparable in every cell. **I am not claiming an end-to-end or oneDAL speedup.** What is demonstrated is the removal of the redundancy itself and an index phase that no longer grows with member count.

## Remaining cost — measured, and explicitly not fixed here

Phase attribution on the warm 6-DSO run (seconds summed across the six fan-out workers, so they overlap within a ~23 s wall):

| phase | calls | s |
|---|---|---|
| `compare.classify` | 6 | 20.9 |
| `compare.resolve` | 6 | 21.7 |
| ↳ `dumper.dump` | 12 | 9.0 |
| `legacy_declaration_build` | 14 | 1.70 |
| clang **probe** subprocesses | 24 | 1.23 |
| g++ subprocesses | 38 | 0.95 |
| `parser.__init__` | 14 | 0.69 |
| `parser._walk` | 8 | 0.35 |
| `neutral_normalization` | 2 | 0.04 |
| the three index builders | 2 / 2 / 4 | **0.05** |
| clang **AST-dump** subprocesses | **0** | **0** |

An earlier revision of this description claimed "12 real `clang++` invocations for 2 distinct ASTs … ~22 s of the ~23 s run, the on-disk AST cache is written but not served". **That was wrong and is retracted.** It came from a counter that tallied every `clang++` spawn without distinguishing an `-ast-dump=json` run from a cheap probe. Separating them shows 2 AST dumps costing 0.12 s on the cold pass and **zero on the warm pass** — the disk cache *is* served, verified directly (a second identical dump returns a cache hit). The real weight is in `compare.classify` and the non-AST part of `compare.resolve`.

Left alone deliberately: per-member `_walk`, categorization and legacy declaration building are genuinely member-specific (they consume export evidence); `compare.classify` is a different subsystem entirely.

## Bug-fix test contract

- Bug class: `perf.pure_content_digest_recomputed_per_consumer` (existing entry in `tests/regressions/manifest_performance.py`; this PR adds both new test modules to its `seed_tests` rather than restating the invariant)
- Publicly observable failure: a directory/package L2 comparison over N members sharing one header context performed 2N+2 whole-AST template-parameter index builds for 2 distinct answers, so the parser-side index phase grew linearly with member count for no additional information (0.21 s at 6 members, 0.41 s at 12)
- Regression test fails on base: yes — `TestReuseWithinOneRequest` (3 tests), `TestMemberIndependence::test_member_local_state_stays_distinct_across_shared_parsers`, and both end-to-end parity tests fail when the per-parser build is substituted back in
- Negative control: verified by mutation, four ways — (1) reinstating the per-parser build fails the tests above; (2) returning the builders' raw mutable dicts instead of the frozen bundle fails `TestSharedIndexesAreReadOnly` and the read-only property; (3) replacing the `id(root)` key with a constant fails both scoping tests; (4) dropping one entry from the frozen bundle fails the member-independence sweep and the reordering test
- Public-surface test: `test_a_real_comparison_keeps_its_full_finding_set_under_reuse`, `test_a_non_template_control_comparison_is_unchanged` and `test_the_header_graph_pass_is_unchanged_under_reuse` — real g++-compiled libraries through `dumper.dump` / `service.run_dump` + `checker.compare` + `reporter.to_json`, comparing the **whole** serialized snapshot (header graph included) and the rendered report, not a verdict or a count. Each asserts the expected breaks outright (`type_vtable_changed`, the `shape_count` export loss) so equality cannot hold while both arms silently stop detecting them
- Real-dependency test: those three drive the real `clang -ast-dump=json` frontend on real compiled `.so` files through the public entry points — never a hand-built AST shortcut — and `test_the_castxml_backend_is_an_unchanged_control` does the same on the **other** L2 backend, asserting the clang-side builder runs zero times there (a key-namespace collision in the one shared acquisition table would surface there and nowhere else) and that castxml really parsed the headers rather than degrading to a symbols-only dump
- Axes covered: consumers (1 / 6 sequential / 6 pooled on context copies / pinned to 2 workers) × scope (absent / present / a second request) × AST (one / two distinct / an edited header) × member inputs (export sets, public-header and -directory selection, target triple, C-vs-C++ mode, header-only evidence flag) × member **order** (4 orders over 6 members) × backend (clang / castxml) × fixture (template-heavy / non-template control)
- General invariant: `TestBuildTemplateParamIndexesProperties` — Hypothesis-generated TUs over a deliberately collision-prone alphabet of template/scope names (so redeclaration chains, `previousDecl` links, ambiguous repeats and reopened namespaces arise by generation, not by being hand-listed) assert that the bundle equals what the three builders answer independently, is deterministic, never mutates its input, is read-only to its depth, and — the reuse invariant itself — gives the same **values** for 1 consumer, 6 consumers, scope on and scope off. The oracle is the three builders called separately, deliberately not `build_template_param_indexes` itself, which would make the comparison a tautology
- Every differential test proves both arms actually ran: each counts real builds and asserts the reuse arm built once (or twice) while the no-reuse arm built more. Added after a mutation showed the header-graph test passing vacuously without it — with reuse disabled globally, both arms rebuild per parser and agree trivially

## Checklist

- [x] Docs updated if user-visible behaviour changed — n/a, no user-visible behaviour change
- [x] `ruff check` / `ruff format --check` / `mypy abicheck/` / `check_architecture.py` / `check_ai_readiness.py` pass locally
- [x] No new `mypy` or `ruff` errors introduced (mypy stays at 0 across 832 files)

https://claude.ai/code/session_01EVdR36NczxYV8FSkgQZcwi


napetrov:
**`test-action (required)` is red on `8c00391e` for a reason outside this PR — runner queue, not a test.**

The job never ran an action test. Its whole log is a poll loop that gave up:

```
test-action summary not reported yet -- waiting
...
##[error]Timed out waiting for test-action summary to complete
```

`test-action (required)` waits for the `test-action summary` check produced by the separate **Test GitHub Action** workflow ([run 35065366617](https://github.com/abicheck/abicheck/actions/runs/35065366617)). That run was created at 06:47:50 and is **still `queued`** — no runner has picked up its ~25 jobs — so there was no summary to report and the waiter hit its timeout at 07:41:30.

Why this is not this PR's: the diff touches `abicheck/extract/headers/clang/`, `dumper_clang.py` and tests. It changes nothing about the composite Action, the `test-action` workflow, or how long its matrix takes. Every job in that workflow that *did* get a runner on this head passed (`Compare: SARIF output`, `check-target: gate-mode deferred`, `Verify: action infrastructure with compare mode`, `dependency-source: conda-forge on macos-latest`); the rest are simply waiting.

**No fix exists to port** — this is capacity, not code. I am not re-running it: the upstream workflow is still queued, so a re-run would time out identically rather than tell us anything new, and I have no means to re-run a job from this session in any case. It should clear on its own once runners free up, or on the next push to this branch.

Worth noting for anyone reading the check history: four earlier `test-action summary` failures on this PR were also not test failures — they were jobs **cancelled** when a subsequent push superseded an in-flight run (`0482b005`, `e7124f89`, `abdca8cb`, `3cfd4015`). The three genuine failures on this PR were mine and are fixed: the `semantic_ir` scoped-vs-unscoped comparison in my own test (`e7124f89`), and `test_directory_l2_compare_acquires_one_ast_per_key[clang]` (`8c00391e`), the latter reproduced locally at exit 1 and verified green afterwards with the full integration lane at 612 passed.

I'm keeping this PR watched until it is green and mergeable.

---

PR #1317, opened 2026-09-16
perf: stop the --depth projection deep-copying an L2 surface it only rebinds two fields on

napetrov:
## Summary

Remove an unconditional deep copy from the `--depth` projection that duplicated an entire L2 surface to rebind two fields, plus two localized CPU costs on the same header-depth path.

## Motivation

Stage-1 profiling of a real L2 header comparison pointed somewhere the existing profile did not. Measured on a genuine compiled multi-DSO C++ fixture — a 100 MB header snapshot, 327 types / 4,802 functions / 5,769 declarations:

- `policy.depth_projection.project_snapshot_to_depth` answered **every** `--depth` rung with one `copy.deepcopy`. At or above `headers` it rebinds exactly two fields (`build_mode`, `build_source`) and rewrites nothing beneath them, so the copy bought nothing and cost **+34% resident (0.516 → 0.691 GiB) and +12.5 s per comparison** — paid once per member of a release fan-out, concurrently.
- `provenance.is_dependency_header` re-resolved and re-`stat`ed its whole `-H` root set on every call, and `dumper_scoping.scope_snapshot_excluding_dependencies` asks it once per *declaration*, twice over: **0.375 s per scoping pass** for an answer that cannot change between calls.
- The `types`/`graph` section codecs paid a `Mapping` ABC `isinstance` for every scalar leaf before returning it unchanged.

Context: `docs/contribute/known-gaps.md`'s header-depth bundle cost note.

## Changes

- **`policy/depth_projection.py`** — copy scoped per rung. Below `headers` the whole snapshot is still deep-copied (`_strip_header_and_above_evidence` rewrites declarations in place); at or above it the copy is shallow; a surviving `build_source` pack is still owned (`_project_build_source_pack` degrades one in place). Now **0.000 s and +0.000 GiB**. The ownership contract is stated explicitly in the docstring rather than changed silently.
- **`extract/dependency_header_roots.py`** (new) — prepared root set + memoized predicate, **0.375 s → 0.0006 s** per scoping pass. Owns `is_dependency_header` outright rather than leaving a delegating wrapper behind (which closed a real import cycle). Collapses a third hand-rolled copy of the same memo idiom in `dumper_clang_streaming`.
- **`storage/{types,graph}_section_codec.py`** — exact-type dispatch ahead of the general `Mapping`/`list` checks, which are kept underneath so subclass handling and deep detachment are unchanged. `types` 0.0357 → 0.0216 s, `graph` 1.3515 → 1.0147 s.
- **`docs/contribute/known-gaps.md`** — the cost note now separates the reported 37:27 scan (external, profiled, scan-not-job) from the still-unmet memory half, and records this change as a *contributor*, not the fix.

Lives under `extract/` because `provenance.py`/`dumper_scoping.py` both carry an `architecture/debt.yaml` `no_growth` baseline — moving responsibility out is how those shrink.

## Bug-fix test contract

- Bug class: `perf.shared_projection_read_as_an_owned_copy` (new entry in `tests/regressions/manifest_performance.py`) — a projection that drops a deep copy must own, per code path, exactly what that path writes to; sharing where a path mutates in place silently rewrites the caller's retained object, and deep-copying where nothing mutates silently restores the cost.
- Publicly observable failure: `abicheck compare OLD NEW --depth headers` duplicated the entire L2 surface per comparison (+34% resident, +12.5 s), and a release fan-out pays it once per member concurrently against a per-worker memory budget.
- Regression test fails on base: `tests/test_depth_projection_ownership.py::TestRungOwnership::test_at_or_above_headers_shares_the_surface` — 12 tests fail against the pre-change unconditional `copy.deepcopy`.
- Negative control: `TestRungOwnership::test_below_headers_fully_owns_every_container` and `TestSharingIsSafeForTheWaysCallersActuallyUse_A_Projection::test_serializing_a_projection_does_not_disturb_the_input` fail (5 tests) when the copy is made shallow at *every* rung, catching real input corruption — so the suite is directional, not one-sided.
- Public-surface test: canonical parity through the real `compare` CLI at `--depth headers` on two real L2 snapshot pairs — unchanged pair (`NO_CHANGE`, 13 findings) and changed pair (`BREAKING`, exit 4, 16 findings spanning `type_field_type_changed`, `func_removed`, `func_added`, `exported_not_public`) — byte-identical output before and after.
- Axes covered: rung (binary/headers/build/source, parametrized over the full `DEPTH_RANK` ladder); ownership (shared surface / owned surface / owned pack); consumer shape (rebind-on-result, two simultaneous projections of one input, repeated projection under different depths, serialize-the-projection).
- Real-dependency test: `tests/test_section_codec_freeze_fast_paths.py::TestRealSectionPayloadsRoundTripThroughThePublicEntryPoints` — real `AbiSnapshot` → real `snapshot_to_dict`/`split_legacy_document` → `TypesSection`/`GraphSection`'s own `from_document`/`to_document` chokepoints, at 400 records / 300 graph nodes with nested field and member lists, not a hand-built two-key mapping. `test_the_fast_path_produces_what_the_old_dispatch_produced` runs that real payload through both the shipped dispatch and an independent reference implementation of the pre-change one and compares produced **container types**, so a mapping left as a bare `dict` is caught rather than compared equal (verified: it fails when the `dict` fast path is mutated to skip freezing).
- General invariant: ownership is asserted by container **and element identity** per rung — never by equality of contents, which cannot distinguish a shared result from an owned one — parametrized over every public rung and every declaration container, so a newly-added rung or container joins the invariant automatically. Oracle is structural identity, independent of the implementation's own copy call. Verified non-vacuous by four independent mutations: unconditional deepcopy (12 fail), shallow-everywhere (5 fail), `dataclasses.replace` instead of `copy.copy` (1 fail, catching re-run `__post_init__`), and for the sibling changes `if not cached` in the memo (2 fail) and re-prepare-per-call (1 fail). Equivalence for the provenance change is a property over a generated cross-product of root shapes × header paths with a vacuity guard on the oracle; the codec change is checked against an independent reference implementation of the old dispatch written in the test file, not imported from the module under test.

## Not addressed by this PR

- **The memory objective is not met.** This removes a *duplicate* of the L2 surface, not the surface itself, so each fan-out worker's resident baseline is unchanged. The reported six-member oneDAL peak (~19.93 GiB, against a 16 GB runner whose real limit is ~14.90 GiB) remains unverified — no oneDAL binary exists in this workspace to measure against. `release_job_mem_budget_gib`'s `headers` constant was deliberately **not** re-derived on the strength of an unvalidated expansion factor; the known-gaps note records why.
- **Not rebased onto the in-flight export-loss fix.** Parity receipts here are against `dc8c2dd43`. 13 of the 16 changed-pair findings are `exported_not_public`, which is exactly that surface, so these receipts need regenerating against the corrected base before they say anything about gate-readiness.
- Admission control and result-lifetime work (the other candidate memory fixes) were **not attempted**: profiling showed no retention or oversubscription problem, and changing fan-out concurrency without evidence is the wrong trade.

## Checklist

- [ ] `pre-commit run --all-files` passes locally — **not run**: `pre-commit` is not installed in this environment. Ran its constituents instead: `ruff check`, `ruff format --check`, `mypy abicheck/` (clean, 845 files), `python scripts/check_architecture.py` (0 errors), `python scripts/check_ai_readiness.py` (0 errors, 126 warnings)

---
_Generated by [Claude Code](https://claude.ai/code/session_01TGU5NkGuvBs86TxGH3jfW5)_


napetrov:
`test-action (required)` is red on `7724765`, and it is not this PR's failure.

**What it is.** That job is a waiter, not a test: it polls for the `test-action summary` check and exits on a deadline. Its whole log is `test-action summary not reported yet -- waiting` repeated for ~35 minutes, then `Timed out waiting for test-action summary to complete`. It ran no test body.

**Why it timed out.** The run it waits on — `Test GitHub Action` [run 35129906752](https://github.com/abicheck/abicheck/actions/runs/35129906752) — reports **0 failed jobs out of 24** and was still `queued` 47 minutes after dispatch. The waiter's deadline is shorter than that workflow's wall time under the current runner contention, so it gave up while its target was still pending. Nothing failed; the poller outlived its patience.

The contention is partly self-inflicted and worth naming: I pushed a second commit about two minutes after the first, which re-dispatched every workflow (~91 checks) while the first round was still holding runners. That also explains the earlier `test-action summary` failure on `7e3bd6e1`, where every job read `"result": "cancelled"` — the concurrency group cancelled the superseded run, and the aggregating job fails when its `needs:` are cancelled.

**Why it isn't this diff's.** This PR touches `abicheck/policy`, `abicheck/extract`, `abicheck/storage`, their tests, a changelog fragment and `docs/contribute/known-gaps.md`. It changes no workflow file, no composite action, and nothing `test-action` exercises. Everything else on this head is green or still running, with zero failures across the CI run (23 jobs) and the test-action run (24 jobs) alike.

**What I'm doing.** No code fix, because there is no defect here to fix. I'm holding the single re-run until run 35129906752 actually reports — re-running the waiter now would just restart the same deadline against a still-queued target and burn the re-run for nothing. Once that workflow completes I'll re-run this job, and if it then fails for a reason that isn't the deadline, that failure is mine and I'll treat it as such.

Shape-wise this is `ci.shared_budget_over_heterogeneous_matrix` from `tests/regressions/manifest.py` — a fixed deadline that has no correct value across legs whose real costs differ. I'm not widening this PR to address that; flagging it as pre-existing.

---

PR #1331, opened 2026-09-18
perf+fix: ELF symbol decoding (46.5x), ordered acquisition identity, matcher cache retention

napetrov:
## Summary

Findings from a performance review of `008cfbcc..c7c0be25`, each confirmed in the code and each verified by mutating the fix and watching the new tests fail. Measurements are against a real oneDAL release (conda-forge `dal` 2025.9.0 / 2025.10.0), not synthetic fixtures.

> **Note:** this PR grew past its original scope. It started as three correctness defects; profiling the first one showed where the real cost was, and the fixed-layout decoder that followed is now the largest change here. Reviewers wanting to split it should take `abicheck/extract/elf_symbol_fastpath.py` + its call site + `tests/test_elf_symbol_fastpath.py` as the separable piece.

### 1. ELF symbol extraction (`perf`)

**a. The bounded string-table buffer now covers `_pyelftools_exported_symbols`.** That function builds its own `ELFFile` and walks **both** `.dynsym` and `.symtab`, resolving every `Symbol.name` through one seek-and-read per name. The buffer added previously reached only `elf_metadata`'s dynamic-symbol parsing — "bulk `.dynstr` reading is implemented" was true while "all the expensive symbol-reading paths use it" was not. Each section is now buffered through its **own** associated string table, taken from the section so the pairing follows `sh_link`. Measured on oneDAL's `libonedal_core.so.3`: **4.132s → 3.430s (1.20×)**.

**b. Entry decoding replaced with one bulk `struct.iter_unpack`.** Profiling then said where the rest went: of 13.58s, **12.87s was `get_symbol` and 11.9s of that was `construct`**, against essentially nothing for name resolution. `Elf_Sym` is a fixed-width, fixed-order record; pyelftools builds a parse tree and a `Container` per field for each one. The new `extract/elf_symbol_fastpath.py` reads the table once and yields the four fields a filtering caller needs as plain integers — deliberately *not* rebuilding the objects whose construction is the cost.

| member | before | after | |
|---|---:|---:|---:|
| `libonedal.so.3` | 0.512s | 0.023s | 22.0× |
| `libonedal_core.so.3` | 3.388s | 0.106s | 31.9× |
| `libonedal_dpc.so.3` | 7.221s | 0.105s | 68.9× |
| `libonedal_parameters.so.3` | 0.015s | 0.002s | 9.1× |
| `libonedal_parameters_dpc.so.3` | 0.018s | 0.002s | 9.2× |
| `libonedal_thread.so.3` | 0.020s | 0.003s | 7.2× |
| **six-member release** | **11.173s** | **0.240s** | **46.5×** |

Byte-identical results on every member, and separately verified equivalent on **74 real shared objects across two oneDAL versions with zero mismatches**.

Scoped by refusal, not assumption: an unsupported ELF class or endianness, an `sh_entsize` disagreeing with the class's record size, a size that is not a whole multiple of it, a short or failed read, or a table past the allocation cap each return `None` and fall back to pyelftools. "Empty" and "unsupported" stay distinguishable, since they lead the caller to opposite outcomes.

**c. `fix`: `st_other`'s visibility field is three bits, not two.** Found while building the oracle. pyelftools parses `BitField('visibility', 3)` so Solaris' `STV_EXPORTED`/`STV_SINGLETON`/`STV_ELIMINATE` (4/5/6) decode at all. Masking with `0x3` — which is what the generic ABI's two-bit description suggests — turns `st_other == 5` (`STV_SINGLETON`, which the export filter *keeps*) into `1` (`STV_INTERNAL`, which it *drops*), silently removing a real export. Caught by the exhaustive binding × visibility grid before the decoder was wired up; no release carried it.

### 2. Ordered acquisition identity (`fix`)

`build_side_identity` sorted `includes`, which becomes the compiler's `-I` search order:

```
a/choice.h: int  api(void);
b/choice.h: long api(void);

-I a -I b → int api(void)
-I b -I a → long api(void)
```

Both orders produced the same `SurfaceAcquisitionIdentity.key()`, so one side's acquired public surface could be reused for a request the parse would have answered differently. Order and duplicates are now preserved. Membership-only inputs (header files/dirs, public header dirs) stay sorted, with a companion test pinning that scoping so the fix cannot be over-applied.

### 3. Matcher cache (`fix`)

**a. `SpellingMatch` is immutable in fact, not only in its docstring.** `__slots__` bounds *which* attributes exist; it does not make them read-only. Assigning to a cached match's `_text` changed the value returned to every later reader of that key.

**b. The retained-bytes budget counts the compiled patterns it keeps alive.** A match entry holds its pattern through `_keepalive`, outliving the 64-entry vocabulary cache that compiled it, so an evicted vocabulary's pattern stayed retained while `retained_bytes` reported only the bookkeeping around it.

**c. That estimate is now calibrated rather than reasoned.** `_PATTERN_BYTES_PER_CHAR` was 6, from "the `str` plus roughly twice the text again". Measured against the six vocabularies a real oneDAL comparison compiles — patterns from 8,657 to 3,336,273 characters — it undercounts by a stable **1.46–1.51×** (19.09 MB estimated vs 27.94 MB measured on the largest). Now 9, and still an explicit *lower* bound since `sys.getsizeof` does not reach a compiled pattern's internal allocations. Erring low is the wrong direction for a budget meant to bound growth.

## Bug-fix test contract

- Bug class: `perf.optimization_wired_to_one_of_several_equivalent_paths`, `perf.reuse_key_normalizes_an_ordered_input`, `perf.bounded_cache_budget_omits_what_it_retains`, `perf.fixed_layout_fast_path_field_decoding` — four new entries in `tests/regressions/manifest_performance.py`.
- Publicly observable failure: (1) every `dump`/`compare` over an ELF artifact pays one seek-and-read per symbol name plus a `construct` parse per entry, across two tables; (2) a `STV_SINGLETON` export is silently dropped from the export set; (3) two release sides whose `-I` order differs share one acquired surface, so the contract reconciled against is the one the other order's parse produced; (4) a cached `SpellingMatch` mutated by one reader is observed changed by the next, and the matcher cache retains compiled patterns its own budget does not bound.
- Regression test fails on base: yes — by mutation, not assumption. Reverting the buffering wrap fails the engagement test; restoring `sorted(...)` on `includes` fails 5 of 9 include-order tests; dropping the pattern accounting and the read-only `__setattr__` fails 8 of 20 cache tests; reverting `_PATTERN_BYTES_PER_CHAR` to 6 fails 5 calibration tests. Ten plausible decoder errors were mutated and all ten are caught.
- Negative control: every new module carries one. `buffered[0]` must be non-empty, so an implementation returning empty sets cannot pass equality against itself; the permutation test asserts permutations were generated; `test_an_ordinary_pattern_is_still_admitted` stops the bypass swallowing the normal case; `test_a_table_at_the_cap_is_still_decoded` stops the cap rejecting everything; an absurdly large `_PATTERN_BYTES_PER_CHAR` fails, so a lower-bound-only test cannot accept a constant that disables caching.
- Public-surface test: the ELF work is exercised through `_pyelftools_exported_symbols`' real entry point over real gcc-built shared objects, and the include-order fix through `build_side_identity`/`SurfaceAcquisitionIdentity.key()`, the objects the release fan-out actually keys on.
- Axes covered: ELF class (32/64) × endianness (LE/BE); field (`st_name`, binding, visibility, `st_shndx`); refusal (entry-size, ELF class, partial record, short read, oversize, malformed header, stream error, absent stream); symbol table (`.dynsym`/`.symtab`); symbol shape (exported, weak, hidden, static-only, global-absolute, many, long-named); input health (well-formed, truncated); include sequence (permuted, duplicate-bearing, equal); identity input kind (ordered vs membership-only); pattern population (1…16384 spellings, shared/distinct/oversized); `SpellingMatch` slot × operation.
- Real-dependency test: yes, and the oracle is the displaced parser rather than this code's own arithmetic. `tests/test_elf_symbol_fastpath.py` builds its entries with pyelftools' own `Elf_Sym` struct and reads them back with its own `SymbolTableSection`, covering all four class/endianness combinations — this host emits only 64-bit LE, so a gcc-based test would leave both big-endian variants and the differently-ordered ELF32 record unexercised. `tests/test_dumper_elf_symbols_buffering.py` drives real gcc-built shared objects through the real entry point, pairing every equivalence assertion with a mechanism observation so a silently no-op fast path cannot pass by comparing the slow path with itself.
- False-positive removed / real break preserved: export filtering is asserted *unchanged*, not merely quiet. Each fixture is compared for full equality against a control with the mechanism disabled, so a symbol newly admitted or newly dropped fails. Each filter has an input that makes it observable — a weak export for the binding filter, a *global* absolute symbol for the reserved-index filter — because mutation testing showed that without them, dropping `STB_WEAK` or `SHN_ABS` from either decoder changed no result. All four such mutations are now caught.
- General invariant: stated and exercised beyond the reported input in each case. **Decoder** — for every supported (class, endianness) and every field value, the fast path agrees with pyelftools' own parse of the same bytes; refusals return `None`, never a wrong or empty answer. **Fallback** — for every fixture shape, the result with the bulk decoder engaged equals the result with it declining, so the two implementations of the same filter cannot drift. **Includes** — for any two include sequences unequal as sequences, the key differs, over all permutations of several path sets (the only shape that isolates it: swapping one path also changes the set). **Cache** — `retained_bytes` is monotone in what is held, and a put-then-evict round trip returns to its starting value, which a one-directional "it grows" test cannot catch. **Estimator** — at least measured retention and within a small factor of it, across three orders of magnitude of vocabulary size.
- Known unsupported cases: the visibility filter in the *legacy* loop cannot be reached through a gcc-built `.so` — the linker localizes hidden symbols, which is what hidden visibility means for a shared object (probed directly, including a `.hidden` directive on a pure-asm global, which still linked `STB_LOCAL`). The fast path's own visibility decoding is covered exhaustively against pyelftools. Symbol *version* decoding is untouched by this PR. Real-binary equivalence is x86-64 LE only; the ELF32 and big-endian evidence is synthetic, built by pyelftools itself.

## Checklist

- [x] Changelog fragments added (three)
- [ ] Docs updated if user-visible behaviour changed — no user-visible behaviour change
- [x] `ruff` / `mypy abicheck/` / `check_architecture.py` / `check_ai_readiness.py` (0 errors) / `check_bugfix_test_contract.py` pass locally
- [x] Full fast unit suite passes locally

## Measured but deliberately *not* changed: the matcher

The review asked for the matcher to be benchmarked against the real vocabularies before replacing it. Captured from a live oneDAL comparison: **6 vocabularies, 942 real texts**. Zero mismatches between strategies on every text.

| vocabulary | spellings | regex scan | indexed | trie |
|---|---:|---:|---:|---:|
| stdlib qualified | 39,824 | 9732 ms | 3840 ms | **33.8 ms** |
| stdlib unqualified | 35,503 | 9142 ms | 3401 ms | **22.8 ms** |
| records/typedefs | 8,835 | 1068 ms | 277 ms | **27.7 ms** |
| oneDAL's own types | 1,291 | 97 ms | 84 ms | **27.3 ms** |
| small | 144 | **8.7 ms** | 21.6 ms | 19.4 ms |
| small | 134 | **8.3 ms** | 20.9 ms | 19.5 ms |

The trie is up to **401× faster** on scan, and handles the shared-prefix case cleanly (the 1,291-spelling oneDAL vocabulary is nearly all `oneapi::dal::detail::v1::`, where the trie wins 3.6× and a naive candidate bucket manages only 1.16×). **Then retained memory rules it out:** 465.8 MB, against 62.0 MB for the current regex and 10.8 MB for the indexed variant. For a job being pushed toward a memory limit, trading ~20s of CPU for ~466 MB resident is the wrong direction — which is why the review's insistence on measuring retention *separately* mattered.

So: no replacement here. The indexed variant is the candidate worth pursuing (2.4–2.9× faster scan than regex at one-sixth its memory), with a crossover between 144 and 1,291 spellings below which the regex wins. Left for a follow-up rather than bundled into this PR.

## Still open from the review

- The memory-limit matrix (1/2/3/6 active members under the real cgroup limit) and a phase-aligned memory timeline.
- Demangle request overlap and export-index intermediates.
- Removing the extra neutral contract construction where reusable evidence exists.

The `known_gaps` on the four new `BUG_CLASSES` entries record what each fix still leaves open — notably that the vocabulary cache is still bounded by entry *count* rather than by vocabulary bytes, so one coordinated budget across vocabulary keys, compiled matchers and results is only partly answered here.

https://claude.ai/code/session_01RnAhZWWFHCbAaSLPimS5A7


napetrov:
### `integration-tests (windows-latest)` on `7524be2` — not reachable from this PR's diff

One failure out of 232:

```
FAILED tests/test_release_public_surface_integration.py::
  TestPublicTypeChangeRenderedOnce::test_the_verdict_is_unchanged_by_the_fold
  AssertionError: assert 'COMPATIBLE' == 'BREAKING'
```

**Why it isn't this PR's.** That lane was green on `983b1fe`. The only code between `983b1fe` and `7524be2` is the ELF symbol fast path — `abicheck/extract/elf_symbol_fastpath.py` and the `_extract_symbols` call site inside `dumper_elf_symbols._pyelftools_exported_symbols`. That function has exactly one caller in the tree, `dumper._dump_elf` (`abicheck/dumper.py:1516`), which is the ELF-specific dump.

The failing test does not go through it. The same job's own warnings show it running the PE path against MSVC-built binaries:

```
abicheck/service_dump_native_pe.py:144: UserWarning: None of the provided headers
matched exported symbols in 'libB.so'. This commonly happens when a C++ PE binary
uses a name-mangling scheme (e.g. MSVC) different from the compiler used to parse
the headers. Falling back to export-table mode - header-based scoping had no effect.
```

`service_dump_native_pe.py` contains no reference to `_pyelftools_exported_symbols` or `_extract_symbols`. So the changed code is unreachable from this test on this platform, and it cannot have moved the verdict.

**What the failure looks like on its own terms.** The sibling assertions in the same class passed, so the `Cfg` findings were present and the product-level fold behaved — only the verdict differed. The warning above says header-based scoping fell back to export-table mode for this fixture, which means the run had no header-derived layout evidence for the added struct field that the test's `BREAKING` expectation rests on. A verdict that depends on that evidence, on a platform where the scoping demonstrably fell back, is the part worth looking at; this comment is not a claim that the test is wrong, only that this PR did not change what it exercises.

**No fix ported, because there is nothing to port** — I found no existing PR or revert addressing it. I have not widened this PR to chase it, per AGENTS.md's rule against widening on an unrelated failure.

**Re-run:** spending the one permitted re-run here, since the diff-reachability argument above is exactly the "not this PR's" case that a single confirming re-run is for. I'll keep watching and will report if it reproduces.

---

napetrov:
**Resolved:** `integration-tests (windows-latest)` passed on `2c083f9` — the full CI run completed `success` — so `test_the_verdict_is_unchanged_by_the_fold` did not reproduce. That is consistent with the reachability analysis above: the ELF fast path is unreachable from that test's PE code path.

Recording it as a one-off rather than a fix, since nothing was changed to address it. If it recurs on another PR, the thread worth pulling is the one the job's own warnings point at — header-based scoping falling back to export-table mode for that fixture, which leaves the run without the header-derived layout evidence the `BREAKING` expectation depends on.

---

PR #1336, opened 2026-09-19
fix(compare): make the shared type-spelling caches thread-safe under the release fan-out

napetrov:
## Summary

Make the two process-wide type-spelling caches thread-safe, so a concurrent multi-library `compare` stops failing members with an internal `KeyError`, and give an unexpected member failure a diagnosable account of itself.

## Motivation

A directory/package `compare` dispatches every release member through a `ThreadPoolExecutor` in one address space, so `MATCH_CACHE` and `VOCABULARY_CACHE` (`abicheck/compare/spelling_match_cache.py`) are shared by every worker. Their compound operations were unsynchronized.

The GIL already makes each individual `OrderedDict` call atomic, which is exactly why this was neither obvious nor caught: the defect is a *sequence*.

1. Worker A executes `_entries.get(key)` and receives a cached value.
2. Worker B's `put` admits an entry and evicts A's key.
3. Worker A executes `_entries.move_to_end(key)` and raises `KeyError`.

Observed on a real six-member oneDAL release comparison, which failed `libonedal.so`, `libonedal_core.so` and `libonedal_dpc.so` with e.g. `Error comparing libonedal_core.so…: (138821602538640, 'const dense&', 0, 12)` — that tuple is a match-cache key `(id(pattern), text, start, end)` — reporting them as `verdict: ERROR` / `operational: extraction_error` / `scope: incomplete`. Another run of the same inputs at the same revision completed cleanly, so the fault is nondeterministic. The cache's own suite (equivalence, eviction, accounting, immutability) passed at that revision because every one of its tests ran on a single thread.

The second half of the incident was diagnostic: `str(KeyError(...))` is its argument alone, so the release's output named neither the exception class nor the failing subsystem.

## Changes

- **`abicheck/compare/spelling_match_cache.py`** — one narrow lock per cache. Every multi-step transition is now atomic under it: lookup-plus-recency-update, admission-plus-accounting-plus-eviction, pattern reference counting, vocabulary publication, and `clear`. A "Concurrency contract" section in the module docstring states the invariants.
  - **Expensive work never runs under a lock.** Regex compilation (`compile_fn`) and matching (`compute`) happen outside it; publication is rechecked afterwards. Duplicate computation on simultaneous misses is therefore possible **by design and stated** — the alternative, one global lock spanning a multi-megabyte alternation compile, serializes the very member comparisons the fan-out exists to run concurrently. The first publication wins, so one vocabulary still resolves to exactly one compiled pattern, which is what the match cache's `id()`-derived token keying depends on.
  - **`clear()` is a lifecycle reset, not a barrier.** It advances a generation counter; a computation already in flight still completes and serves its caller but is never published into the new epoch. No entry, byte count or pattern reference survives the clear that dropped it, and nothing advertises a quiescence guarantee that is not provided.
  - Pattern identity safety is unchanged: a token is only ever *stored* while `put` holds a strong reference to the pattern it names, so a recycled `id()` cannot resolve to another pattern's results. Immutable results, negative-result caching, oversized-input bypass and the byte/entry budgets are all untouched.
- **`abicheck/frontends/cli/release_member_errors.py`** — an unexpected member failure keeps its identity: the entry gains `error_type`, and the full traceback goes to the `abicheck.release` logger. The classification is deliberately unchanged — `OperationalStatus.EXTRACTION_ERROR` is already defined as "a library failed to dump/extract/compare", which covers an internal comparison failure, so no schema change is bought merely to rename an error. `not_comparable` (ADR-050 D2) and `unsupported` (ADR-065 D6) are not reclassified and emit no traceback. The fan-out's *outer* exception boundary moved here too (`member_dispatch_failure_entry`), next to the inner classification it backs up.
- **`tests/test_spelling_match_cache_concurrency.py`** (new) — forced-interleaving regressions.
- **`tests/test_compare_release_concurrency_integration.py`** (new) — the real concurrent entry point on a compiled four-member ELF release, plus member-failure injection.
- **`tests/regressions/manifest_concurrency.py`** (new) — registers the bug class `concurrency.compound_cache_operation_under_fan_out`, wired into `BUG_CLASSES`.

## Bug-fix test contract

- Bug class: `concurrency.compound_cache_operation_under_fan_out` — new entry in `tests/regressions/manifest_concurrency.py`. A process-wide memo the release fan-out's workers share must make every *compound* state transition atomic, not merely every individual container operation.
- Publicly observable failure: `abicheck compare <old-dir> <new-dir>` over a multi-member release nondeterministically reports members as `verdict: ERROR` / `operational: extraction_error` / `scope: incomplete` with `Error comparing <lib>: (<id>, '<type spelling>', <start>, <end>)`; the same inputs succeed on a rerun.
- Regression test fails on base: yes. Against `950efbc` with only the test files applied, 4 tests fail — `TestForcedGetEvictRace::test_lookup_survives_a_concurrent_eviction_of_the_key_it_read` (`KeyError`, the reported defect), `TestForcedVocabularyRace::test_hit_survives_a_concurrent_eviction_of_the_key_it_matched` (`KeyError(frozenset({'dense'}))`, the same class in the sibling cache), and both `TestClearLifecycle` tests (pre-clear state resurrected / accounting inconsistent after concurrent `clear`).
- Negative control: `test_a_read_never_reports_a_failed_lookup_as_an_empty_match_set` forecloses the forbidden "fix" — an empty tuple is a legitimate cached answer, so a failed lookup answering `()` must stay distinguishable from a genuine empty result. `test_simultaneous_misses_publish_exactly_one_pattern` uses a `threading.Barrier` inside `compile_fn`, which can only be reached if the compile is genuinely lock-free, so a fix that serialized compilation deadlocks it instead of passing. `test_a_failing_compilation_leaves_no_stuck_lock_or_poisoned_entry` joins a second thread that would hang on a leaked lock. `test_findings_are_stable_across_worker_counts` spies on `_compare_release_parallel` and asserts sequential vs. parallel dispatch actually differed, so the differential cannot compare one configuration with itself.
- Public-surface test: yes — `tests/test_compare_release_concurrency_integration.py::TestConcurrentReleaseWorkflow` runs the real `abicheck compare` CLI over a compiled four-member C++ release (`-H` header evidence, `-o json=`), asserting complete member coverage, no `ERROR`/`unsupported` member, `run_outcome.scope != incomplete`, the seeded ABI break still detected (exit 4), stable semantic findings across three repeated runs, and across pinned worker counts 1/2/4.
- Axes covered: transition (lookup-then-recency-update, test-then-insert, insert-then-account-then-evict, compute-then-publish, clear-across-in-flight-computation) × cache (match, vocabulary) × result shape (positive, empty, oversized-bypass) × workers (1, 2, 4, 8).
- General invariant: stated and executed beyond the reported input, not in prose. (a) A cache read is a pure read — it may report a miss, may never raise, may never answer anything but the real value; forced with a mapping hook that parks the reader mid-transition while a sibling's *ordinary public* `put`/`get_or_compile` evicts the key it read (no direct `_entries` mutation, so the cache's own synchronization is what is under test), and the test asserts the interleaving was genuinely attempted and the eviction genuinely happened. (b) Accounting is derivable: `_assert_accounting_consistent` re-derives retained bytes, entry count and per-pattern reference counts from the entries actually present — an **independent** derivation, not the implementation's incremental arithmetic — after every contention run. (c) Every cached result is checked against the **uncached** matcher (`finditer_allow_nested`) on text, spans, ordering and nested matches, across 2/4/8 workers, multiple patterns and positive/empty results, under a monkeypatched 16-entry cap that makes every admission an eviction.

## Checklist

- [x] Changelog fragment added (`changelog.d/20260919_concurrent_release_cache_safety.md`)
- [ ] Docs updated if user-visible behaviour changed — no user-facing behaviour change beyond the added `error_type` field and the restored correctness
- [x] `ruff check` / `ruff format` / `mypy abicheck/` / `check_architecture.py` / `check_ai_readiness.py` pass locally

---

## Validation

Revision: branched from `950efbc` (current `origin/main` — no intervening commits, no open PRs at the time of writing).

**Commands run**

```
ruff check abicheck/ tests/ ; ruff format --check
mypy abicheck/                                   # 0 errors, 896 files
python scripts/check_architecture.py             # 0 errors
python scripts/check_ai_readiness.py             # 0 errors, 128 pre-existing warnings
BUGFIX_CONTRACT_BODY_FILE=... python scripts/check_bugfix_test_contract.py   # OK
python -m pytest tests/test_spelling_match_cache.py \
  tests/test_spelling_match_cache_retention.py \
  tests/test_spelling_match_cache_concurrency.py \
  tests/test_release_member_errors.py tests/test_compare_release.py \
  tests/test_compare_release_parallel_dedup.py \
  tests/test_compare_release_concurrency_integration.py \
  tests/test_provenance_symlink_fixture.py tests/test_type_reachability*.py    # 360 passed
python -m pytest tests/ -m "not integration and not libabigail and not abicc and not slow and not golden" -n 4 -x   # exit 0
```

**Contention timing, before vs. after** (4-core host, best of 3). The point is that the lock did not serialize the expensive work.

- Miss-heavy (the matcher runs on nearly every call, so regex work dominates): 1/2/4 workers at 0.093/0.195/0.357s after vs. 0.092/0.187/0.350s before — ~2% overhead, and computed-matches/s is flat across worker counts *both before and after* (the workload is GIL-bound in `re`, not lock-bound), so no serialization was introduced.
- Hit-heavy microbenchmark (nothing but cache lookups, the worst case for a bookkeeping lock): ~8–25% slower. This is the lock on the hit path and is the correctness cost.
- End-to-end `compare` on the four-member fixture: 1.33s vs. 1.31s min of 3 — within noise.

**Semantic delta / provenance**

None intended and none observed. `_semantic_projection` over the four-member fixture (per-library verdict, finding count, and the sorted (kind, symbol, severity) set) is identical across repeated runs, across worker counts 1/2/4, and the seeded ABI break is still detected at exit 4.

The corrected public-header contract from #1330 is preserved **by construction**: this patch touches `compare/spelling_match_cache.py`, `cli_compare_release_pairwise.py` and `frontends/cli/release_member_errors.py`, and #1330 touched `extract/dependency_header_roots.py`, `extract/path_aliases.py` and `provenance.py` — a disjoint set, with no shared call path modified. `tests/test_provenance_symlink_fixture.py`, `tests/test_provenance_path_aliases.py` and `tests/test_provenance.py` all pass unchanged. Nothing here resolves persisted paths or broadens public ownership, and the fixture-specific oneDAL counts reported for #1330 (748→1186 export obligations, 35→45 missing exports, 5707→5069 undocumented exports, 11476 findings unchanged, exit 4) are the corrected reference this patch builds on top of, not something to revert toward.

## Limitations

- **Full oneDAL validation remains pending.** The oneDAL snapshot directory and release ELF tree were not available in this environment, so the six-member run that produced the report has not been re-run. The end-to-end claim here rests on a **synthetic compiled four-member C++ fixture**, whose members share type spellings deliberately so several workers land on the same match-cache keys, but whose vocabularies are orders of magnitude smaller than oneDAL's (largest real compiled alternation ≈3.3M characters) and which therefore reaches far less eviction pressure per unit of work. This is **not** a verified oneDAL fix. The forced-interleaving tests do not depend on that scale; the end-to-end stability claim does.
- **Only the two type-spelling caches are covered.** The invariant is stated for any process-wide memo the fan-out shares, but no audit of the remaining module-global mutable state reachable from a member comparison was performed. A sibling instance of the same class elsewhere in the engine would not be caught by these tests. Both gaps are recorded as `KnownGap` entries on the registered bug class.
- The bounded multi-worker contention tests are not *forced* interleavings and do pass against the unsynchronized implementation; they assert the accounting and oracle invariants, while the forced tests are the actual falsifiers. Stated here so the evidence is not overclaimed.
- The 11 `bundle_unresolved_intra_dependency` findings about an absent libsycl system provider are untouched: neither suppressed nor papered over with an injected provider. They are an input-configuration matter for a separate report.

## Handoff to the performance/memory workstream

Shared files: `abicheck/compare/spelling_match_cache.py` (both caches), `abicheck/compare/spelling_pattern.py` (its only caller).

API/lifecycle contracts this patch establishes, which a cache-admission/ownership redesign must preserve or explicitly renegotiate:

1. `_MatchCache.get` is a **pure read** — miss or value, never an exception, never a fabricated empty result.
2. `_MatchCache.put(key, matches, pattern, generation=None)` gained a trailing optional `generation`. Admission, byte/entry accounting, pattern reference counting and the eviction loop must remain **one** transition; each reads state the others write.
3. `generation` / `clear()`: `clear` is a lifecycle reset that advances an epoch, **not** a barrier. A computation in flight completes and serves its caller but must not publish into the new epoch. If the redesign wants a real quiescent boundary, that is a different contract and must be stated, not assumed.
4. **Expensive work stays outside the locks.** Duplicate computation on simultaneous misses is permitted; duplicate *publication* is not — one vocabulary must resolve to one pattern object, because the match cache's token is `id(pattern)`.
5. Pattern identity: a token may only be **stored** while a strong reference to its pattern is held. Any admission redesign that decouples the keepalive reference from entry insertion breaks this and reintroduces `id()` reuse.
6. Locks are never held across a caller-supplied callback, and the two caches' locks are never held simultaneously. Keep both properties.

Replacing the giant regex, worker sizing, repeated header acquisition and the admission/ownership redesign are explicitly **not** attempted here; this patch is deliberately small enough to build on. Please integrate it and keep `tests/test_spelling_match_cache_concurrency.py` and `tests/test_compare_release_concurrency_integration.py` green before final validation.

https://claude.ai/code/session_01CBHXHAv6CSSB2Tmqx9FuVr

---
_Generated by [Claude Code](https://claude.ai/code/session_01CBHXHAv6CSSB2Tmqx9FuVr)_


napetrov:
**Collision with #1335 — read before reviewing this PR in isolation.**

#1335 (match-cache ownership / the performance workstream) touches the same three files this PR does — `compare/spelling_match_cache.py`, `compare/spelling_pattern.py`, `cli_compare_release_pairwise.py` — plus the same two test modules. It was not open when this work started. The two cannot both merge as-is. Full side-by-side in [my comment there](https://github.com/abicheck/abicheck/pull/1335#issuecomment-5740604595); the short version, including where that PR is the better answer:

- **#1335 already closes the reported race**, by a different design: `matches_for` holds a module-level `_CACHE_LOCK` across `token_for` + `get` and again across `put`, and nothing outside that module calls `MATCH_CACHE.get`/`.put`. Its coarser single-`RLock` is not the per-cache locking this PR adds, but it is sufficient for the `KeyError`.
- **#1335's `_PatternRegistry` generation token is better than this PR's `id(pattern)` keeping.** It removes the id-reuse hazard structurally, where this PR argues it away from a strong reference. That part of this PR should not survive a merge of the two.

Two things this PR has that #1335 does not, and which are worth keeping whichever locking design lands:

1. **The `clear()` lifecycle contract.** #1335 computes outside the lock and re-acquires to `put`, so a `clear_caches()` landing in that window is admitted afterwards — pre-clear state reappears and re-registers its pattern. The generation counter here closes that, and is independent of the locking design.
2. **Threaded regression tests.** #1335 deliberately has none (`test_spelling_match_cache_retention.py` imports no `threading`; `test_spelling_match_cache.py` argues a thread-based version "would be timing-dependent and... most likely vacuous"). Fair about *stress* tests — but the oneDAL failure was unreachable single-threaded, which is exactly why this module's whole suite passed at `950efbc`. The tests here force the interleaving deterministically rather than racing for it, and fail against `950efbc`.

The member-failure diagnostics in this PR (`error_type` + a traceback on the `abicheck.release` logger, and the outer dispatch boundary moving to `release_member_errors`) are unrelated to the cache and do not conflict with #1335 at all.

**Suggested resolution:** land #1335, then reduce this PR to (a) the generation/`clear` contract, (b) these threaded tests rebased onto that design, and (c) the member-failure diagnostics. Holding this as a draft rather than rebasing speculatively onto an unmerged PR — happy to do that rework once #1335's shape is settled.

---

napetrov:
**`codecov/patch` on `5f728e0` is a stale report, not a real coverage gap.** Recording the evidence rather than pushing at it.

The check reports `43.37% of diff hit`, with 47 lines missing — **byte-identical** to what it reported on the previous head `4beea4a`, down to the per-file breakdown. `5f728e0` changed only test files (155 additions across `tests/test_spelling_match_cache_concurrency.py` and `tests/test_compare_release_concurrency_integration.py`, zero source lines), so an unchanged *percentage* is expected only if the report backing it is also unchanged — i.e. no new upload had arrived for this head when the check posted. The canonical coverage lane is Linux/Python-3.13 in `ci.yml`, and the rest of that workflow set was still in progress when the check fired.

Measured locally under CI's own marker expression, so the comparison is like-for-like:

```
pytest tests/test_spelling_match_cache_concurrency.py \
       tests/test_compare_release_concurrency_integration.py \
       tests/test_spelling_match_cache.py \
       tests/test_spelling_match_cache_retention.py \
       tests/test_release_member_errors.py \
  -m "not integration and not libabigail and not abicc and not slow" \
  --cov=abicheck.compare.spelling_match_cache \
  --cov=abicheck.frontends.cli.release_member_errors --cov-branch

Name                                              Stmts   Miss Branch BrPart  Cover
abicheck/compare/spelling_match_cache.py            178      0     36      0   100%
abicheck/frontends/cli/release_member_errors.py      30      0      8      0   100%
98 passed, 3 deselected
```

100% statement **and** branch on both modules. The three tests deselected by that filter are the `@pytest.mark.integration` class that builds compiled ELF fixtures; they cover release-fan-out behaviour, not these modules, so their exclusion does not move these numbers. Both lines this PR adds to `cli_compare_release_pairwise.py` are also covered, by `TestFanOutExceptionBoundaryInSitu::test_one_worker_failure_does_not_lose_its_siblings`, which drives `_compare_release_parallel`'s real `except` arm.

For the record, the earlier `test-action summary` failure on the superseded head `22aa6ab` was also not a defect: nearly every job in that run reads `cancelled`, because my own push to `5f728e0` superseded it and the summary job then failed on the cancellations.

Not re-running anything yet — the coverage lane's own upload should re-post this check, and forcing a re-run while the workflow set is still in flight would only duplicate it. Keeping the PR watched; if the check is still red once the canonical lane has demonstrably uploaded for this head, that is a real finding and I will treat it as one.

---

PR #1338, opened 2026-09-19
perf(header-graph): release the clang AST before the graph build, and gate the attach's memory

napetrov:
## Summary

Stop holding the parsed clang AST and the L5 header graph at the same time, and — separately — make the header-graph attach's *memory* gated instead of measured-only.

**Read the Results section before the diff.** The handoff's premise — that a member's peak was the AST and the graph held together — is wrong, and this PR's own change does not improve either number a release fan-out's per-member budget is sized from. The measurement that shows why, and the measured ceiling of the one lever that remains, are the more valuable half of this PR.

## Motivation

Follow-up to PR #1335's measurement (`docs/contribute/measurements/header-graph-attach-memory.md`), which localised a release member's peak inside `_attach_header_graph` and left two candidate directions plus an explicitly ungated memory dimension.

`_attach_header_graph` parsed a whole `clang -ast-dump=json` document into Python dicts and then built the header graph while that tree was still resident. The graph build only ever reads four pure, compact projections of the AST, so it never needed the tree at all.

## Changes

- **`abicheck/buildsource/header_graph_ast_projection.py`** (new) — `HeaderGraphAstProjection` + `project_header_graph_ast`, holding exactly what `build_header_only_graph` reads: `index_declared_type_files`, `parse_clang_ast_types`, `parse_clang_ast_calls`, and the lazily-computed `index_declared_entity_files`. The AST-sourced seeding (`seed_ast_graph` and its two helpers) moved here with it, since `header_graph.py` sits at its `architecture/debt.yaml` line cap; the snapshot-sourced fallback stayed behind.
- **`service_header_graph_attach.py`** — project, drop the tree, build, then drop the projection too.
- **`build_header_only_graph`** — new `ast_projection=` alongside `ast_root=`; passing both raises.
- **`scripts/check_header_graph_perf.py`** — gates `attach_peak_rss_mib` and `attach_end_rss_mib` alongside the three time metrics (report schema `abicheck-header-graph-perf/3`), each measured in a fresh subprocess over a new STL-bearing fixture. `performance.yml` gates both PR-vs-base.
- Docs: the measurement doc and `known-gaps.md` carry the reattribution, the full per-run data, and the measured ceiling of the remaining lever.

## Results

oneDAL 2024.7 `libonedal_core.so.2` (conda-forge `dal`/`dal-devel`), default `castxml` backend, cold AST cache per run (`XDG_CACHE_HOME` to a fresh empty dir, never inherited), no `tracemalloc`. Three runs per side, then two more with the final code. MiB:

| | before (×3) | after (×2, final) |
|---|---|---|
| graph build's own cost | +147.4 / +147.5 / +142.7 | **−25.6 / −30.0** |
| attach peak (`VmHWM`) | 2215.4 / 2218.0 / 2215.4 | 2215.2 / 2214.9 |
| RSS at end of attach | 1287.6 / 1286.1 / 1275.9 | 1295.2 / 1289.8 |
| graph nodes / edges | 49481 / 98330 | 49481 / 98330 |

- **Peak: unchanged (0.05%).** It lives inside `json.load`, where the document is held as one `str` while the tree is built from it — `document + tree`, never `tree + graph`. Checked rather than assumed that the document copy is avoidable: on a 247 MB AST, `json.load(fh)` peaks at 640.9 MiB and an explicit read/decode/`del raw`/`loads` at 641.0. CPython already drops the source buffer.
- **Mid-attach residency: ~175 MiB lower** at the graph-build point (1302–1307 vs 1475–1483). Reproducible.
- **Steady-state retention: ~8–12 MiB *higher*.** Likely the inverse of the pinning effect the original measurement identified — freeing the AST early returns its arenas, so the graph faults in fresh pages instead of reusing ones the parse had dirtied. This survived a real bug fix along the way (the projection was a local, so it outlived the graph build, its only consumer); freeing it at the build's end moved the build from +24 to −27 MiB but did not close this gap.

**So: keep or drop is a reviewer's call, and here is the case.** It is evidence-identical and strictly non-worse during the attach; the ~10 MiB steady-state cost is real; and the projection is the executable statement of what a pruned tree would have to preserve, which is the only direction left with a measured ceiling.

**That remaining lever:** prune the *tree* to the keys the four readers touch. A recursive key-whitelist copy of the same 247 MB AST is 224.9 MiB of 359.8 MiB — **62%**, so ~38% is fields nothing reads (~400 MiB of oneDAL's 2215 MiB peak). Not attempted here: it must run as an `object_pairs_hook`, which this repo has already measured at 13–30% wall time across the *whole* document against a 50% `attach_ms` gate, and a whitelist wrong in one key drops edges silently rather than raising.

Not claimed: a six-member acceptance run. None was performed.

## Bug-fix test contract

- Bug class: `perf.evidence_released_before_its_consumer_runs` (new entry in `tests/regressions/manifest_performance.py`) — reordering when a large intermediate is released must not change what the structure built from it contains.
- Publicly observable failure: a projection missing one of the four readers yields a graph with fewer nodes/edges, so `declaration_renamed` / `public_reachability_changed` findings and the `evidence_metrics` / `layer_coverage` report sections silently shrink. It raises nothing anywhere.
- Regression test fails on base: `tests/test_header_graph_ast_projection.py::TestAstReleasedBeforeGraphBuild` fails against the pre-change ordering (verified by mutating `ast_root = None` away: `assert True is False`).
- Negative control: each of the four projection members blanked in turn fails 3–6 tests; blanking the laziness rule fails 5. Verified by mutation, not asserted.
- Public-surface test: real `compare` CLI on a real old/new pair (the L5 graph diff runs from `cli_buildsource_helpers`, never from `checker.compare()`) — 1913 findings, same verdict, same exit code, every report section byte-equal apart from one `extractor.duration_seconds`.
- Axes covered: entry point (`ast_root` / `ast_projection` / neither) × evidence (type edges / call edges / reference edges / none) × snapshot seeding (empty / seeded function / seeded variable).
- Malicious fixture + side-effect absence: this diff adds no interpolation to any workflow — its entire non-comment change to `.github/workflows/performance.yml` is two literal flag arguments (`--regress-min-delta-ms-attach_peak_rss 48`, `--regress-min-delta-ms-attach_end_rss 48`), with no `${{ }}` expression added anywhere. The executing guard is `tests/test_workflow_untrusted_text_interpolation.py`, which is repo-wide rather than per-site (`workflow_paths()` globs every `.github/workflows/*.yml`, so `performance.yml` is in scope by construction) and runs hostile payloads through a real shell rather than asserting YAML text: `test_hostile_pr_body_is_data_not_commands` executes each payload and asserts the side effect is absent, `test_the_harness_detects_a_genuinely_unsafe_step` and `test_at_least_one_payload_fires_against_the_unsafe_script` are its negative controls, so a harness that had stopped executing anything fails rather than passing vacuously. All pass on this head (17 passed, 2 skipped).
- General invariant: for **any** AST, the graph built from its projection is indistinguishable from the graph built from the AST — checked over a structural enumeration rather than one fixture, against an oracle that is the four readers applied directly (not the projection's own code). That distinction is load-bearing: because `build_header_only_graph(snap, ast_root)` now projects internally, a plain two-path differential is blind to a blanked member, and a `type_files={}` mutation survived the whole matrix before `TestEveryProjectionMemberIsLoadBearing` was added. Plus a vacuity guard on the oracle itself. The two CI-fix commits on this branch each state their own class the same way — see below.

## Two CI failures this PR caused, and fixed

Both were mine, and both are now green.

1. **`Header-graph attach-cost regression (PR vs base)` — `ModuleNotFoundError`.** That job deliberately runs **head's** copy of `check_header_graph_perf.py` against **base's** installed package, so every `abicheck` import in that script is a cross-version compatibility surface. Moving `HEADER_CALL_GRAPH_PASS` to its new owner module took the base measurement down before it recorded a single sample, and the job's existing degrade-to-report-only grep matches a different symbol, so it failed outright rather than degrading. Fixed by resolving the constant from either owner — still by import, never a hard-coded literal, since a literal would keep "working" after a future rename while verifying a pass name no build stamps. Class checked by really hiding the module across four cases (pre-move shape, this build, a rename, both-missing-must-raise).

2. **The new memory metric failed on a *better-than-baseline* result.** `attach_retained_mib` was a signed delta (RSS after the attach minus RSS before it), but `perf_measurement.is_gateable` is written for wall-clock durations and rejects anything `<= 0`. A retained-memory delta is legitimately negative whenever the attach releases more than it allocates — exactly what the clang backend does, since its AST comes from the in-process memo the primary pass wrote, so the attach frees a tree it never allocated. CI failed with `measured attach_retained_mib=-28.37890625 is not a gateable value`. Fixed the metric rather than weakening the shared predicate: it is now `attach_end_rss_mib`, absolute RSS, which cannot be negative. The arithmetic moved into `_memory_metrics()` purely so it is testable without a clang toolchain — this defect was only reachable through a full measurement before, which is why it shipped.

## Note for reviewers

The last commits on this branch are **unsigned**: the environment's commit-signing helper (`gpg.ssh.program = /tmp/code-sign`) was deleted mid-session by another process's `/tmp` cleanup and could not be restored. Earlier commits on the branch are signed. The same deletion made every `git commit` fail in that container, which is why a local `verify.py --profile pr` reported 43 unit failures that CI does not see — they are all tests that drive real git, they pass with `commit.gpgsign=false` (550 passed, 0 failed), and a bare `git init && git commit` in an empty directory reproduces the failure with no abicheck code involved.

## Checklist

- [x] `ruff check` / `ruff format --check` / `mypy abicheck/` / `check_architecture.py` / `check_ai_readiness.py` / `check_docs_contract.py` / `check_bugfix_test_contract.py` all clean

https://claude.ai/code/session_01LbCbTsQEjDU6MjrUDSXpsL


napetrov:
## `codecov/patch` is the only red check, and I cannot reproduce it

Status on `0942d4b`: **CI is green** — 23/23 jobs in the main workflow, 8/8 in `performance.yml` (including `Header-graph attach-cost regression (PR vs base)`, which I broke in the first push and fixed in `e251465`). The one remaining failure is `codecov/patch`: *80.00% of diff hit (target 90.00%)*.

I measured it locally instead of guessing, and got a different answer.

**Method.** Ran the unit lane with CI's exact command —

```
pytest tests/ -m "not integration and not libabigail and not abicc and not slow" \
  -n auto --dist worksteal --cov=abicheck --cov-fail-under=95
```

— then computed patch coverage the way Codecov defines it: intersect the diff's **added** lines (`git diff -U0 7c1e332...HEAD -- abicheck/`) with the run's executed lines.

**Result — every added line under `abicheck/` is covered:**

| file | added lines covered |
|---|---|
| `abicheck/buildsource/header_graph_ast_projection.py` | 53/53 (100%) |
| `abicheck/buildsource/header_graph.py` | 7/7 (100%) |
| `abicheck/service_header_graph_attach.py` | 8/8 (100%) |
| `abicheck/buildsource/source_graph_findings.py` | 2/2 (100%) |
| **total** | **70/70 (100%)** |

Whole-project coverage on the same run: **96.39%**, i.e. the 95% floor is met. `header_graph_ast_projection.py` does not appear in the `--cov-report=term-missing:skip-covered` output at all, which is that reporter's way of saying the file is fully covered.

Codecov's comment attributes *7 missing and 1 partial* to that same file. I could not inspect its line-level data to reconcile the two — `api.codecov.io` is blocked by this environment's egress proxy — so I'm reporting a disagreement, not asserting the check is wrong.

**The one hypothesis the timestamps support:** Codecov reported at `00:09:55` on `e251465`, whose CI run started ~`23:54`; the unit lane takes ~17 minutes, so it would have finished around `00:11` — *after* Codecov had already evaluated. The same pattern holds for `0942d4b` (reported `00:31:39`, run started `00:14`). A partial or early upload would depress patch coverage exactly this way.

**What I deliberately did not do:** add tests to move a number I cannot reproduce a gap in. Padding coverage against a figure that disagrees with the measured one would be filling a gate rather than verifying behaviour, which is the failure mode `AGENTS.md`'s test-quality section exists to prevent. If a maintainer can see Codecov's line-level report and it names real uncovered lines, point me at them and I'll cover them properly.

**Unrelated, for the record:** a local `verify.py --profile pr` run reported `unit-pr: failed` with 43 failures. Those are an artifact of *this container*, not the branch: its git commit-signing helper (`gpg.ssh.program = /tmp/code-sign`) was deleted mid-session, and `commit.gpgsign=true` is set globally, so every `git commit` fails here — including the ones those tests make in throwaway repos. Verified two ways: a bare `git init && git commit` in an empty scratch directory fails identically with `fatal: cannot exec '/tmp/code-sign'`, and re-running all 43 with `GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false` gives **550 passed, 0 failed**. CI, which has working git, shows 0 failed jobs. The same deletion is why the most recent commits on this branch are unsigned.

---

napetrov:
Closing the loop on the comment above: `codecov/patch` flipped to **success** at `00:53:30`, a couple of minutes after I posted it — "All modified and coverable lines are covered by tests."

That confirms the incomplete-upload diagnosis rather than leaving it a guess: the earlier 80% was scored before the ~17-minute unit lane had finished uploading, and the complete upload agrees exactly with the local measurement in that comment (70/70 added lines, 100%). No code change was needed, and none was made for it.

The PR is now green end to end — 23/23 main CI jobs, 8/8 in `performance.yml`, `codecov/patch` and `codecov/project` green — and `mergeable_state` is `clean`.

---

